PR #43, opened 2021-02-08
Add attribute and name readme docs

channa-figure:
I'm not sure, if I need to add more detail.  I copied what was on our provenance-blockchain project, but updated the example commands to work properly.


iramiller:
If this is in response to the `spec` tickets/cards on the project then the `readme.md` files are not matching the intended documentation format.  We should be using this format instead: `x/distribution/spec/README.md` with all of the specific docs for messages, abci state transistions, etc.

PR #51, opened 2021-02-09
Feature/get denom meta endpoint

channa-figure:
Add new endpoint to get the denomination metadata from bank keeper


x/marker/keeper/query_server.go
```diff
@@ -149,3 +149,13 @@ func (k Keeper) Access(c context.Context, req *types.QueryAccessRequest) (*types
...
 	return &types.QueryAccessResponse{Accounts: marker.GetAccessList()}, nil
 }
+
+// DenomMetadata query for metadata on denom
+func (k Keeper) DenomMetadata(c context.Context, req *types.QueryDenomMetadataRequest) (*types.QueryDenomMetadataResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	ctx := sdk.UnwrapSDKContext(c)
+	metadata := k.bankKeeper.GetDenomMetaData(ctx, req.Denom)
```
dpederson-figure, line 159: Can we validate denom here? Always good practice to validate rpc request inputs.

dpederson-figure (approved):

x/marker/keeper/query_server.go
channa-figure: I added a check similar to one in the cosmos sdk bank module to check that the denom is set.

iramiller (approved):
LGTM.

Issue #78, opened 2021-02-17
Implement migration function


iramiller:
## Summary

Implement a migration function for 0.40 that supports the Provenance Blockchain modules and properly registers all legacy amino types.

## Problem Definition

The current migration module uses a migration function from the base sdk that does not properly call our modules to register amino codecs which results in errors when attempting to migrate from 0.39.  https://github.com/provenance-io/provenance/blob/a094578bb969b1242a2d052c7348e20dd45a63b7/app/migrate.go#L73

## Proposal

Replace the migrate function wiring in `migrate.go` to include all of the functionality from `github.com/cosmos/cosmos-sdk/x/genutil/legacy/v040/migrate.go#L39` but also register all of the Provenance Blockchain modules in the RegisterLegacyAminoCodec step.  Add migration wiring for attribute, marker, name, and metadata modules

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #101, opened 2021-02-24
Implementation of migration function for provenance blockchain

iramiller:
## Description

This PR implements a custom migration function for existing v039 provenance blockchains to the current public chain.  The migration was tested against an export file from an existing chain.

The follow command used a 2GB export json file and processed successfully.
```
./build/provenanced -t migrate v1.0.0 ./local_path/export.json > migrate.json 
```

- Fixes an issue in the marker migration where untyped marker_type values would cause a panic
- Fixes an issue with metadata export that used `encoding/json` instead of Amino json
- Fixes a gogo proto compatibility issue with proto Timestamp
- Fixes migration of metadata groups to prevent null points
- Fixes an issue with GroupSpecification identifier being set to Record Group type in migration

closes: #78 

---

write a little note why.


internal/legacy/v040/migrate.go
```diff
@@ -0,0 +1,297 @@
...
+		// the respective key.
+		appState[v040gentypes.ModuleName] = v040Codec.MustMarshalJSON(migrateGenutil(genutilGenState))
+	}
+
+	// Default capabilities initialization
+	capGenesis := captypes.DefaultGenesis()
+	appState[captypes.ModuleName] = v040Codec.MustMarshalJSON(capGenesis)
+
+	/*
+		TODO: default configuration for IBC
```
iramiller, line 241: this is a gaia config... it is not clear that we actually want this.  I filed an issue #102 for future follow up.

iramiller:
this one needs 👀 from @dpederson-figure on the idea of using the `internal` folder for encapsulating our custom migration functions.

dpederson-figure:
The `internal` package is a nice place to put things to reduce public API surface. So, if we don't want to expose the migration types/functions, it is a great place for them.

dpederson-figure (approved):
LGTM

Issue #58, opened 2021-02-10
Add support for RecordSpecification identifiers to the metadata address class.


iramiller:
## Summary

Add a new RecordSpecification type to the Address class.

## Problem Definition

Add a new RecordSpecification identifier to the metadata class.  Format should be 
```go
PrefixRecordSpecification = "recspec"
hrp = PrefixRecordSpecification
requiredLength = 1 + 16 + 32 // type byte plus size of one uuid and one sha256 hash
```
Should contain the ContractSpecification UUID along with the hash of the record name.  Add an iterator prefix that returns the prefix portion of the id bytes (type byte + record specification uuid bytes). Ensure appropriate tests are added to `address_test.go`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #57, opened 2021-02-10
Add ContractSpecification implementation to the metadata module


iramiller:
## Summary

Implement basic metadata structure for GroupSpecification

## Problem Definition

- [x] Review proto message for appropriate identifiers and structure, update as needed.  Change `repeated ConsiderationSpec` to be a repeated list of `RecordSpecification` identifiers. (this is a new type for the Address class)
- [x] Implement ValidateBasic for object
- [x] Add NewContractSpecification functions for creating an instance of ContractSpecification
- [x] Create test suite to cover basic validation logic
- [x] Implement appropriate Stringer interface method, add test case
- [x] Create associated keeper file for ContractSpecification.
  - [x] Get/Set/Remove methods for ContractSpecifications in state store (include events and instrumentation publishing).  Ensure remove method fails if there are any existing references to object.  Update any indexes that might be useful for rapidly querying/accessing data for queries.  Example: all scopes with a given `value_owner`
  - [ ] ~~Add a special Create method that takes a P8e ContractSpec and parses into ContractSpecification and RecordSpecifications.~~ Split out: #113 
  - [x] Stateful validation of object (pull existing, verify any updates are allowed, etc)
  - [x] Implement iterator using `KVStorePrefixIterator` and appropriate prefix
- [x] Wire up msg_server and query_server endpoints to new keeper methods. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
See #58 for details on the Record Specification identifier.

iramiller:
Moving out of assigned/todo as this is blocked pending merge/resolution of  #56

Issue #88, opened 2021-02-22
Add typed events and metrics to the metadata module


iramiller:
## Summary

Implement Typed Events for Metadata Module

### Typed Event Conversion
The SDK recently added [support for typed events](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-032-typed-events.md) and deprecated the existing style of kv event attributes.  This issue is for refactoring the existing events into proto messages and wiring up typed events to replace the existing ones.

- [ ] Replace Message Events with Typed Versions
- [ ] Ensure all Core Keeper functions that mutate state publish an associated typed event
  - [ ] Set
  - [ ] Remove

### Metric Publishing
Tracking the volume of information being handled on chain and the duration of requests is an important part of monitoring the health of the blockchain. [Metrics](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-013-metrics.md) were add.  Along with ensuring proper events are published the associated metrics should also be put in place.  These two efforts are linked because they typically are inserted at the same point in the code for similar reasons.

- [ ] Being Blocker methods are appropriately timed `defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)`
- [ ] All creation, modification, and validation methods are timed `defer telemetry.MeasureSince(time.Now(), "metadata", "validation|scope|specification", "get|set|validate|etc")`
- [ ] Creation of new records are counted `defer telemetry.IncrCounter(1, "new", "scope|record|specification")`
- [ ] Gauge metrics are set of appropriate amount values (examples, mint/burn amounts, send amounts)
- [ ] Appropriate labels are applied to each counter, gauge 

### Example Metric and Events from staking module msg_server
```
	if msg.Amount.Amount.IsInt64() {
		defer func() {
			telemetry.IncrCounter(1, types.ModuleName, "delegate")
			telemetry.SetGaugeWithLabels(
				[]string{"tx", "msg", msg.Type()},
				float32(msg.Amount.Amount.Int64()),
				[]metrics.Label{telemetry.NewLabel("denom", msg.Amount.Denom)},
			)
		}()
	}

	ctx.EventManager().EmitEvents(sdk.Events{
		sdk.NewEvent(
			types.EventTypeDelegate,
			sdk.NewAttribute(types.AttributeKeyValidator, msg.ValidatorAddress),
[...]

PR #106, opened 2021-02-24
Implement Contract Specification

dwedul-figure:
## Description

Implement the contract specification stuff (formerly known as group specification).

closes: #57, closes: #58

---

write a little note why.


x/metadata/types/address.go
```diff
@@ -318,35 +355,77 @@ func (ma MetadataAddress) Compare(other MetadataAddress) int {
...
-func (ma MetadataAddress) GroupUUID() (uuid.UUID, error) {
+// SecondaryUUID returns the secondary UUID from this MetadataAddress (if applicable).
+// More accurately, this converts bytes 18 to 33 (inclusive) to a UUID.
+func (ma MetadataAddress) SecondaryUUID() (uuid.UUID, error) {
 	if len(ma) < 1 {
 		return uuid.UUID{}, fmt.Errorf("address empty")
 	}
 	// if we don't know this type
-	if ma[0] > 0x04 {
+	if !ma.isTypeOneOf(allowedTypes...) {
```
iramiller, line 405: Not all of these have a second uuid (even when they have a length that is greater than 33).  When written the only MetadataAddress type with two uuids was the RecordGroup (Session) which indicates an object that is part of a scope and has its own uuid.  If the intention is to have a generic second uuid method which might be used by other id types (which?) then it should be scoped private since it currently doesn't provide appropriate context checks.

channa-figure (approved):
LGTM

x/metadata/types/specification.go
```diff
@@ -89,6 +91,121 @@ func (s ScopeSpecification) String() string {
...
+	}
+	if len(s.OwnerAddresses) == 0 {
+		return fmt.Errorf("invalid owner addresses count (expected > 0 got: %d)", len(s.OwnerAddresses))
+	}
+	for i, owner := range s.OwnerAddresses {
+		if _, err = sdk.AccAddressFromBech32(owner); err != nil {
+			return fmt.Errorf("invalid owner address at index %d: %w", i, err)
+		}
+	}
+	for i, owner := range s.OwnerAddresses {
```
iramiller, line 149: duplicate of above block

iramiller (changes requested):
Small concern about the metadata address secondary uuid function.

x/metadata/keeper/specification.go
```diff
@@ -7,31 +7,182 @@ import (
...
+		k.cdc.MustUnmarshalBinaryBare(it.Value(), &contractSpec)
+		if handler(contractSpec) {
+			break
+		}
+	}
+	return nil
+}
+
+// IterateContractSpecsForAddress processes all contract specs associated with an address using a given handler.
+func (k Keeper) IterateContractSpecsForAddress(ctx sdk.Context, address sdk.AccAddress, handler func(contractSpecID types.MetadataAddress) (stop bool)) error {
```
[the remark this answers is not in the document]
dwedul-figure: Works for me. Changed the func name to `IterateContractSpecsForAddress` and parameter name to `ownerAddress`. Made a similar change to `IterateScopeSpecsForAddress` for consistency.

x/metadata/types/address.go
dwedul-figure: Good catch. The change I made reflected the previous functionality, but that wasn't exactly accurate either. Should be fixed now.

x/metadata/types/specification.go
dwedul-figure: Just making doubly sure!!!

Fixed.

x/metadata/keeper/msg_server.go
```diff
@@ -60,61 +60,58 @@ func (k msgServer) ChangeOwnership(
...
+	if err := k.ValidateScopeUpdate(ctx, existing, msg.Scope, msg.Signers); err != nil {
 		return nil, err
 	}
 
-	k.SetScopeSpecification(ctx, *msg.Specification)
+	k.SetScope(ctx, msg.Scope)
 
 	ctx.EventManager().EmitEvent(
 		sdk.NewEvent(
-			sdk.EventTypeMessage,
```
iramiller, line 78: the `message` events are separate from the keeper events ... i.e. there is an event when a msg comes in, there is also an event when a scope is added... the two are separate.  There is a ticket to handle this work already #88 so there isn't a need to make changes here to resolve this issue.

dwedul-figure: I reordered the funcs in this file to match their ordering in the `tx.proto` `service Msg` section. So this chunk is kind of a mess as a diff.

iramiller (dismissed):
LGTM

iramiller (approved):

Issue #105, opened 2021-02-24
Decentralized discovery for object store instances


iramiller:
## Summary

A decentralized discovery method for object stores to connect to each other using the blockchain

## Problem Definition

Existing P8e object stores use a centralized system for connection and discovery.  With the transition to a decentralized public network this central relay point will no longer exist.  A method for an address to publish a list of endpoint(s) where partner object stores can authoritatively discover and connect is required.

## Proposal

Create a record on the blockchain that is controlled by an address and allows it to publish records indicating endpoints available for partners to connect to.

1. Add a new structure and store on chain under the owner address
2. Add appropriate read/write methods that allow the owner address to maintain a given entry
3. Provide a method that returns a list of records for a scope by linking against the "data_access" list of addresses that control which accounts the off-chain information should be provisioned for.

## Implementation

The following simple structure will be used to hold a reference for a locator endpoint for a given account.  The identified account will be the one that owns/controls the record and must sign requests to modify it.

```
message ObjectStoreEndpoint {
  // account address the endpoint is owned by
  string owner = 1;
  // locator endpoint uri
  string locator_uri = 2;
}
```

- [x] Create a new proto message for `ObjectStoreEndpoint`
- [x] Implement ValidateBasic for `ObjectStoreEndpoint`
- [x] Add NewObjectStoreEndpoint functions for creating an instance of `ObjectStoreEndpoint`
- [x] Create test suite to cover basic validation logic
- [x] Implement appropriate Stringer interface method, add test case
- [x] Create associated keeper file for `ObjectStoreEndpoint`.
  - [x] Create a byte code in the `keys.go` file for the new type, add methods to create a kvstore key against the address of the user appended to this new type type byte.
  - [x] Get/Set/Remove methods for object in state store (include events and instrumentation publishing).
  - [x] Special query method that returns all of the registered endpoints for a scope / data_access list
- [x] Wire up msg_server and query_server endpoints to new keeper methods. 
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

iramiller:
This issue/proposal should be reviewed by @scirner22 for object store implementation requirements as well as @rjmarkel for security sign-off.

scirner22:
[...]

PR #150, opened 2021-03-09
Feature/dec obj store

arnabmitra:
## Description

closes: #105 

---

write a little note why.


arnabmitra:
still working on up'ing the codev  🤷  but thought i would put up the PR for review anyways

x/metadata/keeper/msg_server.go
```diff
@@ -411,3 +413,112 @@ func (k msgServer) DeleteRecordSpecification(
...
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+
+	if !k.Keeper.OSLocatorExists(sdkCtx, msg.Locator.Owner) {
+		sdkCtx.Logger().Error("Address not already bound to an URI", "owner", msg.Locator.Owner)
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, types.ErrOSLocatorAlreadyBound.Error())
+	}
+
+	// already valid address, checked in ValidateBasic
+	ownerAddr, _ := sdk.AccAddressFromBech32(msg.Locator.Owner)
```
dpederson-figure, line 467: Ignored error?  At least add a comment why this error isn't returned.

x/metadata/keeper/objectstorelocator.go
```diff
@@ -0,0 +1,216 @@
...
+	key, err := types.GetOsLocatorAddressKeyPrefix(ownerAddress)
+	if err != nil {
+		return types.ObjectStoreLocator{}, false
+	}
+	store := ctx.KVStore(k.storeKey)
+	b := store.Get(key)
+	if b == nil {
+		return types.ObjectStoreLocator{}, false
+	}
+	k.cdc.MustUnmarshalBinaryBare(b, &osLocator)
```
dpederson-figure, line 24: cool with panics here?

channa-figure (commented):
Added comments to sections

x/metadata/keeper/objectstorelocator.go
arnabmitra: great catch @dpederson-figure , no just a rookie mistake from me.

proto/provenance/metadata/v1/query.proto
```diff
@@ -271,3 +308,50 @@ message RecordSpecificationByIDResponse {
   RecordSpecification record_specification    = 1 [(gogoproto.moretags) = "yaml:\"record_specification\""];
   string              record_specification_id = 2 [(gogoproto.moretags) = "yaml:\"record_specification_id\""];
 }
+
+// OSLocationRequest is used for requesting a Objectstore location by name
```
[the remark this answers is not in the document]
arnabmitra: fixing.. my bad

proto/provenance/metadata/v1/tx.proto
```diff
@@ -47,6 +48,15 @@ service Msg {
   rpc AddRecordSpecification(MsgAddRecordSpecificationRequest) returns (MsgAddRecordSpecificationResponse);
   // DeleteRecordSpecification deletes a record specification
   rpc DeleteRecordSpecification(MsgDeleteRecordSpecificationRequest) returns (MsgDeleteRecordSpecificationResponse);
+
+  // Os Locator
+  // BindOSLocator binds a owner address to a uri.
+  rpc BindOSLocator(MsgBindOSLocatorRequest) returns (MsgBindOSLocatorResponse);
+
+  // DeleteName defines a method to verify a particular invariance.
```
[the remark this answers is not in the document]
arnabmitra: Fixed

proto/provenance/metadata/v1/tx.proto
```diff
@@ -283,3 +293,44 @@ message MsgDeleteRecordSpecificationRequest {
...
+message MsgBindOSLocatorResponse {
+  ObjectStoreLocator locator = 1 [(gogoproto.nullable) = false];
+}
+
+// MsgModifyOSLocatorResponse defines the Msg/ModifyObjectLocator response type
+message MsgModifyOSLocatorResponse {
+  ObjectStoreLocator locator = 1 [(gogoproto.nullable) = false];
+}
+
+// MsgDeleteNameRequest defines an sdk.Msg type that is used to remove an existing address/name binding.  The binding
```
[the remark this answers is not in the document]
arnabmitra: fixed..i suck..thanks for catching these

x/metadata/keeper/keeper.go
```diff
@@ -182,3 +196,24 @@ func (k Keeper) ValidatePartiesInvolved(parties []types.Party, requiredParties [
 	}
 	return nil
 }
+
+// ResolvesTo to determines whether a name resolves to a given address.
```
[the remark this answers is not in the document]
arnabmitra: fixed.

x/metadata/keeper/msg_server.go
```diff
@@ -411,3 +413,112 @@ func (k msgServer) DeleteRecordSpecification(
...
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+	}
+	// Emit event and return
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeOsLocatorCreated,
+			sdk.NewAttribute(types.AttributeKeyOSLocatorAddress, msg.Locator.Owner),
+			sdk.NewAttribute(types.AttributeKeyOSLocatorURI, msg.Locator.LocatorUri),
+		),
+	)
```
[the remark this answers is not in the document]
arnabmitra: In this case the owner is the sender, unlike in a contract being executed, imo it is ok for different func/methods to emit events.. but idk.. i would lie typed events..i am ok to change this.. btw i based this a lot on the name module and that follows the ^^ convention  so not sure..

dpederson-figure (dismissed):
Very well written. Nice work @arnabmitra!

dpederson-figure (approved):

x/metadata/keeper/keeper.go
```diff
@@ -85,6 +86,15 @@ type MetadataKeeperI interface {
 	IterateRecordSpecsForContractSpec(ctx sdk.Context, contractSpecID types.MetadataAddress, handler func(recordSpecID types.MetadataAddress) (stop bool)) error
 	// GetRecordSpecificationsForContractSpecificationID returns all the record specifications associated with given contractSpecID
 	GetRecordSpecificationsForContractSpecificationID(ctx sdk.Context, contractSpecID types.MetadataAddress) ([]*types.RecordSpecification, error)
+
+	// GetOSLocator returns the OS locator records for a given name record.
```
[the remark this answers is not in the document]
arnabmitra: thanks!

x/metadata/keeper/query_server.go
```diff
@@ -15,8 +18,13 @@ import (
 	"github.com/provenance-io/provenance/x/metadata/types"
 )
 
+const defaultLimit = 100
+
 var _ types.QueryServer = Keeper{}
 
+// NameRecords within the GenesisState
```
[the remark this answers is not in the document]
arnabmitra: fixed

proto/provenance/metadata/v1/tx.proto
```diff
@@ -283,3 +293,44 @@ message MsgDeleteRecordSpecificationRequest {
...
+
+// MsgModifyOSLocatorRequest defines an sdk.Msg type that is used to modify a uri associated with an owner address.
+message MsgModifyOSLocatorRequest {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+  // The object locator to bind the address to bind to the URI.
+  ObjectStoreLocator locator = 1 [(gogoproto.nullable) = false];
+}
+
+// MsgAddOSLocatorResponse defines the Msg/BindObjectLocator response type.
```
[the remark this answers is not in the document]
arnabmitra: fixed

iramiller:
Might be good for @scirner22 to quick scan the query endpoints in this PR to make sure they cover the P8e teams needs... seems like they are quite comprehensive but doesn't hurt to confirm. Looking solid overall.

x/metadata/keeper/keeper.go
```diff
@@ -182,3 +196,24 @@ func (k Keeper) ValidatePartiesInvolved(parties []types.Party, requiredParties [
 	}
 	return nil
 }
+
+// VerifyCorrectOwner to determines whether the signer resolves to the owner of the OSLocator record.
+func (k Keeper) VerifyCorrectOwner(ctx sdk.Context, ownerAddr sdk.AccAddress) bool { // nolint:interfacer
```
scirner22, line 201: This is one thing I was going to ask about. How does provenance know I'm only pushing public keys/addresses that I have the associated private key for? By reading the doc line it seems like this function is checking if the tx was signed by the address present in the request proto? Right now a p8e instance has one provenance account and signs and pays fees on behalf of possibly many internal key pairs. Its these public keys that are present on scopes and not the provenance account that's signing and sending Txs.

iramiller: For this specific endpoint the answer is straight forward.  The request must be signed by the address the record is for.  A notary could not sign for this transaction.  There is a need to review the larger signing approach for the metadata module though.  

All of the APIs added have used the standard signing methods supported by the SDK (not the independent signature/blockchain notary pattern).  This was done because there are actually several different features that are available for blockchain signatures that overlap with our needs and this module should strive to implement/use blockchain native approaches.  Some of these include module/smart contract accounts, multisig, fee payer/delegation, etc.  The P8e specific memorialize contract is the only endpoint that supports current p8e notary/independent account signature method.

scirner22: That makes sense. How does that look from an implementation perspective? Say there's a p8e key pairs {A, B, C} that are already configured and exist. A user goes into the p8e ui and creates a new key pair. The key pair is saved in p8e and object-store now we need to submit this request to provenance. This new keypair D will sign the tx and add it's public key as the address in the request.

Does provenance complain about that request? Is anything needed to first establish keypair D as a provenance account? Where does the fees get paid from?

arnabmitra: @scirner22 since the keypair(ne one) is assumed to be a valid address/account on chain, i mean this is moot now since the keypair has to pay fee's for the tx so cannot not exist to establish an object locator record, but pointing it out since there is a concept of delegated fee payer coming in the future(afaik)..having said that i can take out the check if that is the consensus 🤷

proto/provenance/metadata/v1/query.proto
```diff
@@ -88,11 +89,42 @@ service Query {
...
+  }
+
+  // for querying oslocator by address.
+  rpc OSLocator(OSLocatorRequest) returns (OSLocatorResponse) {
+    option (google.api.http).get = "/provenance/metadata/v1/locator/{owner}";
+  }
+
+  // for querying oslocator by uri.
+  rpc OSLocatorByURI(OSLocatorByURIRequest) returns (OSLocatorByURIResponse) {
+    option (google.api.http).get = "/provenance/metadata/v1/locator/uri/{uri}";
```
iramiller, line 106: this seems like a bad idea since you have a uri in a uri here ... do we need this query endpoint at all? might be best to just remove it and save future headaches.

x/metadata/keeper/objectstorelocator.go
```diff
@@ -0,0 +1,220 @@
+package keeper
```
iramiller, line 1: the name of this file should be consistent with the types and the proto file names.

x/marker/types/marker.pb.go
```diff
@@ -613,7 +613,10 @@ func (m *Params) Unmarshal(dAtA []byte) error {
 			if err != nil {
 				return err
 			}
-			if (skippy < 0) || (iNdEx+skippy) < 0 {
+			if skippy < 0 {
```
iramiller, line 616: These changes in this PR seem to indicate that your installed proto tools do not match the versions everyone else is using ... that means that these *.pb.go files will change constantly as different team members make commits.  Please verify your proto tools are installed using the make target for them ... and that your PR does not change the `pb.go` files that are not part of you PR/associated with `*.proto` changes.

iramiller (changes requested):
All those extra `pb.go` files are changed due to a mismatch in proto tools generating from the `make proto-gen` target... Please update/reinstall and regen/submit.

x/metadata/keeper/keeper.go
iramiller: We really really need the `authz` module from upstream that supports the "feepayer" concept to make a lot of the Figure use cases much easier to deal with.  Right now we have to do a bunch of fund shuffling to ensure the accounts that need to sign have the coin to pay the fees.

proto/provenance/metadata/v1/query.proto
arnabmitra: {uri} is just the path param,  i can remove the endpoint, was thinking a client might want to query all the addresses it has pointing to it's object store. 🤷

x/metadata/keeper/objectstorelocator.go
arnabmitra: good point..changed

x/marker/types/marker.pb.go
arnabmitra: what version should i be using ..yes i can revert this  

protoc --version
libprotoc 3.14.0

x/metadata/keeper/objectstorelocator.go
arnabmitra: changed

x/metadata/keeper/keeper.go
scirner22: I think this check is what we want because without it we would need some other type of check to prove your submitting public keys that you own the private key for, but I'm trying to understand what prereq steps we will need to take for every new keypair now.
Is there an endpoint to hit to establish a new address/account on chain? It does seem like what you mentioned is exactly what p8e would want though. The ability for accounts to sign transactions but have the sole figure p8e account to fund the fee.

iramiller: (note: its actually the feegrant module we want for this example, the authz module would let one account sign for another which I guess also applies here).

arnabmitra: right now with the state of current code base, fee's 
* would get paid from the signer
i.e the keypair that you established(and the current user setting up their locator)
 when you establish the keypair in an app/ui, transfer them hash to pay fees(from probably figure tech or provenance accnt, but not sure on that), 
that would
1. establish their account on chain
2. give them fees to establish their os locator record.

iramiller: I think this will be the rough flow even with fee payer.. with the payer option though instead of sending the new account hash the money account would create a grant allowing the target account to spend their funds for fees.

x/marker/types/marker.pb.go
arnabmitra: after much back and forth, i think my code-gen is same as others

dpederson-figure (dismissed):

x/metadata/types/msg.go
```diff
@@ -672,3 +679,141 @@ func (msg MsgDeleteRecordSpecificationRequest) ValidateBasic() error {
...
+}
+func (msg MsgDeleteOSLocatorRequest) Route() string {
+	return ModuleName
+}
+
+func (msg MsgDeleteOSLocatorRequest) Type() string {
+	return TypeMsgDeleteOSLocatorRequest
+}
+
+func (msg MsgDeleteOSLocatorRequest) ValidateBasic() error {
```
iramiller, line 734: lots of code coverage warnings for this file ... these could be mitigated pretty quickly with a couple more basic tests added to the `msg_test.go` file

x/metadata/types/oslocatorparams.go
```diff
@@ -0,0 +1,49 @@
+package types
```
iramiller, line 1: this file also should have a corresponding `_test.go` file

iramiller (commented):
This is looking great... I do think it would be worth adding a bit of unit testing to the files added/modified in the `types` folder (msg.go, objectstore.go) because there is quite a bit of validation logic for those messages in there.  The unittests are super easy to work with for stuff in the types folder because you don't have to setup a full in process node to test them.

x/metadata/keeper/keeper.go
scirner22: Awesome. Thanks for working on this @arnabmitra . I'll probably reach out when I'm ready to start trying this. I'm probably a full week away from that point.

channa-figure (approved):
Nice work!

x/metadata/types/oslocatorparams.go
arnabmitra: image
 added

iramiller (approved):

iramiller (dismissed):
Great PR!

x/metadata/types/msg.go
arnabmitra: did add some for this too.. looks green'ish  on new code

iramiller (approved):

Issue #113, opened 2021-02-26
Add ability to convert p8e ContractSpec to new ContractSpecification and RecordSpecifications.


dwedul-figure:
## Summary

Add a special Create method that takes a P8e ContractSpec and parses into ContractSpecification and RecordSpecifications.

## Problem Definition

The P8e system generates a complete contract specification using a Kotlin class as a starting point during a build process.  This contract specification output contains a full collection of inputs used by the class as well as a collection of consideration/constraints.  The type formats are all slightly different than an on chain ContractSpecification as P8e uses a set of nested, repeated Definitions for several different purposes.

The `legacy/v040/migrate.go` file contains an implementation that maps from a legacy v039 contractspec to the current ContractSpecification and RecordSpecifications.  A discussion with the P8e team on how they wish to support recording the spec going forward is required.  If they are willing to change their format to match the current one then no further work is required.  If the `v0` format is required then a new AddP8eContractSpec method will need to be created that take the v0 format objects and processes them into the current form.

- [ ] Add a special Create method that takes a P8e ContractSpec and parses into ContractSpecification and RecordSpecifications.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
An implementation for this process exists in the legacy migration functions.
https://github.com/provenance-io/provenance/blob/401c6d8acba9e1cf85dc7ec0a2e56804b816f949/x/metadata/legacy/v040/migrate.go#L354

piercetrey-figure:
After discussing with the @provenance-io/p8e team, yes this is indeed still something that we will need as there would be too many changes needed on our side to get done in the near-term

iramiller:
> After discussing with the @provenance-io/p8e team, yes this is indeed still something that we will need as there would be too many changes needed on our side to get done in the near-term

That was never really the question I was looking to have @channa-figure get answered ... the part we need to know/ensure is that the implementation of this feature lines up with what p8e is expecting.

scirner22:
I'm confused what this issue relates to. I thought the last pbc release contained everything p8e needed to migrate from simple client to pbc 2.0 grpc?

iramiller:
All the base types and specification components are implemented and there are grpc query endpoints for reading/writing these various pieces as well as validation/integration tests covering their implementations.  If P8e were starting on integrating with the blockchain the endpoints needed are there.

[...]

PR #167, opened 2021-03-12
Convert p8e contract spec

channa-figure:
## Description

Add a temporary addcontractspec grpc endpoint to support adding a v39 `ContractSpec` as a v40 `ContractSpecification`.  

closes: #113 

---

write a little note why.


x/metadata/types/msg.go
```diff
@@ -423,6 +425,46 @@ func (msg MsgAddScopeSpecificationRequest) ValidateBasic() error {
...
+// GetSignBytes gets the bytes for the message signer to sign on
+func (msg MsgAddP8EContractSpecRequest) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&msg))
+}
+
+// ValidateBasic performs a quick validity check
+func (msg MsgAddP8EContractSpecRequest) ValidateBasic() error {
+	if len(msg.Signers) < 1 {
+		return fmt.Errorf("at least one signer is required")
+	}
```
iramiller, line 464: the internal request should be parsed into the new types here and each of those should have their validatebasic called.

x/metadata/keeper/msg_server.go
```diff
@@ -411,3 +411,49 @@ func (k msgServer) DeleteRecordSpecification(
 
 	return &types.MsgDeleteRecordSpecificationResponse{}, nil
 }
+
+func (k msgServer) AddP8EContractSpec(
+	goCtx context.Context,
+	msg *types.MsgAddP8EContractSpecRequest,
+) (*types.MsgAddP8EContractSpecResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	newspec, newrecords, err := types.ConvertP8eContractSpec(&msg.Contractspec, msg.Signers)
```
iramiller, line 420: These output specs, records specs are not validated with validatebasic (see other comment about adding this conversion step to the validatebasic flow for this request

iramiller (commented):
Looks like a solid start

x/metadata/types/msg.go
dwedul-figure: That seems pretty wasteful to do here. Whenever we have a MsgAddP8EContractSpecRequest we're going to want to convert it. But the pattern we have is to have ValidateBasice() just return an error. So we'd be converting it, checking the converted pieces, then throwing them away for the error only to need to convert them again later.

In my opinion, we'd just want really superficial validation here, then we'd do the conversion work and validate all the converted things later on.

dwedul-figure: ... of course, after I say all that, I look at my own code and realize that converting the msg and throwing away the converted pieces is exactly what I was doing myself. So who knows.

x/metadata/keeper/msg_server.go
channa-figure: I added more validations in the convert method and this code is also exercised in the messages validate basic.

x/metadata/types/msg.go
channa-figure: I was thinking it would be a bit much too, but @iramiller made a good point about gas usage.

iramiller: It is worth keeping in mind that ValidateBasic checks are performed in a different part of the request flow than the message handler ... they are called in the CheckTx step which is ran whenever a transaction is added to the mem pool.. The purpose of this check is to prevent invalid transaction from hitting the inside of a block ...

The resources used if you let this go into DeliverTx and it gets written into a block as an invalid transaction are infinitely higher than just some extra cpu cycles for processing that occurs here.  If it goes to the next step it gets recorded on the chain forever.

dpederson-figure (dismissed):
lgtm

iramiller (approved):
LGTM

PR #196, opened 2021-03-23
CLI Add/Update/Remove Contract Specification

channa-figure:
## Description

Add cli command for Add/Update/Remove operations on `ContractSpecification`

closes: #195 
---

write a little note why.


x/metadata/client/cli/tx.go
```diff
@@ -242,3 +246,148 @@ func ModifyOsLocatorCmd() *cobra.Command {
...
+			specificationID, err = types.MetadataAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+
+			if !specificationID.IsContractSpecificationAddress() {
+				return fmt.Errorf("invalid contract specification id: %s", args[0])
+			}
+
+			msg := types.NewMsgDeleteContractSpecificationRequest(specificationID, strings.Split(args[1], ","))
```
iramiller, line 381: Does this signers approach work?  Have you tested this command against a localnet-start instance for example?

iramiller (commented):
This seems good ... will approve if these commands have been tested successfully against the localnet or `make run` instances.

x/metadata/client/cli/tx.go
channa-figure: Updated it to user a signers flag if present.

x/metadata/client/cli/tx.go
```diff
@@ -242,3 +250,163 @@ func ModifyOsLocatorCmd() *cobra.Command {
...
+	cmd.Flags().String(FlagSigners, "", "comma delimited list of bech32 addresses")
+}
+
+func parseSigners(cmd *cobra.Command, client *client.Context) []string {
+	flagSet := cmd.Flags()
+	if flagSet.Changed(FlagSigners) {
+		signerList, _ := flagSet.GetString(FlagSigners)
+		return strings.Split(signerList, ",")
+	}
+	return []string{client.GetFromAddress().String()}
```
iramiller, line 342: This could alternatively remove the from address from the signers (if present) and insert the from address as the first address in the list... that would support use of both flags together.  If thats not done then you would need an exclusive switch to prevent an invalid operation where it signs with potentially the wrong key

iramiller (approved):
Overall this looks good.  I suspect there could be a confusing interaction with the `--from` and `--signers` flags though.  Optional solution to that noted in review.

Issue #208, opened 2021-03-26
Detailed Denom Metadata restrictions


iramiller:
## Summary

Denom metadata is intended to offer the ability to associate different display values and exponents with a given denom.  The current denom metadata function has only minimal validation/restriction on allowed values.

## Problem Definition

A malicious user could create a denom display value for their coin that attempts to impersonate another coin.

## Proposal

Require that all denoms entered in the denommetadata consist of the base marker denom value and some sort of official SI unit prefix in full or using its abbreviation.  https://physics.nist.gov/cuu/Units/prefixes.html

For example: `nhash` or `nanohash` when used as a base unit means that adding `hash` would be allowed and its exponent must be 9 as it is removing the `nano/n` prefix. `Mhash` would be allowed (or `megahash`) for a million hash if it included the `abs(-9)+6 == 15` exponent.

`1,000,000.000000000hash | 1,000,000,000,000,000nhash | 1Mhash`

```go
// DenomUnit represents a struct that describes a given
// denomination unit of the basic token.
message DenomUnit {
  // denom represents the string name of the given denom unit (e.g uatom).
  string denom = 1;
  // exponent represents power of 10 exponent that one must
  // raise the base_denom to in order to equal the given DenomUnit's denom
  // 1 denom = 1^exponent base_denom
  // (e.g. with a base_denom of uatom, one can create a DenomUnit of 'atom' with
  // exponent = 6, thus: 1 atom = 10^6 uatom).
  uint32 exponent = 2;
  // aliases is a list of string aliases for the given denom
  repeated string aliases = 3;
}

// Metadata represents a struct that describes
// a basic token.
message Metadata {
  string description = 1;
  // denom_units represents the list of DenomUnit's for a given coin
  repeated DenomUnit denom_units = 2;
  // base represents the base denom (should be the DenomUnit with exponent = 0).
  string base = 3;
  // display indicates the suggested denom that should be
  // displayed in clients.
  string display = 4;
}
```
Given the above metadata structure currently only the base value is checked (must match the marker denom) and the marker must be in the proposed status.

### Implementation 

[...]

PR #213, opened 2021-03-29
Restrict denom metadata

dwedul-figure:
## Description

Restrict denom metadata.

closes: #208

---

write a little note why.


proto/provenance/marker/v1/tx.proto
```diff
@@ -150,10 +154,39 @@ message MsgTransferResponse {}
...
-      [(gogoproto.nullable) = false, (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/x/bank/types.Metadata"];
+      [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/x/bank/types.Metadata"];
+  // administrator is the address of the administrator making this request.
   string administrator = 2;
+  // description is A description of this denomination.
+  // It is used to populate the Metadata.description field.
+  string description = 3;
+  // base_name is the root of the name of this denomination. SI Prefix strings will be added to this to create the
+  // desired denom units.
+  string base_name = 4;
```
iramiller, line 184: This seems like it would be the marker denom itself (as the base_name) and yet the problem with a user does not think to create their denom with the smallest unit of measurement in mind will exist.

x/marker/keeper/marker.go
```diff
@@ -562,21 +562,27 @@ func (k Keeper) TransferCoin(ctx sdk.Context, from, to, admin sdk.AccAddress, am
...
 	}
-	m, err := k.GetMarkerByDenom(ctx, metadata.Base)
-	if err != nil {
-		return fmt.Errorf("marker not found for %s: %w", metadata.Base, err)
+	marker, markerErr := k.GetMarkerByDenom(ctx, metadata.Base)
+	if markerErr != nil {
+		return fmt.Errorf("marker not found for %s: %w", metadata.Base, markerErr)
 	}
-	if !m.GetManager().Equals(caller) && !m.AddressHasAccess(caller, types.Access_Admin) {
+	if !marker.GetManager().Equals(caller) && !marker.AddressHasAccess(caller, types.Access_Admin) {
```
iramiller, line 574: manager is only set on proposed/finalized markers ... this value is cleared and the permissions list takes over for active markers this check should be structured as an OR not an AND

dwedul-figure: The only thing changed on this line is the variable name `m` was changed to `marker` because it was being used several lines down.

I thought about switching it to an or as well, but then I noticed that the equals are negated. So this line as it is returns an error if the caller is neither the manager nor an admin.

iramiller: Ah... serves me right for reviewing in a hurry.

iramiller (approved):
LGTM.

Issue #205, opened 2021-03-26
CLI Add/Update/Remove Record 


channa-figure:
## Summary

Add cli command for Add/Update/Remove operations on `Record`. 

## Proposal

Add cli command for Add/Update/Remove operations. 

Notes: 

requires a recordspecification id which in turn includes a contract specification, optional session_id, if not given would create a new session using the associated contract specification. This will fail if the scope has a scopespecification that does not include the provided contractspecification as an authorized method of maintenance. It must be possible to specify the array of parties in order to satisfy the contractspecification

`Session` will not have a delete command.  Deleting of session will happen where scopes are removed and there is no longer a reference to it.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #214, opened 2021-03-29
Cli write remove records

channa-figure:
## Description

Add cli add/update/delete records

closes: #205
---

write a little note why.


x/metadata/client/cli/tx.go
```diff
@@ -371,21 +372,215 @@ func WriteContractSpecificationCmd() *cobra.Command {
...
+		Long: fmt.Sprintf(`Add/Update metadata record to the provenance blockchain.
+[scope-id]                 - scope metaaddress for the record
+[record-spec-id]           - associated record specification metaaddress
+[name]                     - record name
+[process]                  - comma delimited structure of process name, id (hash or bech32 address), and method: Example: processname,hashvalue,method
+[inputs]                   - semicolon delimited list of input structures.  Example: name,soure-value(hash or metaaddress),typename,status(proposed,record);...
+[outputs]                  - semicolon delimited list of outputs structures. Example: hash-value,status(pass,skip,fail);...
+[parties-involved]         - semicolon delimited list of party structures(address,role). Accepted roles: originator,servicer,investor,custodian,owner,affiliate,omnibus,provenance
+[contract-id | session-id] - a contract or session id. 
+[contract-name]            - contract name must be set if using contract id to create a new session
```
iramiller, line 392: this could probably come from the contract-specification that was referenced from `contract-id` above

x/metadata/client/cli/cli_test.go
```diff
@@ -1410,6 +1410,7 @@ func runTxCmdTestCases(s *IntegrationCLITestSuite, testCases []txCmdTestCase) {
 				require.EqualError(t, err, tc.expectErrMsg, "%s expected error message", cmdName)
 			} else if tc.expectErr {
 				require.Error(t, err, "%s expected error", cmdName)
+				require.EqualError(t, err, tc.expectErrMsg)
```
dwedul-figure, line 1413: At this point, you know that `tc.expectErrMsg` is `""`.

x/metadata/client/cli/tx.go
```diff
@@ -309,24 +313,24 @@ func WriteContractSpecificationCmd() *cobra.Command {
...
-[owners] - comma delimited list of bech32 owner addresses
-[parties-involved] - comma delimited list of party types.  Accepted values: originator,servicer,investor,custodian,owner,affiliate,omnibus,provenance
-[source-value] - source identifier of type hash or resourceid
-[classname] - Name of contract specification
-[description-name] - optional- description name identifier
-[description] - optional - description text
-[website-url] - optional - address of website
-[icon-url] - optional - address to a image to be used as an icon
-		`,
+contractspec-id]   - contract specification metaaddress
```
dwedul-figure, line 316: Looks like a '[' got chopped off the beginning.
```suggestion
[contractspec-id]   - contract specification metaaddress
```

x/metadata/client/cli/cli_test.go
channa-figure: No, the error message is set in the test as a string.  I'm testing that the cli command validates the data correctly before broadcasting it to the message server.  I find this valuable to assure it fails were I expected it to fail.

x/metadata/client/cli/tx.go
channa-figure: Thanks, fixed it.

x/metadata/client/cli/tx.go
```diff
@@ -371,21 +372,215 @@ func WriteContractSpecificationCmd() *cobra.Command {
...
+				return err
+			}
+			var sessionID types.MetadataAddress
+			var writeSessionMsg *types.MsgWriteSessionRequest
+			switch {
+			case contractOrSessionID.IsSessionAddress():
+				record.SessionId = contractOrSessionID
+			case contractOrSessionID.IsContractSpecificationAddress():
+				if len(args) != 9 {
+					return fmt.Errorf("contract name must be provided")
```
iramiller, line 458: again ... could pull this from the contract spec if it was looked up

x/metadata/client/cli/tx.go
```diff
@@ -406,7 +601,7 @@ $ %s tx metadata recspec1qh... recordname inputname1,typename1,hashvalue;inputen
 				return err
 			}
 
-			resultType := definitionType(args[4])
+			resultType := types.DefinitionType(types.DefinitionType_value[fmt.Sprintf("DEFINITION_TYPE_%s", strings.ToUpper(args[4]))])
```
iramiller, line 607: lots of chance for stuff to go wrong in this like ... that map lookup could be checked for an invalid result... on the other hand if this simply blows up and returns to the console that may be sufficient

x/metadata/client/cli/cli_test.go
dwedul-figure: Four lines up, there is `if len(tc.expectErrMsg) > 0 {` which does the `EqualError` check. So if it gets to this `else if tc.expectErr`, you know that len(tc.expectErrMsg) is zero.

There's a good chance we could completely remove the `tc.expectErr` flag from these tests. I was going to, but since we were both in there making changes last week, I figured that'd be too much of a merge pain. It's only there when we want to make sure an error is returned, but we don't care what the error is.

iramiller (commented):
We shouldn't worry about the singers fields being the wrong type right now as that is a breaking api change ... we do need to address that very quickly though before we get a lot of users integrated with the spec.

x/metadata/client/cli/cli_test.go
dwedul-figure: Also, there is no reason to ever call `Error` followed by `EqualError`.  The `EqualError` check includes making sure that the provided error isn't nil.

channa-figure: Ah, I see what you mean with it being checked above.  Totally, missed that.  However, on the note about caring what the error message is, I agree that we really don't care what the error message itself is.  It is more for checking that it failed in the place I expected.  I find it valuable and have caught many issues when using it.

x/metadata/client/cli/tx.go
channa-figure: I was hoping that was the case when I first wrote it.  However, when we are doing a write session there is a validation of Session.name == contractspec.classname (https://github.com/provenance-io/provenance/blob/b29880c08e6ff856fc681ed7c83c13ca520c2483/x/metadata/keeper/session.go#L148).  I was trying to avoid the need to write a query for contract spec information in the cli command.

iramiller: > when we are doing a write session there is a validation of Session.name == contractspec.classname 
Kind of an odd situation then ... we have enough info to determine what the correct answer is for the user but we make them give us the correct answer anyway...
Is it reasonable to do a query in a cli method? I can't say I have tried that before...

x/metadata/client/cli/tx.go
iramiller: Validate basic should error on unspecified as it is not a valid option to use ...

x/metadata/client/cli/cli_test.go
dwedul-figure: 99% of the test setups we have that include both an expected error message as well as a boolean to indicate whether or not we expect an error can be simplified by removing the boolean and using the existence of a non-empty error string to indicate that we expect the test to fail. That's my point about the `tc.expectErr` flag there. If it's always true when `tc.expectErrMsg` is not empty, then it's not needed. The only time we would want such a flag in the test case setup is when we want to run a test that we expect will give an error, but we don't want to define the expected error.

But if `tc.expectErrMsg` is not empty, then we want to make sure that a) an error was generated, and b, it equals the provided string. Both of these are done with `EqualError`. If `tc.expectErrMsg` is an empty string, but `tc.expectErr` is true, then we expect an error, but don't care about the message. So we should use `Error` and we should NOT call `EqualError` because we don't have the expected error message to compare against.

In fact, if your tests are passing with this line there, it shows that there isn't a single test case that has `len(tc.expectErrMsg) == 0 && tc.expectErr == true`. Otherwise the `EqualError` test would fail because you're comparing the error's message to an empty string, and it's highly unlikely that the error's message is an empty string.

x/metadata/client/cli/tx.go
channa-figure: Yeah, well let me give it a try to do a query within the tx.  I wasn't sure if we wanted to do that, since I haven't seen it in our modules yet.

x/metadata/client/cli/cli_test.go
channa-figure: Sorry, totally misunderstood what you were trying to say in your comment (thought you were saying we shouldn't check what the error message is).  Yes, this made sense when I took a deeper look at where my mistake was.

x/metadata/client/cli/tx.go
iramiller: > Is it reasonable to do a query in a cli method? I can't say I have tried that before...
based on looking at how query flags are handled in the CLI I would say there is a good reason why we have not seen this done before... it is not really a great idea because of the root level separation in the cli between queries and tx.

x/metadata/keeper/session.go
```diff
@@ -145,8 +145,8 @@ func (k Keeper) ValidateSessionUpdate(ctx sdk.Context, existing *types.Session,
 		return fmt.Errorf("cannot find contract specification %s", proposed.SpecificationId)
 	}
 
-	if proposed.GetName() != contractSpec.GetClassName() {
-		return fmt.Errorf("proposed name does not match contract spec. expected %s, got %s)", proposed.GetName(), contractSpec.GetClassName())
+	if len(proposed.GetName()) == 0 && existing == nil {
```
iramiller, line 148: seems like a good fix for this

iramiller (approved):
LGTM

iramiller (approved):

Issue #188, opened 2021-03-22
Remove deprecated ModuleCdc amino encoding from attribute module


iramiller:
## Summary of Bug

The attribute module is currently using ModuleCdc (Amino) encoding for storage in the KV store.  This use is deprecated in favor of the Binary Marshaller (proto) format.

## Mitigation

- [x] Refactor Marshall/Unmarshall methods to use the `cdc` binary encoder
```go
x/attribute/keeper/keeper.go:
   88  		record := types.Attribute{}
   89: 		if err := types.ModuleCdc.UnmarshalBinaryBare(iterator.Value(), &record); err != nil {
   90  			return err

  132  	// Store the sanitized account attribute
  133: 	bz, err := types.ModuleCdc.MarshalBinaryBare(&attr)
  134  	if err != nil {

  158  		attr := types.Attribute{}
  159: 		if err := types.ModuleCdc.UnmarshalBinaryBare(it.Value(), &attr); err != nil {
  160  			return err

  184  		attr := types.Attribute{}
  185: 		if err = types.ModuleCdc.UnmarshalBinaryBare(it.Value(), &attr); err != nil {
  186  			return

  213  	// Store the sanitized account attribute
  214: 	bz, err := types.ModuleCdc.MarshalBinaryBare(&attr)
  215  	if err != nil {
```
- [x] Create a migration function that extracts any existing attributes encoded in amino and reloads them using proto encoding. The use of the export/import genesis methods is one possible way to do this.
____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #217, opened 2021-03-31
Bug/attribute amino cdc

arnabmitra:
## Description

closes: #188 

---

write a little note why.

### DIFF
--- a/x/attribute/keeper/keeper.go
+++ b/x/attribute/keeper/keeper.go
@@ -88,7 +88,7 @@ func (k Keeper) IterateRecords(ctx sdk.Context, prefix []byte, handle Handler) e
 	for ; iterator.Valid(); iterator.Next() {
 		record := types.Attribute{}
 		// get proto objects for legacy prefix with legacy amino codec.
-		if bytes.Equal(types.AttributeKeyPrefix, prefix) {
+		if bytes.Equal(types.AttributeKeyPrefixAmino, prefix) {
 			if err := types.ModuleCdc.UnmarshalBinaryBare(iterator.Value(), &record); err != nil {
 				return err
 			}
--- a/x/attribute/keeper/keeper.go
+++ b/x/attribute/keeper/keeper.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


x/attribute/keeper/keeper.go
```diff
@@ -78,20 +79,24 @@ func (k Keeper) GetAttributes(ctx sdk.Context, acc sdk.AccAddress, name string)
...
-	iterator := sdk.KVStorePrefixIterator(store, types.AttributeKeyPrefix)
+	iterator := sdk.KVStorePrefixIterator(store, prefix)
 	defer iterator.Close()
 	// Iterate over records, processing callbacks.
 	for ; iterator.Valid(); iterator.Next() {
 		record := types.Attribute{}
-		if err := types.ModuleCdc.UnmarshalBinaryBare(iterator.Value(), &record); err != nil {
-			return err
+		// get proto objects for legacy prefix with legacy amino codec.
+		if bytes.Equal(types.AttributeKeyPrefix, prefix) {
```
iramiller, line 91: Following line is using amino, probably want amino prefix here

x/attribute/keeper/keeper.go
```diff
@@ -78,20 +79,24 @@ func (k Keeper) GetAttributes(ctx sdk.Context, acc sdk.AccAddress, name string)
 }
 
 // IterateRecords iterates over all the stored attribute records and passes them to a callback function.
-func (k Keeper) IterateRecords(ctx sdk.Context, handle Handler) error {
+func (k Keeper) IterateRecords(ctx sdk.Context, prefix []byte, handle Handler) error {
```
iramiller, line 82: Having this prefix param here kind of leaks some implementation details out of the keeper that I am not sure we really should...

x/attribute/keeper/keeper.go
arnabmitra: good catch! :( changed

x/attribute/keeper/keeper.go
arnabmitra: we can remove this after we do the conversion, but technically an iterate handler could be using to iterate over different prefixes(rather keys, like the name module) without writing methods for each, i thought having a generic handler already violates the encapsulation principle here, meaning that the caller is responsible for how/what they iterate/filter through..but for this one we can remove the param after this conversion..or we can write the a custom method for that prefix(dup code, but again can be removed)..

iramiller (approved):

PR #236, opened 2021-04-08
Remove extra Object Store Locator storage.

dwedul-figure:
## Description

* Change OSLocatorExists to take in an AccAddress instead of a string. This matches the signature of the other keeper functions.
* Remove double setting of OSL entries as two separate entries. The second entry wasn't needed.
* Clean up DeleteRecord to be more efficient.
* Update ModifyRecord to just write over the existing record.
* Change name of GetOsLocatorAddressKeyPrefix to GetOSLocatorKey to better match the other key generation funcs.

---

write a little note why.

### DIFF
--- a/x/metadata/keeper/objectstore.go
+++ b/x/metadata/keeper/objectstore.go
@@ -8,11 +8,9 @@ import (
 	"github.com/provenance-io/provenance/x/metadata/types"
 )
 
-// Handler is a name record handler function for use with IterateRecords.
-type ObjectStoreHandler func(record types.ObjectStoreLocator) error
-
-func (k Keeper) GetOsLocatorRecord(ctx sdk.Context, ownerAddress sdk.AccAddress) (osLocator types.ObjectStoreLocator, found bool) {
-	key, err := types.GetOsLocatorAddressKeyPrefix(ownerAddress)
+// GetOsLocatorRecord Gets the object store locator entry from the kvstore for the given owner address.
+func (k Keeper) GetOsLocatorRecord(ctx sdk.Context, ownerAddr sdk.AccAddress) (osLocator types.ObjectStoreLocator, found bool) {
+	key, err := types.GetOSLocatorKey(ownerAddr)
 	if err != nil {
 		return types.ObjectStoreLocator{}, false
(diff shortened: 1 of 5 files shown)


x/metadata/keeper/objectstore.go
```diff
@@ -8,60 +8,41 @@ import (
...
+// An error is returned if no account exists for the address.
+// An error is returned if an OS Locator already exists for the address.
 func (k Keeper) SetOSLocatorRecord(ctx sdk.Context, ownerAddr sdk.AccAddress, uri string) error {
-	var err error
-
 	urlToPersist, err := k.checkValidURI(uri, ctx)
 	if err != nil {
 		return err
 	}
 	if account := k.authKeeper.GetAccount(ctx, ownerAddr); account == nil {
```
iramiller, line 42: This check shouldn't be here ... the owner account would have to sign a request to set a locator in the first place and doing this extra account exists check in the keeper leads to methods like the import locator that genesis and test processes need to set locators without an explicit dependency.  Using the auth module to check if an account exists seems like an antipattern.

x/metadata/keeper/objectstore.go
```diff
@@ -189,14 +141,11 @@ func (k Keeper) ModifyRecord(ctx sdk.Context, ownerAddr sdk.AccAddress, uri stri
 	return nil
 }
 
-// ImportLocatorRecord binds a name to an address. An error is returned if no account exists for the address.
+// ImportLocatorRecord binds a name to an address in the kvstore.
+// Different from SetOSLocatorRecord in there is less validation.
```
iramiller, line 145: As noted in the other comment, the use of the auth module to check if an account exists when the normal msg call path has already required it to be the signer and pay the gas fees is an invalid constraint that leads to a duplicate method like this one for setting locators during genesis or similar flows where the dependency/order of account creation can not be determined reliably ahead of time.

iramiller (approved):
A couple of comments that really apply to the structure of this code outside of what the PR is addressing.  Approved as is because it seems fine right now but perhaps the other pieces should be refactored as well.

PR #247, opened 2021-04-13
Add typed events and metrics to the marker module

channa-figure:
## Description

Implement Typed Events for Marker Module

closes: #87 

---

write a little note why.


x/marker/handler_test.go
```diff
@@ -72,58 +81,768 @@ func TestInvalidProposal(t *testing.T) {
...
+			nil,
+			0,
+		},
+		{
+			"should successfully activate marker",
+			types.NewMsgActivateRequest(hotdogDenom, s.user1Addr),
+			[]string{s.user1},
+			"",
+			types.NewEventMarkerActivate(hotdogDenom, s.user1),
+			4, //finalize marker will emit a send and message event from bank module
```
iramiller, line 357: the event index stuff feels kind of brittle... perhaps a helper method that does a "contains event" check instead... otherwise if the bank module (or another future module involved here) changes events these tests will break...

x/marker/keeper/msg_server.go
```diff
@@ -80,17 +81,26 @@ func (k msgServer) AddMarker(goCtx context.Context, msg *types.MsgAddMarkerReque
...
+	if err := ctx.EventManager().EmitTypedEvent(markerAddEvent); err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		telemetry.IncrCounterWithLabels(
+			[]string{types.ModuleName, "add", "marker"},
+			1,
+			[]metrics.Label{
+				telemetry.NewLabel(types.EventTelemetryLabelAccess, markerAddEvent.Amount),
```
iramiller, line 95: Access label, Amount value?

iramiller (commented):
Overall this looks like a drastic improvement.  The assertions of expected events (results) when testing the handler is particularly notable.  One concern I have though is that the msg_server is emitting these marker events and not the specific keeper methods.

As a specific example... the `Burn()` msg_server method should emit the basic "message" events the SDK uses... while the `k.Keeper.BurnCoin` method should emit the `types.NewEventMarkerBurn()` event itself.  With this type of structure the proper event will be emitted whenever that keeper method is used (either from the msg_server message invoke or even if used directly).  Making this change should cause a proper set of events to be published when a composite action is invoked (perhaps mint/burn together, or mint and transfer, etc).

On a related note if we use a "contains event" approach in the unit tests then we will be more flexible to event dispatch changes in the future either from updates to this module or to upstream SDK modules.

x/marker/handler_test.go
channa-figure: Yes, that is even better.  I will do that.  My original idea was to clear it after each test so it would always be the event at index 0, but that isn't possible...

x/marker/keeper/msg_server.go
channa-figure: Good catch...

iramiller (approved):
This one is a big step forward this module.  Nice work.

Issue #275, opened 2021-04-27
Legacy P8e contract specification uuid issue


iramiller:
## Summary of Bug

Update contract specification uuid creation for P8e and the type name used for session naming due to mismatched values

## Version

1.2.0-HEAD

### Update Contract Spec UUID Generation
The contract specification id should be based on the first 16 bytes of the hash of the overall contract specification proto value for the legacy P8e implementation.

https://github.com/provenance-io/provenance/blob/4b06b815431b38f6b22bbf9c76c8f7900a98601b/x/metadata/types/p8e.go#L24-L32
```
// uuid calculation
"github.com/golang/protobuf/proto"

// pull the bytes of the provided contract spec object
bz, err := proto.Marshal(&spec)
if err != nil {
	return "", err
}
// calculate a sha512 hash of this value
raw := sha512.Sum512(bz)
specUUID, err := uuid.FromBytes(raw[0:16])
if err != nil {
	return newSpec, nil, err
}
id := ContractSpecMetadataAddress(specUUID)
```

###
Ensure the memorialize contract is pointing at the correct hash value (undo recent change) such that the hash used will match the hash calculated above.

### Update the session name value 
https://github.com/provenance-io/provenance/blob/4b06b815431b38f6b22bbf9c76c8f7900a98601b/x/metadata/types/p8e.go#L177
```
p8EData.Session.Name = msg.Contract.Definition.Name
```

____

## For Admin Use

- [ ] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #277, opened 2021-04-27
Fix p8e contract specification uuid.

dwedul-figure:
## Description

For `WriteP8EContractSpec`, use a hash of the proto as the source hash. Also use the first 16 bytes of that hash as the uuid. This more closely reflects what was previously done.

For `P8EMemorializeContract` get the contract spec uuid from the `contract.Spec` area. Also use the `contract.Definition` for the session name (instead of using the `.Spec`).

Fix the `SetRecord` event key (was `scope_id` but should be `record_id`).

closes: #275

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
+### Bug Fixes
+
+* More mapping fixes related to `WriteP8EContractSpec` and `P8EMemorializeContract` #275
+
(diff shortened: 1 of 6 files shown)


piercetrey-figure (approved):
This appears to be working as intended with the new hash now, thanks!

iramiller (approved):
The only concern I have is if the updated migrate test hashes actually match up correctly or not ... we shouldn't have needed to do a conversion on the spec/sha512 as coming from the old chain there were proper ids in place already...

Issue #276, opened 2021-04-27
Support for arbitrary session context state information


iramiller:
## Summary

A flexible method for persisting context state with Sessions in the metadata module is required.

## Problem Definition

The client side operations for p8e need to persist additional context information within a session.  Extending the session proto specification with a new field of type `Any` would allow client specific data to be persisted along with a session to restore previous state during processing.

Adding a proto `Any` field for session context state would allow a client provided type structure to be used to add additional fields without requiring further changes to the metadata endpoints.

Potential downsides would be increased message size when a client adds significant extra data in this field.  This risk is offset by the client paying gas fees for their use.

## Proposal

Extend the Session message with a new filed
https://github.com/provenance-io/provenance/blob/4305f62b24212d388ffc6082e8c289a50d8a9345/proto/provenance/metadata/v1/scope.proto#L102-L122

Insert a field for session state into the existing message with a type of `Any`... update cli, query, and tx methods as required to persist this data as raw bytes when provided.  Default value should remain null (as well as being null for any existing sessions)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #278, opened 2021-04-27
Add context field to sessions.

dwedul-figure:
## Description

Add a `context` field to the `Session` message.

Also add a `write-session` CLI command and standardize the `Use` lines for the metadata tx CLI commands based on the cobra documentation.

closes: #276 

---

write a little note why.


piercetrey-figure:
@dwedul-figure I think we will need a way to set this context on the `MsgP8eMemorializeContractRequest` and have that be set on the session via the conversion

dwedul-figure:
> @dwedul-figure I think we will need a way to set this context on the `MsgP8eMemorializeContractRequest` and have that be set on the session via the conversion

Doh! Yup. Totally forgot that part. Fixed now though. I added it to the `Contract` message: `MsgP8eMemorializeContractRequest.Contract.Context`. It's an `Any` field there too, and it just gets straight up copied over into the `Session.Context` field.

iramiller (dismissed):
One thing to potentially look for is a validation function for that proto any type url ... there are restrictions on the values that are allowed in there per spec... but I wouldn't try to write your own if there isn't one easily usable within the protobuf SDK.

iramiller:
> A URL/resource name that uniquely identifies the type of the serialized
> protocol buffer message. This string must contain at least
> one "/" character. The last segment of the URL's path must represent
> the fully qualified name of the type (as in
> `path/google.protobuf.Duration`). The name should be in a canonical form
> (e.g., leading "." is not accepted).
> 
> In practice, teams usually precompile into the binary all types that they
> expect it to use in the context of Any. However, for URLs which use the
> scheme `http`, `https`, or no scheme, one can optionally set up a type
> server that maps type URLs to message definitions as follows:
> 
> * If no scheme is provided, `https` is assumed.
> * An HTTP GET on the URL must yield a [google.protobuf.Type][]
> 	value in binary format, or produce an error.
> * Applications are allowed to cache lookup results based on the
> 	URL, or have them precompiled into a binary to avoid any
> 	lookup. Therefore, binary compatibility needs to be preserved
> 	on changes to types. (Use versioned type names to manage
> 	breaking changes.)
> 
> Note: this functionality is not currently available in the official
> protobuf release, and it is not used for type URLs beginning with
> type.googleapis.com.
> 
> Schemes other than `http`, `https` (or the empty scheme) might be
> used with implementation specific semantics.

piercetrey-figure:
I think I am running into issues when sending an Any value related to the stuff @iramiller mentioned above. I have tried several different ways (different prefixes, setting the any type url to `github.com/provenance-io/provenance/x/metadata/types/p8e.UUID` thinking maybe I had to use the go_package version) and seem to always get something like `UNKNOWN: unable to resolve type URL /provenance.metadata.v1.p8e.UUID: tx parse error: invalid request`.

Initially I was using a proto from within p8e, but hten switched to this UUID proto that is present in the provenance p8e protos, hoping that it would resolve, but no luck so far.

dwedul-figure:
Note for the future: The Cosmos SDK was trying to do some automatic validation of the `Any` field and failing because it couldn't find the `TypeURL`s being provided. So we decided to just use a byte array for the `Context` instead. Since the `Session` already is already associated with a process, it shouldn't have to tell others what the `Context` is.

channa-figure (approved):
LGTM, I like the edits to the cobra commands.

PR #336, opened 2021-05-26
P8EMemorializeContract: Limit scope field changes.

dwedul-figure:
…preserve its owners and value owners. Also only add to the data access list (instead of fully overwritting it).

## Description

This PR makes the following changes to the processing done during P8EMemorializeContract:
1.  If there's already a Scope with the given scope id, don't update the `Owners` or `ValueOwnerAddress` fields.
2. If there's already a scope with the given scope id, copy it's existing `DataAccess` values and add any new ones that might be missing (but don't remove any existing values).

### Details:

Currently, during a P8EMemorializeContract call, the `Scope.Owners`, `Scope.ValueOwnerAddress`, and `Scope.DataAccess` values are populated using only info in the `MsgP8EMemorializeContractRequest`.

This has caused a problem in the following scenario:
1) `MsgP8EMemorializeContractRequest` is called with a contract just for an `ORIGINATOR`.
2) `MsgP8EMemorializeContractRequest` is called with a contract that has an `ORIGINATOR` and a `SERVICING`, that is to be added to the previously created scope.
3) `MsgP8EMemorializeContractRequest` is called again for the contract with just for an `ORIGINATOR`.

Problem: Now, both the `ORIGINATOR` and `SERVICING` parties are owners on the scope, so both need to sign for step number 3 (which would then actually remove the `SERVICING` owner).

After some discussion, we decided that we don't want `MsgP8EMemorializeContractRequest` to update the `Scope.Owners`, or `Scope.ValueOwnerAddress` fields. We also want the `Scope.DataAccess` values to only be additive. E.g., during step number 2 (above), the SERVICING party would be added to the `DataAccess` list, but during step 3, they'd still be there (they wouldn't be removed).

---

write a little note why.

### DIFF
--- a/x/metadata/keeper/keeper.go
+++ b/x/metadata/keeper/keeper.go
@@ -245,7 +245,7 @@ func (k Keeper) EmitEvent(ctx sdk.Context, event proto.Message) {
 }
 
 // unionUnique gets a union of the provided sets of strings without any duplicates.
-func (k Keeper) unionNoDups(sets ...[]string) []string {
+func (k Keeper) UnionDistinct(sets ...[]string) []string {
 	retval := []string{}
 	for _, s := range sets {
 		for _, v := range s {
--- a/x/metadata/keeper/keeper.go
+++ b/x/metadata/keeper/keeper.go
@@ -244,6 +244,26 @@ func (k Keeper) EmitEvent(ctx sdk.Context, event proto.Message) {
 	}
 }
 
+// unionUnique gets a union of the provided sets of strings without any duplicates.
+func (k Keeper) unionNoDups(sets ...[]string) []string {
+	retval := []string{}
+	for _, s := range sets {
+		for _, v := range s {
+			f := false
+			for _, r := range retval {
+				if r == v {
+					f = true
+					break
+				}
+			}
+			if !f {
+				retval = append(retval, v)
+			}
+		}
+	}
+	return retval
+}
+
 func (k Keeper) checkValidURI(uri string, ctx sdk.Context) (*url.URL, error) {
 	urlToPersist, err := url.Parse(uri)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


x/metadata/keeper/keeper.go
```diff
@@ -244,6 +244,26 @@ func (k Keeper) EmitEvent(ctx sdk.Context, event proto.Message) {
 	}
 }
 
+// unionUnique gets a union of the provided sets of strings without any duplicates.
+func (k Keeper) unionNoDups(sets ...[]string) []string {
```
channa-figure, line 248: I know it might seem trivial, but shouldn't we write a simple test around this?

channa-figure (changes requested):

x/metadata/keeper/msg_server.go
```diff
@@ -491,6 +491,15 @@ func (k msgServer) P8EMemorializeContract(
 		return nil, err
 	}
 
+	existingScope, found := k.GetScope(ctx, p8EData.Scope.ScopeId)
+	if found {
+		// We don't want to update the scope's owners or value owner.
+		p8EData.Scope.Owners = existingScope.Owners
+		p8EData.Scope.ValueOwnerAddress = existingScope.ValueOwnerAddress
+		// We only want to add to the data access list.
+		p8EData.Scope.DataAccess = k.unionNoDups(existingScope.DataAccess, p8EData.Scope.DataAccess)
```
iramiller, line 500: Nit: I would have used `Distinct` here instead of `NoDups`.

iramiller (commented):
Overall looks good... holding approve for a bit in light of automerge so the comment can be seen.

channa-figure (approved):
Nice work.  LGTM

iramiller (approved):
Awesome.

x/metadata/keeper/keeper.go
dwedul-figure: Added some tests.

x/metadata/keeper/msg_server.go
dwedul-figure: Made the change. My brain kept trying to use `Unique` here but that didn't feel right, so I punted.

Issue #314, opened 2021-05-18
Add support for update and specific attribute delete requests


iramiller:
## Summary

Add support for updating a specific attribute on an account and for removing a specific attribute even when there is more than one with the same name

## Problem Definition

Currently the attribute module only supports set and delete operations.  The delete operation does not include the ability to differentiate between multiple attributes on a single address with the same name, it will remove them all.

## Proposal

Leave the current add/delete operations in place and add two new operations.  One that updates a specific entry and another that deletes a specific entry.  In order for these two methods to be supported the existing attribute instance to be changed will have to be provided (possibly through its hash) so the appropriate entry can be found.  If the entry does not exist the calls should error (vs setting a non-existing attribute or performing a no-op when the attribute is already missing).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #349, opened 2021-06-08
Add support for update and specific attribute delete requests

channa-figure:
## Description

Add support for updating a specific attribute on an account and for removing a specific attribute even when there is more than one with the same name

closes: #314 

---

write a little note why.


x/attribute/keeper/keeper.go
```diff
@@ -163,21 +163,103 @@ func (k Keeper) SetAttribute(
...
+	if ownerAcc := k.authKeeper.GetAccount(ctx, owner); ownerAcc == nil {
+		return fmt.Errorf("no account found for owner address \"%s\"", owner.String())
+	}
+
+	if !k.nameKeeper.ResolvesTo(ctx, updateAttribute.Name, owner) {
+		return fmt.Errorf("\"%s\" does not resolve to address \"%s\"", updateAttribute.Name, owner.String())
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	it := sdk.KVStorePrefixIterator(store, types.AccountAttributesNameKeyPrefix(accountAddress, originalAttribute.Name))
```
dwedul-figure, line 209: The `originalAttribute.Name` hasn't been normalized, but the `updateAttribute.Name` has. Is it correct to use `originalAttribute.Name` here?

x/attribute/keeper/keeper.go
```diff
@@ -163,21 +163,103 @@ func (k Keeper) SetAttribute(
...
+
+	if err = originalAttribute.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if err = updateAttribute.ValidateBasic(); err != nil {
+		return err
+	}
+
+	if updateAttribute.Name != originalAttribute.Name {
```
dwedul-figure, line 184: Should the normalized versions of these be compared instead of the given ones?

x/attribute/keeper/msg_server.go
```diff
@@ -66,6 +66,60 @@ func (k msgServer) AddAttribute(goCtx context.Context, msg *types.MsgAddAttribut
...
+				telemetry.NewLabel(types.EventTelemetryLabelType, msg.UpdateAttributeType.String()),
+				telemetry.NewLabel(types.EventTelemetryLabelAccount, msg.Account),
+				telemetry.NewLabel(types.EventTelemetryLabelOwner, msg.Owner),
+			},
+		)
+	}()
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeAttributeAdded,
```
dwedul-figure, line 114: I think this should be `EventTypeAttributeAdded`. Also, it's not yet using typed events.

Since the actual attribute added/updated/deleted events are emitted while the keeper is doing stuff, these events could probably be better changed to indicate that tx has completed successfully. In the Metadata module, I just made a generic `EventTxCompleted` for that and give it different values for the different endpoints/methods.

x/attribute/types/events.go
```diff
@@ -36,10 +42,31 @@ func NewEventAttributeAdd(attribute Attribute, owner string) *EventAttributeAdd
 	}
 }
 
+func NewEventAttributeUpdate(originalAttribute Attribute, updateAttribute Attribute, owner string) *EventAttributeUpdate {
+	return &EventAttributeUpdate{
+		Name:          originalAttribute.Name,
+		OriginalValue: string(originalAttribute.GetValue()),
+		OriginalType:  originalAttribute.AttributeType.String(),
+		UpdateValue:   string(updateAttribute.GetValue()),
```
dwedul-figure, line 50: Probably want to base64 encode these rather than using `string(...)`.

x/attribute/types/msgs.go
```diff
@@ -9,14 +9,18 @@ import (
 )
 
 const (
-	TypeMsgAddAttribute    = "add_attribute"
-	TypeMsgDeleteAttribute = "delete_attribute"
+	TypeMsgAddAttribute             = "add_attribute"
+	TypeMsgUpdateAttribute          = "update_attribute"
+	TypeMsgDeleteAttribute          = "delete_attribute"
+	TypeMsgDeleteAttributeWithValue = "delete_attribute_with_value"
```
dwedul-figure, line 15: Nit: It's called "Delete distinct" in all the other places.
```suggestion
	TypeMsgDeleteAttributeWithValue = "delete_distinct_attribute"
```
Or maybe `"delete_attribute_distinct"`.

dwedul-figure (commented):
Just a few notes and questions.

As a side note: The attribute module is kind of painful. :( It looks like there's a lot of duplicated validation logic, and it'd be nice if the keeper methods were more light-weight (like in the Metadata module). It'd be nice to overhaul it at some point when we get time (ha!).

x/attribute/client/cli/tx.go
```diff
@@ -80,6 +77,107 @@ func NewAddAccountAttributeCmd() *cobra.Command {
 	return cmd
 }
 
+//  NewUpdateAccountAttributeCmd creates a command for adding an account attributes.
+func NewUpdateAccountAttributeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add [name] [address] [original-type] [original-value] [update-type] [update-value]",
```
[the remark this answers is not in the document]
channa-figure: Good catch...copy and paste...

x/attribute/client/cli/tx.go
```diff
@@ -80,6 +77,107 @@ func NewAddAccountAttributeCmd() *cobra.Command {
...
+	} else {
+		encodedValue = []byte(value)
+	}
+	return encodedValue, nil
+}
+
+// NewDeleteDistinctAccountAttributeCmd creates a command for removing account attributes with specific name value.
+func NewDeleteDistinctAccountAttributeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete [name] [address] [type] [value]",
```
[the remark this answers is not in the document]
channa-figure: Same as before...copy and paste overlook.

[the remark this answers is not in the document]
channa-figure: Added aliases to the attribute module command.  Also, I added them to the marker module in another ticket.  I'm not a big fan of them, that is why I overlook them.  Thanks for catching that.

x/attribute/types/msgs.go
channa-figure: Thanks for this catch.  I had initial used `with_value` for everything.  Then did a refactor.

x/attribute/types/events.go
channa-figure: Not sure if I agree here.  I'm just string what ever value is stored in the kv store, they might already be stored as a base64 encoded, like the attribute type bytes.  We also have the `add` doing the same thing, so I just followed that.  Unless, that should also be doing encoding...

channa-figure:
@dwedul-figure Thanks for the detailed review.  I overlooked the fact that our keeper tests don't validate the correct values are stored.  Just that they finish without errors.  We might want to add a way to do that with our tests.

iramiller:
These new msgs should be added as operations to the simulation testing for the attribute module.

x/attribute/types/events.go
iramiller: I don't think we should be using string context here at all.  The attribute value on chain is `[]byte` array therefor these should be encoded as `base64` strings of that value.

x/attribute/keeper/msg_server.go
```diff
@@ -66,6 +66,60 @@ func (k msgServer) AddAttribute(goCtx context.Context, msg *types.MsgAddAttribut
...
+	}
+
+	err = k.Keeper.UpdateAttribute(ctx, originalAttribute, updateAttribute, ownerAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		telemetry.IncrCounterWithLabels(
+			[]string{types.ModuleName, types.EventTelemetryKeyAdd},
```
dwedul-figure, line 98: This should be `types. EventTelemetryKeyUpdate`.

x/attribute/keeper/msg_server.go
```diff
@@ -66,6 +66,60 @@ func (k msgServer) AddAttribute(goCtx context.Context, msg *types.MsgAddAttribut
...
+				telemetry.NewLabel(types.EventTelemetryLabelType, msg.UpdateAttributeType.String()),
+				telemetry.NewLabel(types.EventTelemetryLabelAccount, msg.Account),
+				telemetry.NewLabel(types.EventTelemetryLabelOwner, msg.Owner),
+			},
+		)
+	}()
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTelemetryKeyUpdate,
```
dwedul-figure, line 114: This should be `types. EventTypeAttributeUpdated` (this one might not yet exist).

dwedul-figure (commented):
Just a couple more telemetry tweaks.

x/attribute/types/events.go
channa-figure: I changed them to be base64 encoded.  @dwedul-figure I was thinking how the proto was handled in the CLI too.  I too believe it should be handled like the AttributeType_bytes.

x/attribute/keeper/msg_server.go
channa-figure: Thanks, delete distinct didn't exist either.  So I added that.

x/attribute/simulation/operations.go
```diff
@@ -94,6 +118,50 @@ func SimulateMsgAddAttribute(k keeper.Keeper, ak authkeeper.AccountKeeperI, bk b
...
+		var attributes []types.Attribute
+		if err := k.IterateRecords(ctx, types.AttributeKeyPrefix, func(attribute types.Attribute) error {
+			attributes = append(attributes, attribute)
+			return nil
+		}); err != nil {
+			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgUpdateAttribute, "iterator of existing attributes failed"), nil, err
+		}
+
+		if len(attributes) == 0 {
+			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgUpdateAttribute, "no attributes available to delete"), nil, nil
```
iramiller, line 135: Nit: 

> no attributes available to delete

no attributes available to update

iramiller (approved):
Looks good!

Issue #154, opened 2021-03-11
Update to Cosmos SDK Release 0.43


iramiller:
## Summary

NOTE: update this ticket with the release version these changes land in.

Cosmos SDK 0.42 originally had many changes included that would break the provenance app.  These changes were postponed but will come in a future release (0.43).

 In addition to these fixes version 0.43 contains extensive breaking changes to the banking API and account addresses.  These changes are expected to have an impact on the `marker` module and will require changes/testing.

### Areas of Concern for Marker Module

- Updated coin metadata including name/symbol fields that must be supported by marker module
- Supply interfaces removed and changed to coin types
- Gov module contains extensive new supply constraint checks that panic the chain if they fail.
- Variable length addresses have been merged, this will require denom addresses to be based on the combination of the marker module name and the denom name as a "child account"
- SetSupply and SetBalance were removed (should be unused)

### General Concern

- Wallet path has been change from SetFullFundraiserPath to Coin and Purpose only.
- Addesses are now simple byte arrays less than 255 in length (up from 20 minimum previously)
- Support for secp256r1 was added.

## Proposal

Update to Cosmos SDK release 0.43
Update `makefile` third party proto dependency targets to SDK release version, ~pull in secp256r1 key type~ (moved to #364)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
Migration to `ibc-go` is required.  See migration guide here:
https://github.com/cosmos/ibc-go/blob/v1.0.0-alpha1/docs/migrations/ibc-migration-043.md

PR #366, opened 2021-06-24
Update to Cosmos sdk Release v0.43.0

channa-figure:
## Description

closes: #154 

---

write a little note why.


app/app.go
```diff
@@ -126,6 +128,8 @@ const (
 	DefaultFeeDenom = "nhash" // nano-hash
 	// DefaultReDnmString is the allowed denom regex expression
 	DefaultReDnmString = `[a-zA-Z][a-zA-Z0-9/\-\.]{2,127}`
+	// DefaultReDnmString is the allowed denom regex expression
```
iramiller, line 131: copy/paste error

x/attribute/keeper/genesis.go
```diff
@@ -29,7 +29,7 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) (data *types.GenesisState) {
 		return nil
 	}
 
-	if err := k.IterateRecords(ctx, types.AttributeKeyPrefix, appendToRecords); err != nil {
+	if err := k.IterateRecords(ctx, types.AttributeKeyPrefixLegacy, appendToRecords); err != nil {
```
iramiller, line 32: you want to be internally consistent in using the new key prefix everywhere ... the only place the legacy one will exist is in the migration function of the module upgrade handler that transitions from version 1 (default, current) to version 2 (instance with new prefix/0.43 support)

iramiller (commented):
Overall this looks like a solid update to the length prefixed address format required for v043+.  It feels like the migration/legacy functions could be streamlined a little bit but these are not really that big of a deal especially as this code will drop out when we add v044 migration stuff at some point in the future.  

The only part of the migration code that feels like it should probably be tweaked slightly for clarity relates to when/where we introduce "legacy" in method names/params.  Within a `legacy/v042` folder functions that have "legacy" in the name could be presumed to be the implementations from <v042 (aka v040 for example). We should be consistent in not suffixing these functions with legacy when the implementation correctly matches what was used for the version of the folder they are in.

iramiller (approved):
I think this is ready to merge and build on for the rest of the 1.6.0 release.

Issue #372, opened 2021-07-07
Updated marker metadata


iramiller:
## Summary

Cosmos SDK 0.43 adds new name and symbol fields to the denom metadata.  Add support for setting these new values in the marker denom metadata fields.

## Problem Definition

Name and symbol are useful for providing additional context behind a marker.  These fields are also required for the bank denom metadata structure and must be set.  Updating the governance proposal #369 and the basic marker denom metadata setting endpoints are required.

## Proposal

Add the ability to set/update the denom metadata marker fields `symbol` and `name`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #382, opened 2021-07-20
Add support for name and symbol fields to marker metadata.

dwedul-figure:
## Description

Adds support of the new name and symbol fields to the marker metadata stuff.

closes: #372

---

write a little note why.


channa-figure:
Everything looks good to me, but would like to know if we have to do any migrations with the existing marker metadata that does not have these properties populated.  In the metadata validate basic it validates that these fields are not empty, not sure if that would cause a problem with the existing data.

app/upgrades.go
```diff
@@ -76,6 +76,35 @@ var handlers = map[string]appUpgrade{
 	"desert":  {},
 	"eigengrau": {
 		Handler: func(app *App, ctx sdk.Context, plan upgradetypes.Plan) (module.VersionMap, error) {
+			nhashName := "Hash"
+			nhashSymbol := "HASH"
+			nhash, found := app.BankKeeper.GetDenomMetaData(ctx, "nhash")
+			if found {
```
iramiller, line 82: If you have the set all option... why have the set partial option as well?

dwedul-figure: I wanted to handle if it's not found. I thought about returning an error, but then I figured it was just as easy to just create it if it's not found. I like the get, update, set way better because then we're not undoing anything that might possibly have been changed since it was created. But if it hasn't been created yet, we know we're not undoing anything. And even though I know that there's very very little chance that it has been changed from the original `bluetiful`, I'm just erring on the side of caution.

iramiller (approved):
Nice!

Issue #265, opened 2021-04-26
Integrate authz module into provenance and use it to transfer restricted marker's


arnabmitra:
## Summary

Integrate authz module into provenance and use it to transfer restricted marker's with permission's from both admin and transferor.

## Problem Definition
Original issue is detailed here https://github.com/provenance-io/provenance/pull/261

Basically restricted marker's need permission of the transferor and admin to move.

Authz does provide a standardized way of granting, exec'ing and removing the grant(once exec'ed) and should probably be used.

Disadvantages are more orchestration by client, but that is expected for restricted markers.

## Proposal

 1. Owner of restricted marker(Transferor) authz grant permission to the admin to move coin 
 2. Administrator of restricted marker gives authz grant to transferor allow transfer coin(needed to prevent direct invoke of the rpc method) 
 3. Admin can now transfer coins via authz execMessage(authz allows msg to have only one signer, so signer of transfer obj remains owner of marker, but signer of authz exec becomes the grantee i.e admin here)
 4. Owner can transfer too by normal invoke but method now checks admin has given grant(this will be using the authz keeper (i think there has to be custom to remove the admin grant to owner grant once exec'ed)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
This is largely a duplicate of #262

arnabmitra:
good with closing this

dpederson-figure:
Per proposal 2) above, you'll need to add authz provwasm hooks for smart contracts for when the SC is admin.

iramiller:
Will be interested in seeing how this plays out... was expecting the case more like 4) above where the source account records an auth for SC withdraw and the SC just does the normal transfer which should succeed because it checks for an existing auth grant... perhaps this isn't how it will actually play out though. 

We will probably need some authz hooks in any case... especially if the authz module is used to do something like authorize a secp256r1 address for a primary users phone against their main account which would have wider implications.

PR #383, opened 2021-07-21
Add authz to marker module tranfer

channa-figure:
## Description

Adds authz permission granting to the marker module's transfer message

closes: #265 

---

write a little note why.


x/marker/client/cli/tx.go
```diff
@@ -539,12 +546,135 @@ func GetNewTransferCmd() *cobra.Command {
...
+func GetCmdGrantAuthorization() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "grant-authz [grantee] [authorization_type]",
+		Aliases: []string{"ga"},
+		Args:    cobra.ExactArgs(2),
+		Short:   "Grant authorization to an address",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`grant authorization to an address to execute an authorization type [transfer]:
+
+Examples:
```
dwedul-figure, line 582: `Examples` should be its own field in the `cobra.Command` rather than just extra text in the `Long` field.

x/marker/client/cli/tx.go
```diff
@@ -539,12 +546,135 @@ func GetNewTransferCmd() *cobra.Command {
...
+func GetCmdRevokeAuthorization() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "revoke-authz [grantee] [authorization_type]",
+		Short:   "Revoke authorization to an address",
+		Aliases: []string{"ra"},
+		Args:    cobra.ExactArgs(2),
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`revoke authorization to a grantee address for authorization type [transfer]
+
+Examples:
```
dwedul-figure, line 647: Same here with `Examples`.

dwedul-figure (commented):
Couple small tweaks to the CLI.

x/marker/client/cli/tx.go
channa-figure: I'm a bit on the fence about this.  I'm trying to keep it consistent with how we are doing it in the current file, where all the other commands have it in the `Long`.  The `metadata` module is the only one that uses the `Example` currently.  Also, the cosmos-sdk uses the `Long` as well.  Here is where I got my command from: https://github.com/cosmos/cosmos-sdk/blob/f961e01bcf19a0f4254864c44306c5126d912e45/x/authz/client/cli/tx.go#L61.  If I modify it, I would want to change all the other commands too, but I would also like to keep this PR to only contain changes for authz support.

channa-figure: I made a ticket to clean up the whole project to use the same pattern. #399 This will be a good ticket for new members of the team.

dwedul-figure: Ah. Yeah. I know I've seen other places that include the examples in the `Long` like that. I'm not even really sure when the Examples show up and when they don't. So it probably doesn't even really matter.

I have a dream that one day, I'll go through and standardize all the CLI usage stuff, but eh, that's a lot of work for some cosmetics.

Good to go!

dwedul-figure (approved):
LGTM

iramiller (approved):
This authz flow looks solid.  Having `from` accounts submit an approval closes a very important signature/approval flow for smart contract use of restricted markers.

Issue #403, opened 2021-07-29
Update server and client configuration defaults based on testnet/mainnet


iramiller:
## Summary

Set configuration defaults that match current mainnet and testnet provenance networks

## Problem Definition

When new nodes/validators join the Provenance networks their starting configuration is often based on what the `provenanced init` command sets up instead of the customized configuration used in the official testnet/mainnet repositories.  This feature would update the defaults in the binary such that the initialized config would match these environments

## Proposal

Add additional configuration default overrides to the `provenanced init` flow so that configuration matches the recommended settings in the testnet/mainnet repositories.  Check the `--testnet` flag status to determine which environment to use as a template/reference for config values.

NOTE: For the `testnet` configuration do not include the seed nodes or related network settings that would conflict with setting up a development environment.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #405, opened 2021-07-30
Update init cmd defaults and create config cmd

dwedul-figure:
## Description

* Update the `provenanced init` command to generate a genesis file based on the mainnet config or testnet config (depending on the `-t` or `--testnet` flag).
* Build out the `provenanced config` command for interacting with configuration.
  * `provenanced config get [<key1> [<key2> ...]]` for looking up and outputting values.
  * `provenanced config set <key1> <value1> [<key2> <value2> ...]` for setting config values in their files.
  * `provenanced config changed [<key1> [<key2>...]` for getting values that are different from their defaults.
  * `provenanced config pack` for creating a single json config containing only non-default config values.
  * `provenanced config unpack` for creating the multiple toml config files.
* Overhauled the command initialization process to handle the config files as needed.
* Cleared up a bit of confusion over the `minimum-gas-prices` default.

closes: #403
closes: #425 

---

write a little note why.


cmd/provenanced/cmd/init.go
```diff
@@ -139,53 +123,142 @@ func Init(
...
+func initGenFile(
+	cmd *cobra.Command,
+	cdc codec.JSONCodec,
+	mbm module.BasicManager,
+	isTestnet bool,
+	chainID, genFile string,
+) error {
+	minDeposit := int64(1000000000000)  // 1,000,000,000,000
+	downtimeJailDurationStr := "86400s" // 1 day
+	maxValidators := uint32(20)
```
iramiller, line 159: This should be 100 now.

cmd/provenanced/cmd/init.go
```diff
@@ -139,53 +123,142 @@ func Init(
...
+	maxGas := int64(60000000) // 60,000,000
+	unrestrictedDenomRegex := `[a-zA-Z][a-zA-Z0-9\-\.]{7,64}`
+	if isTestnet {
+		mustFprintf(cmd.OutOrStdout(), "Using testnet defaults\n")
+		minDeposit = 10000000            // 10,000,000
+		downtimeJailDurationStr = "600s" // 10 minutes
+		maxValidators = 100
+		maxGas = -1
+		unrestrictedDenomRegex = `[a-zA-Z][a-zA-Z0-9\-\.]{2,64}`
```
iramiller, line 168: This was changed on testnet as well as mainnet to implement the restricted marker denom system.  Both require 8 char minimum now.  A question is also if the mixed case denom should be allowed vs lower case only...

cmd/provenanced/cmd/config.go
```diff
@@ -0,0 +1,654 @@
...
+	"github.com/cosmos/cosmos-sdk/server"
+	"github.com/cosmos/cosmos-sdk/version"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
+	// entryNotFound is a magic index value that indicates a thing wasn't found (and so no index is applicable).
+	entryNotFound = -1
```
fkneeland-figure, line 20: Why an integer rather than a boolean?

fkneeland-figure: Oh... b/c you are using the index returned from another package.... nvm

cmd/provenanced/config/client.go
```diff
@@ -50,48 +49,48 @@ func (c *ClientConfig) SetBroadcastMode(broadcastMode string) {
...
+		WithChainID(config.ChainID).
 		WithKeyringDir(ctx.HomeDir)
 
-	keyring, err := client.NewKeyringFromBackend(ctx, conf.KeyringBackend)
-	if err != nil {
-		return ctx, fmt.Errorf("couldn't get key ring: %v", err)
+	keyring, krerr := client.NewKeyringFromBackend(ctx, config.KeyringBackend)
+	if krerr != nil {
+		return ctx, fmt.Errorf("couldn't get key ring: %v", krerr)
 	}
```
channa-figure, line 82: We talked about making the error message a little clearer here.  Currently, the error message is `Error: could not apply client config to client context: couldn't get key ring: Specified keyring backend not available` and maybe we could add a little more detail  like you had suggested: `The keyring-backend value needs to be manually updated in <file>.` ?  Also, the current error doesn't state the name of the keyring entered `Specified keyring backend`, could we also add the `config.KeyringBackend` value?

channa-figure (changes requested):
Everything looks good to me,  but I did add a comment for an error message change.  Once we get that in or at least discuss if we still want it.  I will approve this PR.  This will be really helpful feature for things I end up having to do manually all the time.

cmd/provenanced/config/client.go
dwedul-figure: I can add some direction about updating it. Providing the location of where to make the change is difficult to iron out, though, because there's no way to tell if it was from a config file or an env var. But I guess I can provide the file and also mention env vars.

Working on the update now.

cmd/provenanced/config/filenames.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"path/filepath"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/spf13/cobra"
+)
+
+const (
+	// These are not configuration values.
+	// They are strings that are also defined deep inside some other stuff that is out of our control.
```
fkneeland-figure, line 12: Might be nice to put a link here for that?

cmd/provenanced/config/client.go
dwedul-figure: Updated it to now look like this:
`Error: could not apply client config build/run/provenanced/config/client.toml to client context - it may need to be updated manually: couldn't get key ring kwallet: Specified keyring backend not available`

channa-figure: That looks good to me!

channa-figure (approved):
LGTM.  Nice work!

Issue #473, opened 2021-09-10
Consider the value owner when deleting a scope


dwedul-figure:
## Summary

When deleting a scope, the value owner should be taken into consideration.

## Problem Definition

Currently, when updating a scope's `ValueOwner` field, 
* If the existing value is a marker, one of the signers must have `Access_Withdraw` authority on the marker.
* If the existing value isn't a marker (and isn't empty) it must also be one of the signers.

However, when a scope is being deleted, only the owners must sign. This could be used to circumvent the update requirements on the `ValueOwner` field.

## Proposal

When deleting a scope, if the `ValueOwner` isn't empty:
* If it's a marker, one of the signers must have `Access_Withdraw` authority on the marker.
* Otherwise, it must also be one of the signers.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
Thinking more about this it would be withdraw access and not delete as delete is for the marker itself and withdraw/deposit are used for aggregate asset modification.

PR #477, opened 2021-09-13
Update Scope Removal Validation to match Update Signing reqs.

dwedul-figure:
## Description

Update the signing requirements for deleting a scope to match those for updating a scope.

closes: #473

---

write a little note why.


x/metadata/keeper/scope.go
```diff
@@ -281,20 +247,55 @@ func (k Keeper) ValidateScopeUpdate(ctx sdk.Context, existing, proposed types.Sc
...
 	return nil
 }
 
+func (k Keeper) validateScopeUpdateValueOwner(ctx sdk.Context, existing, proposed string, signers []string) error {
+	// If they're the same, we don't need to do anything.
+	if existing == proposed {
+		return nil
+	}
+	if len(existing) > 0 {
+		if k.AccountIsMarker(ctx, existing) {
```
arnabmitra, line 268: since `k.AccountIsMarker(ctx, existing)` does read the state store, and is used late(potentially), maybe store it?

arnabmitra (dismissed):

x/metadata/keeper/scope.go
arnabmitra: oh sorry did not see this was a draft :(

dwedul-figure: All good! It's a good call and kind of nicer to change before putting it into ready for review!

dwedul-figure: I just double checked. It's actually checking two different values. `AccountIsMarker` is called once for `existing` and once for `proposed`. There's a short-circuit at the top when the two are equal, so we know it's not making the same check twice. The calls to `HasSignerWithMarkerValueAuthority` are similar with the added difference of needed different permissions.

arnabmitra: Makes sense, thanks for verifying!

dwedul-figure: You were right about being able to save a lookup. Both `k.AccountIsMarker(ctx, existing)` and `k. HasSignerWithMarkerValueAuthority(ctx, existing, ...)` were doing the same marker account lookup. So I combined them into one function so they can both use the same thing.

CHANGELOG.md
```diff
@@ -65,6 +65,11 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * Add a single node docker based development environment [#311](https://github.com/provenance-io/provenance/issues/311)
   * Add make targets `devnet-start` and `devnet-stop`
   * Add `networks/dev/mnemonics` for adding accounts to development environment
+* When deleting a scope, require the same permissions as when updating it [#473](https://github.com/provenance-io/provenance/issues/473)
```
fkneeland-figure, line 68: Do you want to move these up to the unreleased section instead of in v1.7.0?

dwedul-figure: Good catch! Fixed.

fkneeland-figure (dismissed):

x/metadata/keeper/signatures.go
```diff
@@ -13,53 +13,35 @@ import (
...
+			saddr, serr := sdk.AccAddressFromBech32(signer)
+			// If the signer address is okay, check it for the role. If it checks out, they've got auth and we're done.
+			if serr == nil && marker.AddressHasAccess(saddr, role) {
+				hasAuth = true
+				break
 			}
 		}
 	}
-	return false
+	return
```
iramiller, line 44: NIT: Naked returns like this are my least favorite go feature ... especially in this case where it is not setting "out" params but is relying on the default value of the bool being false.  `return true` and `return false` are more clear and explicit.

iramiller (commented):
Did you do a comparison in the stack pressure of the sorted party version as well as the computational efficiency? Memory allocations can be quite expensive and even if you need to traverse the list multiple times the list sizes are generally small here... further in many cases the short circuit break on first non-equal find can be expected to improve performance as well.

x/metadata/keeper/signatures.go
dwedul-figure: Yeah. I'm not a fan of naked returns like this either. Since the function returns `(bool, bool)`, I wanted to name them in the function signature. And I thought `golang` _wants_ you to use the naked returns when they're named. But I tried providing the return values and the linter doesn't complain, so I'll commit that change and push it up.

dwedul-figure:
> Did you do a comparison in the stack pressure of the sorted party version as well as the computational efficiency? Memory allocations can be quite expensive and even if you need to traverse the list multiple times the list sizes are generally small here... further in many cases the short circuit break on first non-equal find can be expected to improve performance as well.

I did not do any such comparisons. I just noticed that the old `EqualParties` function would return true if `x` is `a, a, b` and `y` is `a, b, b`. But now I see that, during basic validation, we check for and disallow duplicate owner entries. So I'll swap this back to the old version.

dwedul-figure:
> > Did you do a comparison in the stack pressure of the sorted party version as well as the computational efficiency? Memory allocations can be quite expensive and even if you need to traverse the list multiple times the list sizes are generally small here... further in many cases the short circuit break on first non-equal find can be expected to improve performance as well.
> 
> I did not do any such comparisons. I just noticed that the old `EqualParties` function would return true if `x` is `a, a, b` and `y` is `a, b, b`. But now I see that, during basic validation, we check for and disallow duplicate owner entries. So I'll swap this back to the old version.

Update: I also found that the old `EqualParties` function would return true if `x` is `a,a` and `y` is `a,b`. So you've actually got to do the double loop twice.

iramiller:
> I did not do any such comparisons. I just noticed that the old EqualParties function would return true if x is a, a, b and y is a, b, b. But now I see that, during basic validation, we check for and disallow duplicate owner entries. So I'll swap this back to the old version.

I have had this type of complexity on my mind of late after seeing some of the stuff the sdk found that was impacting performance ... but its all shooting in the dark if you are not measuring anything.  I don't know that the old way was better (obviously not if its missing test cases!) but when I saw the sorts/reallocations that made me think of it.

The extensive guarding tests are the real deal though, something that works correctly is better than something that works half way but faster...  definitely appreciate the extensive and complete test cases in these PRs.

iramiller (approved):
LGTM

Issue #484, opened 2021-09-16
Issue with marker migration in eigengrau upgrade 


iramiller:
## Summary of Bug

Marker clone issue in eigengrau upgrade handler

## Version

1.7.0

## Steps to Reproduce

During migration of `pio-testnet-1` using `eigengrau` handler the following error occured:

```
 12:05AM INF Migrating Marker Module from Version 1 to 2
 panic: merger not found for type:big.Word
 goroutine 1 [running]:
 github.com/gogo/protobuf/proto.(*mergeInfo).computeMergeInfo(0xc0000a4f40)
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:662 +0xed9
 github.com/gogo/protobuf/proto.(*mergeInfo).merge(0xc0000a4f40, {0xc002000c60}, {0xc0000a7b78})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:113 +0x58
 github.com/gogo/protobuf/proto.(*mergeInfo).computeMergeInfo.func30({0xc0000a4c40}, {0x1afe2e0})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:587 +0x8b
 github.com/gogo/protobuf/proto.(*mergeInfo).merge(0xc0000a4c40, {0x34b8040}, {0x99})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:139 +0x305
 github.com/gogo/protobuf/proto.(*mergeInfo).computeMergeInfo.func28({0xc0000a48c0}, {0x40})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:555 +0x26
 github.com/gogo/protobuf/proto.(*mergeInfo).merge(0xc0000a48c0, {0x1de1e60}, {0xc001714bd0})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:139 +0x305
 github.com/gogo/protobuf/proto.(*InternalMessageInfo).Merge(0x347a000, {0x23c6c88, 0xc0000a7b60}, {0x23c6c88, 0xc0000a7aa0})
         github.com/gogo/protobuf@v1.3.3/proto/table_merge.go:50 +0xb6
 github.com/provenance-io/provenance/x/marker/types.(*MarkerAccount).XXX_Merge(0x1df0060, {0x23c6c88, 0xc0000a7aa0})
         github.com/provenance-io/provenance/x/marker/types/marker.pb.go:204 +0x3a
[...]

PR #486, opened 2021-09-16
484 marker permission migration error on eigengrau upgrade

channa-figure:
## Description

Fixed marker permissions migration to not use `Clone` method.

Add panic for `eigengrau` upgrade.

closes: #484 

---

write a little note why.

### DIFF
--- a/x/attribute/keeper/migrations.go
+++ b/x/attribute/keeper/migrations.go
@@ -18,5 +18,7 @@ func NewMigrator(keeper Keeper) Migrator {
 // Migrate1to2 migrates from version 1 to 2 to convert attribute keys from 20 to 32 length
 func (m *Migrator) Migrate1to2(ctx sdk.Context) error {
 	ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2")
-	return v042.MigrateAddressLength(ctx, m.keeper.storeKey)
+	err := v042.MigrateAddressLength(ctx, m.keeper.storeKey)
+	ctx.Logger().Info("Finished Migrating Attribute Module from Version 1 to 2")
+	return err
 }
--- a/x/attribute/keeper/migrations.go
+++ b/x/attribute/keeper/migrations.go
@@ -17,7 +17,7 @@ func NewMigrator(keeper Keeper) Migrator {
 
 // Migrate1to2 migrates from version 1 to 2 to convert attribute keys from 20 to 32 length
 func (m *Migrator) Migrate1to2(ctx sdk.Context) error {
-	ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2")
+	ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2 (1/1)")
 	err := v042.MigrateAddressLength(ctx, m.keeper.storeKey)
 	ctx.Logger().Info("Finished Migrating Attribute Module from Version 1 to 2")
 	return err
--- a/x/marker/keeper/migrations.go
+++ b/x/marker/keeper/migrations.go
@@ -18,8 +18,11 @@ func NewMigrator(keeper Keeper) Migrator {
 // Migrate1to2 migrates from version 1 to 2.
 func (m *Migrator) Migrate1to2(ctx sdk.Context) error {
 	ctx.Logger().Info("Migrating Marker Module from Version 1 to 2")
-	if err := v042.MigrateMarkerAddressKeys(ctx, m.keeper.storeKey); err != nil {
+	err := v042.MigrateMarkerAddressKeys(ctx, m.keeper.storeKey)
+	if err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


x/marker/keeper/migrations.go
```diff
@@ -18,8 +18,11 @@ func NewMigrator(keeper Keeper) Migrator {
 // Migrate1to2 migrates from version 1 to 2.
 func (m *Migrator) Migrate1to2(ctx sdk.Context) error {
 	ctx.Logger().Info("Migrating Marker Module from Version 1 to 2")
-	if err := v042.MigrateMarkerAddressKeys(ctx, m.keeper.storeKey); err != nil {
+	err := v042.MigrateMarkerAddressKeys(ctx, m.keeper.storeKey)
+	if err != nil {
```
fkneeland-figure, line 22: #nit not important but isn't the original way more idiomatic to go?

x/attribute/keeper/migrations.go
```diff
@@ -17,6 +17,8 @@ func NewMigrator(keeper Keeper) Migrator {
 
 // Migrate1to2 migrates from version 1 to 2 to convert attribute keys from 20 to 32 length
 func (m *Migrator) Migrate1to2(ctx sdk.Context) error {
-	ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2")
-	return v042.MigrateAddressLength(ctx, m.keeper.storeKey)
+	ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2 (1/1)")
+	err := v042.MigrateAddressLength(ctx, m.keeper.storeKey)
+	ctx.Logger().Info("Finished Migrating Attribute Module from Version 1 to 2")
```
fkneeland-figure, line 22: #nit Not important but instead of creating a new variable and then returning it after the logging statement, would it be a little clearer to put the logging in a defer statement?
```golang
ctx.Logger().Info("Migrating Attribute Module from Version 1 to 2 (1/1)")
defer ctx.Logger().Info("Finished Migrating Attribute Module from Version 1 to 2")
return v042.MigrateAddressLength(ctx, m.keeper.storeKey)
```

x/metadata/legacy/v042/store.go
```diff
@@ -7,23 +7,29 @@ import (
...
 		return err
 	}
+	ctx.Logger().Info("Migrating Metadata Scope Specs (4/5)")
 	err = MigrateAddressScopeSpecCacheKey(ctx, storeKey)
 	if err != nil {
 		return err
 	}
-	return MigrateAddressContractSpecCacheKey(ctx, storeKey)
+	ctx.Logger().Info("Migrating Metadata Contract Specs (5/5)")
+	err = MigrateAddressContractSpecCacheKey(ctx, storeKey)
```
fkneeland-figure, line 31: #nit this could just be:
```golang
return MigrateAddressContractSpecCacheKey(ctx, storeKey)
```

fkneeland-figure (approved):

x/metadata/legacy/v042/store.go
channa-figure: Oh, yes.  This was overlooked.  I think I was going to put a logging statement after that but removed it.

x/marker/keeper/migrations.go
channa-figure: You are correct, but I wanted to use the err variable in the next statement.

x/attribute/keeper/migrations.go
channa-figure: Good point, I didn't think of that.  I will use something like this next time.  Since, if I were to change it now.  I would feel like I need to run the tests again which takes hours.

fkneeland-figure (dismissed):

arnabmitra (approved):

iramiller (dismissed):

x/marker/legacy/v042/store.go
```diff
@@ -33,7 +35,11 @@ func MigrateMarkerPermissions(ctx sdk.Context, k MarkerKeeperI) error {
 			invalid := marker.AddressListForPermission(types.Access_Transfer)
 			// invalid permission grants exist, remediation required
 			if len(invalid) > 0 {
-				m := marker.(*types.MarkerAccount)
+				m, ok := marker.(*types.MarkerAccount)
+				if !ok {
+					err = fmt.Errorf("unable to cast MarkerAccountI to MarkerAccount: %v", marker)
+					return true
```
fkneeland-figure, line 41: Why return true instead of return err?

channa-figure: The is the return value type of the function that is being passed into the iterator.  The true means stop and we are setting the err value which is outside the iterator function.

fkneeland-figure (approved):

iramiller (approved):

PR #515, opened 2021-09-30
Base setup for postgres indexer

channa-figure:
## Description

This is just a work in progress and will wait for the configuration issue to be fixed before finishing a seamless dev instance with postgres indexing enabled.

closes: #XXXX

---

write a little note why.


docker/postgres-indexer/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This docker container is for using the postgres indexer.  
+
+To configure postgres indexing change indexer configuration in the `config.toml`
+
+```toml
+[tx_index]
+
+indexer = "psql"
+psql-conn = "postgresql://postgres:password1@localhost:5432/tendermint?sslmode=disable"
```
dwedul-figure, line 11: The `psql-conn` config entry is not in the current Tendermint config file template. It's in the config object though, so it shows up when you do `provenanced config get tx_index`. Manually adding and setting it in the file works, but it's use is risky for a few reasons:
1. If the config is unpacked (i.e. in the `toml` files) and the `provenanced config set` command is used for anything, this entry will be wiped out from the config file. Lemma: `provenanced config set psql-conn 'whatever'` won't actually set anything on an unpacked config.
2. If the config is packed (i.e. in the `json` file) and contains this entry, and is then unpacked (i.e. `provenanced config unpack`) the entry will be wiped out from the config.
3. Since it's not currently in the config file template, I'm a bit nervous that it's not actually ready.

A safer option would be to instruct them use environment variables:
```
> export PIO_TX_INDEX_INDEXER='PSQL'
> export PIO_TX_INDEX_PSQL_CONN='postgresql://postgres:password1@localhost:5432/tendermint?sslmode=disable'
```
The [direnv](https://direnv.net/) system can be handy for this too (and `.envrc` is already ignored by git).

Another alternative is to provide them as flags when starting the node: `--tx_index.indexer psql --tx_index.psql--conn 'postgresql://postgres:password1@localhost:5432/tendermint?sslmode=disable'`.

Those two options only really handle the config aspect, though. It still remains that the `psql` indexer might not be a feature that's ready yet.

Also of note: In the next version of Tendermint, this section is renamed to `tx-index` (hyphen instead of underscore). The environment variable names will end up the same, but the flags will change and the config file group entry will change.

SpicyLemon: With the `custom.toml` addition (#990), we have a safe place to put this config stuff. The env var and flags stuff listed in the previous comment are still valid, but now there's a safe config file to put this stuff too. The file contents would be the same as listed in this readme, it's just that it should go in `custom.toml` instead of any of the other config files.

Issue #354, opened 2021-06-15
Msg Based Fee Support


iramiller:
## Summary

As part of the provenance blockchain economics certain messages will require a base fee to be paid in addition to the normal gas consumption.  This feature is for the creation of a new antehandler that consumes a fee based on the msgs contained in a tx according to a fee schedule that is persisted on chain.

## Problem Definition

In order to create a balanced token economy on Provenance without the use of inflation to offset low volume/low gas prices additional fees must be collected.  The gas price must also remain low to facilitate basic token transfers.  Due to these two constraints certain operations must be priced significantly higher.

## Proposal

- [ ] Create a new module that allows a msg fee schedule to be created
   - Set a sdk.FeeToken amount against a msg type.
   - Fees should be payable to either the fee payer module (default) or an alternate account address
   - Governance proposals should be used to create entries in the fee table.
   - A default "unknown" msg type fee should be supported
- [ ] Create a new `MemPoolFeeDecorator` ante handler that deducts fee amounts prior to determining the gas price/fee ratio based on a fee schedule for different msg types.

### Future Work

The basic fees per msg type is a useful starting point for fees however certain msgs are abstract and represent many different business operations (MemorializeContract, Add Scope on Metadata module).  For these msg types the module itself must define a way to handle the determination of if any extra fees are required.  One possible way to do this would be a new StdMsg interface type that would define an extra method for returning a msg type string that is extended with additional data that can be appended to the type string for a messages.

```
// Annotation of a memorialize contract request with a scope specification id
provenance.metadata.v1.MsgP8eMemorializeContractRequest#scopespec1qnwg86nsatx5pl56muw0v9ytlz3qu3jx6m
```
Above example shows a scope specification annotation on the end of a type url.  This enhancement could be used to resolve to a specific fee schedule row allowing a fee to be targeted to a specific user.  The enhanced data might be used to add an additional fee to the base fee or replace the fee entirely.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #517, opened 2021-09-30
Feature/msg based fees

channa-figure:
## Description

Adding of the msg based fee module

closes: #354 
# Concepts

# Additional Msg Fees

As part of the provenance blockchain economics certain messages *may* require an additional fee to be paid in
addition to the normal gas consumption. 

Additional Fees are assessed and finally consumed based on msgType of the msgs contained in the transaction,
and the fee schedule that is persisted on chain, created via governance.

Additional fee can be in any *denom*.

# Base Fee
Base fee is currently paid fees, paid in base denom and determined by gas value passed into the Tx.
The value collected remains the same.

# Total Fees
Total fees = Additional Fees (if any) + Base Fee
Total fees continue to be passed as `sdk.Coins` the Tx accepts for fee entry currently.
e.g usd.example is the denom(assuming there is a marker/coin of type usd.example) in which additional fee is being charged in
```bash
--fees 382199010nhash,99usd.example 
```

# Additional fee assessed in base denom i.e nhash
To preserve backwards compatability of all invokes, clients continue accepting fees in sdk.Coins
`type Coins []Coin`, and because the code needs to distinguish between base fee and additional fee,
the msgfees module introduces an additional param, described in 06_params.md, called `DefaultFloorGasPrice`
to differentiate between base fee and additional fee when additional fee is in same denom as default base denom i.e nhash.

This fee is charged initially by the antehandler, if any excess fee is left, once additional fee are paid, that's collected
at the end of the Tx also(same as current behavior)

For e.g
Additional fee = 10000nhash
Gas = 10000
Fee passed in = 19070000nhash

In this client passes in an extra 10000nhash (1905 * 10000 +10000 = 19060000nhash).
Current behavior is maintained and tx passes and charges 19050000 initially and 1000 nhash plus 1000nhash extra fee passed in
the deliverTx stage.
Thus, this will protect against future changes like priority mempool as well as keep current behaviour same as current production. 

# Authz and wamsd messages
Authz and wasmd messages are dispatched via the submessages route, so they get charged and assessed the same additional
fee if set on a submessage, caveat being they forfeit all their fees if they fail (since we have no way upfront of knowing what 
the submessages maybe)

For e.g
Let's say a `MsgSend` has a fee of 100usd.local and a smartcontract does 3 MsgSend operations as per the logic of the 
smart contract, the code will expect additional fees of 300 usd.local to be present for the Tx to be successful.

# Simulation and knowing how much additional fee to be paid.

Current simulation method looks like this  
```kotlin
val cosmosService = cosmos.tx.v1beta1.ServiceGrpc.newBlockingStub(channel)
cosmosService.simulate(SimulateRequest.newBuilder().setTx(txFinal).build()).gasInfo.gasUsed
            
```

In the future we recommend using the method 
```kotlin
val msgFeeClient = io.provenance.msgfees.v1.QueryGrpc.newBlockingStub(channel)
msgFeeClient.simulate(CalculateTxFeesRequest.newBuilder().setTx(txFinal).build())

```

or from the cmd line as

```bash
provenanced tx simulate <required params>
```

---

write a little note why.


iramiller (commented):
First half of review...

x/msgfees/abci.go
```diff
@@ -0,0 +1,10 @@
+package msgfees
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/provenance-io/provenance/x/msgfees/keeper"
+)
+
+// EndBlocker is called at the end of every block
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
```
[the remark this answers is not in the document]
channa-figure: Removed.

app/app.go
```diff
@@ -252,6 +252,8 @@ type App struct {
 	EvidenceKeeper   evidencekeeper.Keeper
 	FeeGrantKeeper   feegrantkeeper.Keeper
 
+	MsgBasedFeeKeeper msgfeekeeper.Keeper
```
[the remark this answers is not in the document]
channa-figure: Yeah, I didn't realize we were kind of all over the place on this.  We have it all over the PR.  I will go through and name them all to match the module name.  No need to point them out else where in the PR.

app/app.go
```diff
@@ -102,31 +97,35 @@ import (
...
+	markertypes "github.com/provenance-io/provenance/x/marker/types"
+	markerwasm "github.com/provenance-io/provenance/x/marker/wasm"
 	"github.com/provenance-io/provenance/x/metadata"
 	metadatakeeper "github.com/provenance-io/provenance/x/metadata/keeper"
 	metadatatypes "github.com/provenance-io/provenance/x/metadata/types"
 	metadatawasm "github.com/provenance-io/provenance/x/metadata/wasm"
+	"github.com/provenance-io/provenance/x/msgfees"
+	msgfeekeeper "github.com/provenance-io/provenance/x/msgfees/keeper"
+	msgfeesmodule "github.com/provenance-io/provenance/x/msgfees/module"
+	msgbasedfeestypes "github.com/provenance-io/provenance/x/msgfees/types"
```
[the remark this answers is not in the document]
channa-figure: Again, we have this all over the project.  I will clean it up.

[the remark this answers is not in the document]
arnabmitra: sorry this is my bad :(

app/upgrades.go
```diff
@@ -213,7 +213,7 @@ func CustomUpgradeStoreLoader(app *App, info storetypes.UpgradeInfo) baseapp.Sto
 				"upgrade.deleted", upgrade.Deleted,
 				"upgrade.renamed", upgrade.Renamed,
 			)
-			return upgradetypes.UpgradeStoreLoader(info.Height, &storeUpgrades)
+			return UpgradeStoreLoader(info.Height, &storeUpgrades)
```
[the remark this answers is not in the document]
arnabmitra: oh i think this is a remnant of when we were copying cosmos app instead of forking..not needed..i'll remove it

app/app.go
```diff
@@ -718,7 +746,7 @@ func (app *App) LoadHeight(height int64) error {
 }
 
 // ModuleAccountAddrs returns all the app's module account addresses.
-func (app *App) ModuleAccountAddrs() map[string]bool {
+func ModuleAccountAddrs() map[string]bool {
 	modAccAddrs := make(map[string]bool)
 	for acc := range maccPerms {
```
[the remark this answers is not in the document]
arnabmitra: yup not used..i'll remove

internal/antewrapper/tracing_meter.go
```diff
@@ -35,7 +35,7 @@ var _ sdkgas.GasMeter = &tracingGasMeter{}
 
 // GasConsumed reports the amount of gas consumed at Log.Info level
 func (g *tracingGasMeter) GasConsumed() sdkgas.Gas {
```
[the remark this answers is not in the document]
arnabmitra: sounds good ..changing this now.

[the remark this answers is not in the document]
arnabmitra: done

x/attribute/keeper/genesis.go
```diff
@@ -1,13 +1,16 @@
...
+	"fmt"
+
 	"github.com/provenance-io/provenance/x/attribute/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // InitGenesis creates the initial genesis state for the attribute module.
 func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
+	ctx.Logger().Info(fmt.Sprintf("Params %v", data.Params))
```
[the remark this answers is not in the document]
arnabmitra: removing

internal/handlers/msg_service_router.go
```diff
@@ -0,0 +1,189 @@
...
+					err = antewrapper.EnsureSufficientFees(runtimeGasForMsg(ctx), feeTx.GetFee(), feeGasMeter.FeeConsumed().Add(fee.AdditionalFee),
+						msr.msgBasedFeeKeeper.GetFloorGasPrice(ctx), msr.msgBasedFeeKeeper.GetDefaultFeeDenom())
+					if err != nil {
+						return nil, err
+					}
+				}
+
+				feeGasMeter.ConsumeFee(fee.AdditionalFee, msgTypeURL)
+			}
+			ctx = ctx.WithEventManager(sdk.NewEventManager())
```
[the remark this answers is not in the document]
channa-figure: My mistake not sure if that was a copy and paste error.  Good catch!

x/marker/simulation/genesis_test.go
```diff
@@ -1,73 +0,0 @@
-package simulation_test
```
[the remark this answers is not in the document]
arnabmitra: oh man, no idea but should never have been..restored.

internal/handlers/msg_service_router.go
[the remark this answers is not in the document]
arnabmitra: hold on this is coming from the cosmod sdk code 
```	msr.routes[requestTypeName] = func(ctx sdk.Context, req sdk.Msg) (*sdk.Result, error) {
			ctx = ctx.WithEventManager(sdk.NewEventManager())
			interceptor := func(goCtx context.Context, _ interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
				goCtx = context.WithValue(goCtx, sdk.SdkContextKey, ctx)
				return handler(goCtx, req)
			}
			// Call the method handler from the service description with the handler object.
			// We don't do any decoding here because the decoding was already done.
			res, err := methodHandler(handler, sdk.WrapSDKContext(ctx), noopDecoder, interceptor)
			if err != nil {
				return nil, err
			}

			resMsg, ok := res.(proto.Message)
			if !ok {
				return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidType, "Expecting proto.Message, got %T", resMsg)
			}

			return sdk.WrapServiceResult(ctx, resMsg, err)
		}```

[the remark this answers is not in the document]
Comment (nullpointer0x00) [reply internal/handlers/msg_service_router.go:155]: Yes, after running the tests without it.  There were lots of duplicate event messages.  For example, after removing it there was 30 events when 15 were expected (authz test was 68 when 20 was expected).  My first response is half correct.  It was a copy and paste of the original sdk code, but it isn't an error.

File: app/app.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (channa-figure) [reply app/app.go]: This is now fixed throughout the whole PR.  Keep an eye out if I missed anything.

File: internal/handlers/msg_service_router.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (channa-figure) [reply internal/handlers/msg_service_router.go:158]: Yes, after running the tests without it. There were lots of duplicate event messages. For example, after removing it there was 30 events when 15 were expected (authz test was 68 when 20 was expected). My first response is half correct. It was a copy and paste of the original sdk code, but it isn't an error.

File: internal/handlers/pio_msg_fee_invoker.go [hunk 10]
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/provenance-io/provenance/internal/antewrapper"
+	msgfeestypes "github.com/provenance-io/provenance/x/msgfees/types"
+)
+
+type MsgFeeInvoker struct {
```
dwedul-figure, line 13: Is there a reason to have the `pio_` at the start of this filename? Usually it's nice to have the name match the struct.

dwedul-figure (commented):
Not fully done going through all files, but here's some notes so far.

internal/antewrapper/msg_fees_decorator_test.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	// setup NewMsgFeesDecorator
+	app := suite.app
+	mfd := antewrapper.NewMsgFeesDecorator(app.BankKeeper, app.AccountKeeper, app.FeeGrantKeeper, app.MsgFeesKeeper)
+	antehandler := sdk.ChainAnteDecorators(mfd)
+	return err, antehandler
+}
+
+func check(e error) {
```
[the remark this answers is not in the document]
arnabmitra: good point..changed

internal/handlers/pio_msg_fee_invoker.go
arnabmitra: was hoping to distinguish pio specific handler injection, probably better to stick to conventions..changed

internal/antewrapper/prov_feegrant_test.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+	suite.Require().NoError(err)
+
+	// grant low fee allowance (20atom), to check the tx requesting more than allowed.
+	err = app.FeeGrantKeeper.GrantAllowance(ctx, addr2, addr4, &feegrant.BasicAllowance{
+		SpendLimit: sdk.NewCoins(sdk.NewInt64Coin("atom", 20)),
+	})
+	suite.Require().NoError(err)
+
+	cases := map[string]struct {
```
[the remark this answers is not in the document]
arnabmitra: i agree on the non determinism part of maps, but just test cases mapped to test, so i think it's ok here imo.

internal/antewrapper/msg_fees_decorator_test.go
```diff
@@ -0,0 +1,343 @@
...
+func (suite *AnteTestSuite) TestEnsureMempoolAndMsgFees_6() {
+	err, antehandler := setUpApp(suite, true, "atom", 100)
+	tx, acct1 := createTestTx(suite, err, sdk.NewCoins(sdk.NewInt64Coin("nhash", 190500200)))
+	check(simapp.FundAccount(suite.app, suite.ctx, acct1.GetAddress(), sdk.NewCoins(sdk.NewCoin("nhash", sdk.NewInt(190500100)))))
+	hashPrice := sdk.NewDecCoinFromDec("nhash", sdk.NewDec(1))
+	highGasPrice := []sdk.DecCoin{hashPrice}
+	suite.ctx = suite.ctx.WithMinGasPrices(highGasPrice)
+	_, err = antehandler(suite.ctx, tx, false)
+	suite.Require().NotNil(err, "Decorator should not have errored for insufficient additional fee")
+	suite.Require().Contains(err.Error(), "not enough fees; after deducting fees required,got: \"-100atom,190500200nhash\", required fees: \"100atom,100000nhash\" = \"100000nhash\"(base-fee) +\"100atom\"(additional-fees): insufficient fee", "wrong error message")
```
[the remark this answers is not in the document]
arnabmitra: thanks!

proto/provenance/msgfees/v1/query.proto
```diff
@@ -0,0 +1,81 @@
...
+// Query defines the gRPC querier service for marker module.
+service Query {
+  // Params queries the parameters for x/msgfees
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/provenance/msgfees/v1/params";
+  }
+
+  // Query all Msgs which have fees associated with them.
+  rpc QueryAllMsgFees(QueryAllMsgFeesRequest) returns (QueryAllMsgFeesResponse) {
+    option (google.api.http).get = "/provenance/msgfees/v1/params";
```
dwedul-figure, line 25: This needs to be something different than the Params endpoint.
```suggestion
    option (google.api.http).get = "/provenance/msgfees/v1/all";
```

proto/provenance/msgfees/v1/proposals.proto
```diff
@@ -0,0 +1,62 @@
...
+
+  cosmos.base.v1beta1.Coin additional_fee = 4 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (gogoproto.moretags)     = "yaml:\"additional_fee\""
+  ];
+}
+
+// UpdateMsgFeeProposal defines a governance proposal to update a current msg based fee
+message UpdateMsgFeeProposal {
```
dwedul-figure, line 37: Is there a reason to split out the `Add` and `Update`? Seems like it'd be simpler to just have a `Set`.

dwedul-figure (commented):
A couple proto notes.

internal/antewrapper/prov_feegrant_test.go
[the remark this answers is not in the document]
dwedul-figure: It makes them harder to find when they fail. It also can cause issues where tests sometimes fail because sometimes one test goes first and changes something that another test has a problem with. We've run into that before in this repo.

x/msgfees/keeper/keeper.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		if handle(record) {
+			break
+		}
+	}
+	return nil
+}
+
+// DeductFees deducts fees from the given account, the only reason it exists is that the
+func (k Keeper) DeductFees(bankKeeper cosmosauthtypes.BankKeeper, ctx sdk.Context, acc cosmosauthtypes.AccountI, fees sdk.Coins) error {
```
dwedul-figure, line 137: Incomplete thought.

x/msgfees/types/proposal.go
```diff
@@ -0,0 +1,145 @@
...
+		return ErrEmptyMsgType
+	}
+
+	if !ambfp.AdditionalFee.IsPositive() {
+		return ErrInvalidFee
+	}
+
+	return govtypes.ValidateAbstract(&ambfp)
+}
+func (ambfp AddMsgFeeProposal) String() string {
```
dwedul-figure, line 63: Probably need to run `make lint`.

proto/provenance/msgfees/v1/query.proto
arnabmitra: good catch probably breaks reverse proxies

x/msgfees/keeper/keeper.go
arnabmitra: good catch ..complete comment 
```
// DeductFees deducts fees from the given account, the only reason it exists is that the
// cosmos method does not take in the custom fee collector which is a feature desired from msg fees.
```

proto/provenance/msgfees/v1/proposals.proto
arnabmitra: i would say Update governance proposal makes it clearer to voter's that an existing fee schedule is being changed..internally they can call the same thing imo..
@channa-figure chime in if you think differently.

x/msgfees/types/proposal.go
arnabmitra: only results i get from `make lint` are 
```
x/metadata/types/address.go:178:10: appendAssign: append result not assigned to the same slice (gocritic)
	addr := append(SessionKeyPrefix, bz...)
	        ^
x/metadata/types/address.go:192:10: appendAssign: append result not assigned to the same slice (gocritic)
	addr := append(RecordKeyPrefix, bz...)
	        ^
x/metadata/types/address.go:225:10: appendAssign: append result not assigned to the same slice (gocritic)
	addr := append(RecordSpecificationKeyPrefix, bz...)
	        ^
x/name/legacy/v042/store.go:35:17: appendAssign: append result not assigned to the same slice (gocritic)
		updatedKey := append(updateAddress, nameKey...)
		              ^
x/attribute/types/keys.go:34:9: appendAssign: append result not assigned to the same slice (gocritic)
	key := append(AttributeKeyPrefix, address.MustLengthPrefix(acc.Bytes())...)
	       ^
x/attribute/types/keys.go:46:9: appendAssign: append result not assigned to the same slice (gocritic)
	key := append(AttributeKeyPrefix, address.MustLengthPrefix(acc.Bytes())...)
	       ^
x/name/keeper/keeper.go:102:14: appendAssign: append result not assigned to the same slice (gocritic)
	indexKey := append(addrPrefix, key...) // [0x04] :: [addr-bytes] :: [name-key-bytes]
	            ^
x/name/keeper/keeper.go:199:14: appendAssign: append result not assigned to the same slice (gocritic)
	indexKey := append(addrPrefix, key...) // [0x02] :: [addr-bytes] :: [name-key-bytes]
	            ^
x/attribute/legacy/v042/types.go:21:9: appendAssign: append result not assigned to the same slice (gocritic)
	key := append(AttributeKeyPrefixLegacy, acc.Bytes()...)
	       ^
make: *** [lint] Error 1
```

internal/antewrapper/prov_feegrant_test.go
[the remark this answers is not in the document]
arnabmitra: ok changed to array..
i would say it should be a bad thing if they failed if they were run in a different order, cause tests should pass irrespective of order, but i changed it 🤷  agreed they are harder to debug but also them breaking because of order is code smell or test code smell at the very least.

x/msgfees/spec/01_concepts.md
```diff
@@ -0,0 +1,81 @@
+<!--
+order: 1
+-->
+
+# Concepts
+
+
+
+# Additional Msg Fees
```
dwedul-figure, line 9: These should all be level 2 headers. The only level 1 header should be the one at the top for the whole page.
```suggestion
## Additional Msg Fees
```

x/msgfees/spec/01_concepts.md
```diff
@@ -0,0 +1,81 @@
...
+Total fees continue to be passed as `sdk.Coins` the Tx accepts for fee entry currently.
+e.g usd.example is the denom(assuming there is a marker/coin of type usd.example) in which additional fee is being charged in
+```bash
+--fees 382199010nhash,99usd.example 
+```
+
+# Additional fee assessed in base denom i.e nhash
+To preserve backwards compatability of all invokes, clients continue accepting fees in sdk.Coins
+`type Coins []Coin`, and because the code needs to distinguish between base fee and additional fee,
+the msgfees module introduces an additional param, described in 06_params.md, called `DefaultFloorGasPrice`
```
dwedul-figure, line 36: A link would be pretty handy here on the `06_params.md` text.

x/msgfees/spec/02_state.md
```diff
@@ -0,0 +1,23 @@
...
+message MsgFee {
+string msg_type_url = 1;
+// additional_fee can pay in any Coin( basically a Denom and Amount, Amount can be zero)
+cosmos.base.v1beta1.Coin additional_fee = 2 [
+(gogoproto.nullable)     = false,
+(gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+(gogoproto.moretags)     = "yaml:\"additional_fee\""
+];
+}
+```
```
dwedul-figure, line 21: There should be a link to the message in the proto file in here somewhere. See other state docs for examples.

x/msgfees/spec/04_messages.md
```diff
@@ -0,0 +1,51 @@
+<!--
+order: 4
+-->
+
+
+
+## Msg/GenesisState
+
```
dwedul-figure, line 8: Needs a level 1 header.

x/msgfees/spec/04_messages.md
```diff
@@ -0,0 +1,51 @@
+<!--
```
dwedul-figure, line 1: This file should be called `xx_queries.md` since it's about the queries. The `xx_messages.md` docs are about the transaction messages.

x/msgfees/spec/05_events.md
```diff
@@ -0,0 +1,25 @@
+<!--
+order: 5
+-->
+
+# Events
+
+Existing fee event continue to show total fee charged
+### Any Tx
```
dwedul-figure, line 8: Usually headings only go up by 1 at each step.

dwedul-figure (commented):
Docs could use some love. It's good to have consistency both internally and with the other module docs.

The longer ones should have a table of contents. There's a couple easy ways to do it:
* Throw `<!-- TOC -->` at the top after the level 1 header and blurb, but before the next header. Then run `make update-tocs` to have them automatically generated.
* Or, just run `scripts/update-toc.sh <file1> [<file2> ...]` and it'll create and add the TOCs in the standard spot on those files.

There also needs to be a `README.md` in the `spec` folder.

x/msgfees/types/proposal.go
dwedul-figure: The `make lint` command does 3 things.
1. Runs `golang-ci`.
2. Runs `gofmt -s -d` on the go files.
3. Runs `go mod verify`.

Those are from `golang-ci`, and they prevent the other 2 things from running. I'm guessing you've got a version of `golang-ci` other than `v1.39`, which is why you're getting these errors, but the github action isn't.

The `gofmt` step should be complaining about not having an empty line here.

A run of `make lint` shouldn't identify anything. It should just look like this:
```
$ make lint
/usr/local/bin/golangci-lint run
find . -name '*.go' -type f -not -path "./vendor*" -not -path "./client/*" -not -path "*.git*" -not -path "*.pb.go" | xargs gofmt -d -s
/usr/local/bin/go mod verify
all modules verified
```

dwedul-figure (dismissed):
LGTM

dwedul-figure (approved):
LGTM

iramiller (commented):
Feels like the "enable governance" flow really isn't required for this module ... there isn't any way to use it without governance for setting the fee table (I suppose you could use a software upgrade to do it ... ). The "enable governance" flow was added to the marker module so that some markers could be locked permanently (requiring a software fork to change) ... example would be the staking currency.

x/msgfees/keeper/query_server.go
```diff
@@ -0,0 +1,84 @@
...
+	gasMeter, ok := txCtx.GasMeter().(*antewrapper.FeeGasMeter)
+	if !ok {
+		return nil, fmt.Errorf("unable to extract fee gas meter from transaction context")
+	}
+	baseDenom := k.defaultFeeDenom
+	if request.DefaultBaseDenom != "" {
+		baseDenom = request.DefaultBaseDenom
+	}
+
+	minGasPrice := int64(k.GetFloorGasPrice(ctx))
```
iramiller, line 71: similar to other comment, using a `sdk.Coin` here would be significantly safer then what is happening here (casting a `uint32` to `int64`)

x/msgfees/types/expected_keepers.go
```diff
@@ -0,0 +1,36 @@
...
+	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
+}
+
+// MsgFeesKeeper for additional msg fees.
+type MsgFeesKeeper interface {
+	GetMsgFee(ctx sdk.Context, msgType string) (*MsgFee, error)
+	GetFeeCollectorName() string
+	DeductFees(bankKeeper authtypes.BankKeeper, ctx sdk.Context, acc authtypes.AccountI, fees sdk.Coins) error
+	GetDefaultFeeDenom() string
+	GetFloorGasPrice(ctx sdk.Context) uint32
```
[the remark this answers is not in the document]
arnabmitra: yeah agree, changed this to sdk.Coin

x/msgfees/keeper/query_server.go
arnabmitra: right looks like `minGasPrice := k.GetFloorGasPrice(ctx)` based on the change from ^^ comment

internal/handlers/msg_service_router.go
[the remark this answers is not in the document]
iramiller: I actually remember that bug and this line was a fix for that now. :doh:

x/msgfees/keeper/keeper.go
```diff
@@ -61,17 +61,13 @@ func (k Keeper) GetFeeCollectorName() string {
...
-func (k Keeper) GetDefaultFeeDenom() string {
-	return k.defaultFeeDenom
-}
-
-// GetFloorGasPrice  returns the current minimum gas price used in calculations for charging additional fees
-func (k Keeper) GetFloorGasPrice(ctx sdk.Context) (min uint32) {
-	min = types.DefaultFloorGasPrice
+// GetFloorGasPrice  returns the current minimum gas price in sdk.Coin used in calculations for charging additional fees
+func (k Keeper) GetFloorGasPrice(ctx sdk.Context) sdk.Coin {
+	var min uint32 = types.DefaultFloorGasPrice
```
iramiller, line 66: If this type was also a Coin it would drastically simplify some of these constructs...

x/msgfees/keeper/params.go
```diff
@@ -9,7 +9,7 @@ import (
 // GetParams returns the total set of distribution parameters.
 func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
 	return types.Params{
-		FloorGasPrice: k.GetFloorGasPrice(ctx),
+		FloorGasPrice: uint32(int32(k.GetFloorGasPrice(ctx).Amount.Uint64())),
```
iramiller, line 12: See also here ... if that Param was a coin then this gets much easier.

arnabmitra: hmmm..i was trying not to include the default denom here (since it is specified via another config) but i'll change the param to sdk.Coin now

x/msgfees/keeper/keeper.go
arnabmitra: agreed..changing now

arnabmitra: changed

x/msgfees/keeper/params.go
arnabmitra: Param is of `sdk.Coin` type now

internal/antewrapper/msg_fees_decorator.go
```diff
@@ -0,0 +1,240 @@
...
+		deductFeesFromAcc := afd.accountKeeper.GetAccount(ctx, deductFeesFrom)
+		if deductFeesFromAcc == nil {
+			err = sdkerrors.Wrapf(sdkerrors.ErrUnknownAddress, "fee payer address: %q does not exist", deductFeesFrom)
+			if err != nil {
+				return ctx, err
+			}
+		}
+
+		// get all the coin balances for the fee payer account
+		balancePerCoin := make(sdk.Coins, len(feeCoins))
```
iramiller, line 96: You can use `sdk.NewCoins()` here... this has the side benefit of ensuring the coin array is sorted which can prevent some determinism issues.

internal/antewrapper/msg_fees_decorator.go
```diff
@@ -0,0 +1,240 @@
...
+		msgFees, err := mbfk.GetMsgFee(ctx, typeURL)
+		if err != nil {
+			return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
+		}
+
+		if msgFees == nil {
+			continue
+		}
+		if msgFees.AdditionalFee.IsPositive() {
+			additionalFees = additionalFees.Add(sdk.NewCoin(msgFees.AdditionalFee.Denom, msgFees.AdditionalFee.Amount))
```
iramiller, line 227: If AdditionalFee is already a coin, why are we creating a new coin around it here?

internal/antewrapper/msg_fees_decorator.go
```diff
@@ -0,0 +1,240 @@
...
+func EnsureSufficientFees(gas uint64, feeCoins sdk.Coins, additionalFees sdk.Coins,
+	minGasPriceForAdditionalFeeCalc sdk.Coin) error {
+	// Step 1. Check if fees has enough money to pay additional fees.
+	var hasNeg bool
+	if feeCoins, hasNeg = feeCoins.SafeSub(additionalFees); hasNeg {
+		return sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, DefaultInsufficientFeeMsg+": %q, required additional fee: %q", feeCoins, additionalFees)
+	}
+	// Step 2: check if additional fees in nhash, that base fees and additional fees can be paid
+	// total fees in hash - gas limit * price per gas >= additional fees in hash
+	if containsDenom(additionalFees, minGasPriceForAdditionalFeeCalc.Denom) {
```
iramiller, line 198: cointainsDenom seems to be used once? It can be replaced with one of the two `sdk.Coin` functions that return the amount of a denom
```
// AmountOf returns the amount of a denom from coins
func (coins Coins) AmountOf(denom string) Int {
	mustValidateDenom(denom)
	return coins.AmountOfNoDenomValidation(denom)
}

// AmountOfNoDenomValidation returns the amount of a denom from coins
// without validating the denomination.
func (coins Coins) AmountOfNoDenomValidation(denom string) Int {
	switch len(coins) {
	case 0:
		return ZeroInt()

	case 1:
		coin := coins[0]
		if coin.Denom == denom {
			return coin.Amount
		}
		return ZeroInt()

	default:
		// Binary search the amount of coins remaining
		midIdx := len(coins) / 2 // 2:1, 3:1, 4:2
		coin := coins[midIdx]
		switch {
		case denom < coin.Denom:
			return coins[:midIdx].AmountOfNoDenomValidation(denom)
		case denom == coin.Denom:
			return coin.Amount
		default:
			return coins[midIdx+1:].AmountOfNoDenomValidation(denom)
		}
	}
}
```

iramiller (commented):
Just a couple minor points in one last place, the `msg_fees_decorator`.  In general I was looking to see that we do not create extra coins and when possible we are using the coins/sdk.int API functions as these are exceedingly well tested and include many safeguards for handling.

The sdk.Coins array itself is something we need to handle carefully since depending on how it is used there can be issues when the array is not properly sorted.  This can happen (in at least one instance) because the type can be created from a proto deserialization which will not always have sorted coins.

internal/antewrapper/msg_fees_decorator.go
arnabmitra: good catch ..changed

internal/antewrapper/msg_fees_decorator.go
arnabmitra: i changed to this ```balancePerCoin := sdk.NewCoins()
		for _, fc := range feeCoins {
			balancePerCoin.Add(afd.bankKeeper.GetBalance(ctx, deductFeesFrom, fc.Denom))
		}
```  i think i was missing using the `Add` here which should make it sorted

File: internal/antewrapper/msg_fees_decorator.go (hunk 30 above)
Comment (arnabmitra) [reply internal/antewrapper/msg_fees_decorator.go]: Deleted method, moved the `AmountOf(` call from that method to the if statement

**Review by iramiller — APPROVED** [review]
LGTM

**Review by dwedul-figure — APPROVED** [review]
LGTM

File: x/msgfees/simulation/params.go [hunk 31]
```diff
@@ -0,0 +1,39 @@
+package simulation
+
+// DONTCOVER
```
fkneeland-figure, line 3: What is this for?

fkneeland-figure (approved):

x/msgfees/simulation/params.go
arnabmitra: believe it is to ignore coverage, based on code in simulation pkg, since they are only used for simulation tests. (every module in our project has it)

PR #521, opened 2021-10-04
Fix devnet PIO_HOME value and simplify the container.

dwedul-figure:
## Description

* Fix the `devnet` environment variable `PIO_HOME` to have the correct value for the devnet node. This makes it so you can run commands like `docker exec dev-node provenanced config changed` and it'll work as expected. Similarly, if you `docker exec -it dev-node /bin/sh` into the container, the `provenanced` commands will be using the proper `home` and behave as expected.
* Update `make devnet-generate` to list the node's keys instead of generating testnet nodes that are then not used. The `entrypoint.sh` script handles initializing things if there's no genesis file, and that's the main thing we want done with `make devnet-generate`. The `keys list` command outputs possibly useful info, but is mostly just a harmless command that can be run to trigger the initialization stuff in `entrypoint.sh` (if it hasn't been done yet).

---

write a little note why.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -387,10 +387,9 @@ localnet-stop:
 docker-build-dev: vendor
 	docker build --tag provenance-io/blockchain-dev -f networks/dev/blockchain-dev/Dockerfile .
 
-
 # Generate config files for a single node devnet
 devnet-generate: devnet-stop docker-build-dev
-	@if ! [ -f build/nodedev/config/genesis.json ]; then docker run --rm -v $(CURDIR)/build:/provenance:Z provenance-io/blockchain-dev testnet --v 1 -o . --starting-ip-address 192.168.21.2 --keyring-backend=test --chain-id=chain-dev ; fi
+	docker run --rm -v $(CURDIR)/build:/provenance:Z provenance-io/blockchain-dev keys list
 
 # Run a single node devnet locally
 devnet-up:
--- a/networks/dev/docker-compose.yml
+++ b/networks/dev/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     environment:
       - ID=dev
       - LOG=${LOG:-provenanced.log}
-      - PIO_HOME=/provenance
+      - PIO_HOME=/provenance/nodedev
     volumes:
       - ./build:/provenance:Z
 


Makefile
```diff
@@ -387,10 +387,9 @@ localnet-stop:
 docker-build-dev: vendor
 	docker build --tag provenance-io/blockchain-dev -f networks/dev/blockchain-dev/Dockerfile .
 
-
 # Generate config files for a single node devnet
 devnet-generate: devnet-stop docker-build-dev
-	@if ! [ -f build/nodedev/config/genesis.json ]; then docker run --rm -v $(CURDIR)/build:/provenance:Z provenance-io/blockchain-dev testnet --v 1 -o . --starting-ip-address 192.168.21.2 --keyring-backend=test --chain-id=chain-dev ; fi
+	docker run --rm -v $(CURDIR)/build:/provenance:Z provenance-io/blockchain-dev keys list
```
iramiller, line 392: this doesn't generate a testnet config anymore?

dwedul-figure: Nope. Not for the single-node `devnet` here. The `networks/dev/blockchain-dev/entrypoint.sh` script will call `init` and other setup commands similar to the `make run-config` target (if it hasn't already been done). The other `devnet` docker stuff is then set up to use that stuff rather than anything created by a `testnet` command.

So basically, before this change, the `make devnet-generate` target was using `init` etc. to set things up, then creating a separate 1 node `testnet`, then ignoring all the `testnet` stuff.

Additionally, if you did `make clean`, then `make localnet-generate`, then `make devnet-generate`, you'd get an error like `Error: account tp1ktqzf8mq0v2k8a2yqwa2agvy28exnr80qrs9a7 balance not in genesis state: map[tp1pz38u2vrzfcwn2h25v5gy2r4zdqcnqvxnw6zu6:{Address:tp1pz38u2vrzfcwn2h25v5gy2r4zdqcnqvxnw6zu6 Coins:100000000000000000000nhash}]`. The `gentxs` directory made by the `testnet` call as part of `make localnet-generate` has 4 nodes-worth of genesis transactions, including addresses that aren't know the second time around with `make devnet-generate`.

iramiller: Sounds like a bug in the testnet command generating extra gentx... we should put something into clean all this up I think... the init commands are useful for a single local node... testnet is supposed to generate files in the way you build an actual network (I.e gentx would be used for validators to protect keys).  I suspect that the generate commands are the ones to not follow convention here as the 'config' targets were put in first for the testnet and 'run-config' approach. Definitely getting off topic here, a new issue to properly cleanup all of this should be filed I am thinking.

iramiller (dismissed):
I don't know if we really need a change log entry for build script changes ... it feels like that should be limited to changes in the actual software itself (aka what gets compiled).  Regardless this is a good fix to have.

Makefile
iramiller: Started an independent issue to collect notes on the overall process here: https://github.com/provenance-io/provenance/issues/522

Issue #578, opened 2021-12-22
Update dependencies for 1.8.x release


iramiller:
## Summary

Refresh dependencies for upcoming 1.8.x release

## Proposal

- Update Cosmos SDK to 0.44.5
   - Update Provenance Cosmos fork to pull from 0.44.5
- Update IBC to 2.0.2
- Update provenance/wasmd to reference above versions

Add a migration for fixing existing IBC configurations as part of upgrade handler "Green".  Initial fix is for allow update after expire, add any other configuration changes are required based on Relayer testing.

Roughly:
```
	ibcconnectiontypes "github.com/cosmos/ibc-go/v2/modules/core/03-connection/types"
	ibccoretypes "github.com/cosmos/ibc-go/v2/modules/core/exported"
	ibcctmtypes "github.com/cosmos/ibc-go/v2/modules/light-clients/07-tendermint/types"
	
	[...]
			app.IBCKeeper.ClientKeeper.IterateClients(ctx, func(clientId string, state ibccoretypes.ClientState) bool {
				tc, ok := (c).(*ibcctmtypes.ClientState)
				if ok {
					tc.AllowUpdateAfterExpiry = true
					app.IBCKeeper.ClientKeeper.SetClientState(ctx, tc., c)
				}
				return false
			})
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #598, opened 2022-01-13
Edit upgrade handler to allow expired ibc connections to be updated

fkneeland-figure:
## Description

In this pr I added an upgrade handler to iterate over IBC clients and to `AllowUpdateAfterExpiry` to enable us to restart expired IBC clients. 

Issue: #578 

---

write a little note why.


fkneeland-figure:
I tested the following using the [ibc relayer service](https://github.com/FigureTechnologies/service-ibc-relayer):

- [ ] A Provenance upgrade with no IBC connections
- [ ] A Provenance upgrade with up to date IBC connections
- [ ] A Provenance upgrade with expired IBC connections

fkneeland-figure:
However, the current state of gaia `v6.0.0` wouldn't allow IBC clients to be restarted leaving me with the following msg:

```
Error: foreign client error: error raised while updating client on chain ibc-1: failed building header with error: Light client error for RPC address ibc-1: trusted state outside of trusting period
```

In this scenario ibc-1 was the gaia chain.

app/upgrades.go
```diff
@@ -127,6 +129,14 @@ var handlers = map[string]appUpgrade{
 	},
 	"green": {
 		Handler: func(app *App, ctx sdk.Context, plan upgradetypes.Plan) (module.VersionMap, error) {
+			app.IBCKeeper.ClientKeeper.IterateClients(ctx, func(clientId string, state exported.ClientState) bool {
+				tc, ok := (state).(*ibcctmtypes.ClientState)
+				if ok {
```
arnabmitra, line 134: could potentially check if already true and skip the set, could help if we have a lot of them :notsureif:..but probably not required

arnabmitra (approved):

iramiller (approved):
LGTM ... I don't think we need to optimize this much given there are only a very small handful of connections being iterated over.

app/upgrades.go
iramiller: I think there are only a half dozen total connections we are iterating over here... seems fine to brute force here.

fkneeland-figure: True, however we have very few ibc connections and the time required to do a state update should be very minimal so I don't know that it would be necessary... 🤷‍♂️

Issue #614, opened 2022-01-24
Add Protobuf support with buf.build


egaxhaj-figure:
## Summary

Add support for compiling protos with https://buf.build

## Problem Definition

**Why do we need this feature?**

Allow client developers to generate code for common `gRPC` supported languages other than `golang`.

**What problems may be addressed by introducing this feature?**

Support multiple language code generation.

**What benefits does the Provenance Network stand to gain by including this feature?**

Grow client application developer community to multiple languages.

## Proposal

Use github actions to publish new releases of `proto` files to https://buf.build for easy generation of client code into supported languages.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #630, opened 2022-02-01
614 proto binding support with buf build

egaxhaj-figure:
## Description

closes: #614

---

write a little note why.


third_party/proto/cosmos/bank/v1beta1/query.proto
```diff
@@ -13,7 +13,7 @@ option go_package = "github.com/cosmos/cosmos-sdk/x/bank/types";
 service Query {
   // Balance queries the balance of a single coin for a single account.
   rpc Balance(QueryBalanceRequest) returns (QueryBalanceResponse) {
-    option (google.api.http).get = "/cosmos/bank/v1beta1/balances/{address}/{denom}";
+    option (google.api.http).get = "/cosmos/bank/v1beta1/balances/{address}/by_denom";
```
egaxhaj-figure, line 16: @iramiller @arnabmitra @dwedul-figure I noticed this change in the URL from `/{denom}` to `/by_denom` on our cosmos-sdk fork version [v0.45.0-pio-1](https://github.com/provenance-io/cosmos-sdk/blob/f672cad373eccafefbe58655aae8dcf90d8b1b1b/proto/cosmos/bank/v1beta1/query.proto#L16) which we are referencing in `go.mod`. Is this change intentional?

iramiller: That's in the upstream as well... they had to do this because forward slash is allowed in a denom and the `ibc/` denoms broke this.

egaxhaj-figure: Ah.

iramiller:
`make proto-all` is giving a bunch of errors in the lint step that seem to be related to a path issue for includes...

scripts/proto-update-deps.sh
```diff
@@ -0,0 +1,50 @@
...
+PROTO_EXPR="*/proto/**/*.proto"
+curl -sSL "$CONFIO_PROTO_URL" -o proto/proofs.proto.orig --create-dirs || exit 1
+curl -sSL "$GOGO_PROTO_URL" -o proto/gogoproto/gogo.proto --create-dirs || exit 1
+curl -sSL "$COSMOS_PROTO_URL" -o proto/cosmos_proto/cosmos.proto --create-dirs || exit 1
+curl -sSL "$COSMWASM_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" "$PROTO_EXPR" || exit 1
+curl -sSL "$COSMWASM_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" "$PROTO_EXPR" || exit 1
+curl -sSL "$WASMD_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" --exclude="*/proto/ibc" "$PROTO_EXPR" || exit 1
+curl -sSL "$IBC_GO_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" "$PROTO_EXPR" || exit 1
+curl -sSL "$COSMOS_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" --exclude="*/testutil" "$PROTO_EXPR" || exit 1
+curl -sSL "$TM_TARBALL_URL" | tar zx --strip-components 1 --exclude="*/third_party" "$PROTO_EXPR" || exit 1
```
dwedul-figure, line 37: All of this comment is just nit stuff and bash tips.

You've got `set -e` at the top, so you don't need the `|| exit 1` on any of these lines. Also, unless there's a reason to have a specific exit code, doing `|| exit $?` is more desirable. The `$?` variable is the exit code of the last thing executed. Changing it to `set -ex` might be nice too. the `-x` will cause it to output commands as it executes them. That gives a nice paper-trail for when things go wrong (as well as making it easy to see what it's doing). Alternatively, there's the `-v` flag which will also output commands as it executes them, but it doesn't expand the variables in the commands and generally outputs a little less stuff. But `-x` is closer to default behavior from `make`.

Because of line 8: `dir="$(pwd)"`, this assumes you're running the script from the root of the repo. That's fine because we'll probably be running this using `make` which means we'd be in the repo root. But if you wanted it to work from any directory, you could do this:
```
dir="$( cd "$( dirname "${BASH_SOURCE:-$0}" )/.."; pwd -P )"
```
That then assumes that this script file is in the `scripts/` directory (or really 1 directory inside the root of the repo). But executing the script from anywhere (e.g. while currently in the `scripts/` directory) would then always update the desired location.

egaxhaj-figure: Thank You! This is great. 

`dir="$( cd "$( dirname "${BASH_SOURCE:-$0}" )/.."; pwd -P )"`. This will help when running protoBindings for java/kotling

iramiller (approved):
LGTM, verified works locally.

channa-figure (approved):
LGTM

arnabmitra (approved):

Issue #681, opened 2022-02-24
Transaction size restrictions


iramiller:
## Summary

Currently the blockchain has very high limits for full block size (21MB, 60M gas  on mainnet and 21MB, no gas limit on testnet).  In addition there is no limit on the number of message or the amount of gas spent within a single transaction.  There should be more reasonable limits set on both of these values

## Problem Definition

Without limits on transactions a single one can take up an entire block.  In addition the resource consumption for a single block is high enough to prevent smooth operation of the network.

## Proposal

Update an upgrade proposal within the 1.8.0-RC framework that lowers the block size limit and enforces a maximum amount of gas per block.  In addition modify the antehandler such that a single tx can account for no more than 50% of a given block's resources.

For reference:
```
// Osmosis
"max_bytes": "5242880",
"max_gas": "6000000",
// Cosmos Hub
"max_bytes": "200000",
"max_gas": "2000000",
// Provenance
"max_bytes": "22020096",
"max_gas": "60000000",
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
Roughly the following for the 
```
params := app.GetConsensusParams(ctx)
params.Block.MaxBytes = 1024 * 1024 * 5 // 5MB
params.Block.MaxGas = 60_000_000        // note that 5MB for a block represents ~52M gas @ 10 per byte.
app.StoreConsensusParams(ctx, params)
```

PR #683, opened 2022-02-25
Set a limit on tx gas as a percentage of block maximum

iramiller:
## Description

Adds an antehandler to ensure it is possible to send at least 20tx per block by enforcing a limit of 5% of the block limit on any single transaction

closes: #681 

---

write a little note why.


fkneeland-figure (dismissed):

dwedul-figure (dismissed):
LGTM

internal/antewrapper/tx_gas_limit_decorator.go
```diff
@@ -0,0 +1,39 @@
...
+func (mfd TxGasLimitDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+	// Ensure that the requested gas does not exceed the configured block maximum
+	gas := feeTx.GetGas()
+	gasTxLimit := uint64(0)
+	if ctx.BlockGasMeter() != nil {
+		gasTxLimit = ctx.BlockGasMeter().Limit() * (1 / MinTxPerBlock)
```
arnabmitra, line 31: 1/20 will lead to 0.05 which maybe should be used, i am little worried about determinism of floating point division on gol-lang if say the MinTxPerBlock changes to 33.. i maybe overthinking this for this case but 🤷

iramiller: The current layout was intended to be as readable as possible and when I originally used the decimal value here it seemed less clear what we were targeting.  

As half the equation is a contstant (1/20) that part should be safe ... but we have a variable gas limit through config and that might cause issues if we used a non-evenly divisible value...

arnabmitra: > The current layout was intended to be as readable as possible and when I originally used the decimal value here it seemed less clear what we were targeting.
> 
> As half the equation is a contstant (1/20) that part should be safe ... but we have a variable gas limit through config and that might cause issues if we used a non-evenly divisible value...

i  updated the comment, i don't think ^^ calc works (unless i am missing something)

iramiller: Updatesd with a simplified expression works regardless of the limit specified.

channa-figure (approved):

fkneeland-figure (dismissed):

fkneeland-figure:
Looks like tests are failing even after a re-run:
```
--- FAIL: TestSimTestSuite (0.15s)
    --- FAIL: TestSimTestSuite/TestSimulateMsgBindName (0.06s)
        operations_test.go:89: 
            	Error Trace:	operations_test.go:89
            	Error:      	Should be true
            	Test:       	TestSimTestSuite/TestSimulateMsgBindName
    --- FAIL: TestSimTestSuite/TestSimulateMsgDeleteName (0.04s)
        operations_test.go:[118](https://github.com/provenance-io/provenance/runs/5383434622?check_suite_focus=true#step:6:118): 
            	Error Trace:	operations_test.go:118
            	Error:      	Should be true
            	Test:       	TestSimTestSuite/TestSimulateMsgDeleteName
```

arnabmitra (approved):
lgtm!

fkneeland-figure (approved):

PR #762, opened 2022-03-31
Fix proto-update-deps script

jphorec:
## Description

Conf.io changed their release naming for the ics123 library which `proofs.proto` comes from [as seen here](https://github.com/confio/ics23/releases). This change breaks the `proto-update-deps` script in the `scripts/` directory.  I have updated the path in the script. 

I also noticed that the bash script was not exiting correctly on error which was resulting in the `proofs.proto` file to be overwritten with `404: File Not Found`. I have updated the curl commands to use the `-f` flag and also removed the `|| exit` conditions since the bash script is already setup with `set -e` which will fail on any non 0 exits. 

---

write a little note why.

### DIFF
--- a/scripts/proto-update-deps.sh
+++ b/scripts/proto-update-deps.sh
@@ -13,7 +13,7 @@ EXT_PROTO_DIR="$dir"/third_party
 #rm -rf "${EXT_PROTO_DIR:?}/"*
 
 # Retrieve versions from go.mod (single source of truth)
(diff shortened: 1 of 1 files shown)


dwedul-figure (approved):
LGTM.

+1 for getting rid of those unneeded `|| exit $?`.

iramiller (approved):

dwedul-figure:
> Merging is blocked
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

Closing this PR and recreating it with signed commits: https://github.com/provenance-io/provenance/pull/775

Thanks for the fix!

Issue #842, opened 2022-05-27
Unblock proto update check in workflow


egaxhaj:
## Summary of Bug

An additional check needs to occur to unblock the workflow from firing `exit code 1` when checking to see if third_party proto files require updates. This check needs to verify that proto files were updated an no further changes are required resulting in the workflow passing.

This is continuation work on #838 that resulted in an auto-merge.
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #843, opened 2022-05-27
unblock proto update check

egaxhaj:
1. Checks `go.mod` for version changes in `cosmos-sdk` and `tendermint` libraries
2. Pulls down the Protobuf files for those versions and runs a `diff` against the Protobuf files in `main`
3. Logs error message with recommended steps to update to the latest Protobuf files and exits workflow.
4. If no update is necessary
    - runs`buf lint`
    - runs `buf breaking`

closes: #842 

---

write a little note why.


scripts/proto-update-check.sh
```diff
@@ -6,23 +6,39 @@ set -ex
...
+    echo -e "Found differences in Protobuf files.\n
+    This indicates a version change was detected in one of the following libraries:
+      - github.com/cosmos/cosmos-sdk
+      - github.com/tendermint/tendermint
+
+    Review the diff log above and update accordingly. Perform the following steps locally.
+      1. run: make proto-update-deps
+      2. run: make proto-update-check
+      3. run: diff -rq -x '*.yaml' --exclude=google build/third_party third_party
+      4. commit updates and push\n" >&2
```
dwedul-figure, line 35: Here also with the `echo`, `printf`, and `cat << EOF` thing.
```suggestion
    printf '\n\nDiff log:\n\n%s\n\n" "$DIFF"

    cat << EOF >&2
Found differences in Protobuf files.
This indicates a version change was detected in one of the following libraries:
  - github.com/cosmos/cosmos-sdk
  - github.com/tendermint/tendermint

Review the diff log above and update accordingly. Perform the following steps locally.
  1. run: make proto-update-deps
  2. make sure this then passes: make proto-update-check
  3. commit updates and push
EOF
```

dwedul-figure (commented):
There's a few things that would be nice to have updated for better portability and clarity.

.github/workflows/proto.yml
```diff
@@ -20,6 +27,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - name: Setup go
         uses: actions/setup-go@v3
         with:
```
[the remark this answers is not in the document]
dwedul-figure: I just discoverd `if: always()`, and kind of like that over firing up separate runners for each simple check.
This is what I ended up with: [PR 845's proto.yml](https://github.com/provenance-io/provenance/blob/7c91b77be06529288a563eb4c4e2450be2533998/.github/workflows/proto.yml).

Why did you add the `fetch-depth: 0` here?

scripts/proto-update-check.sh
```diff
@@ -6,23 +6,39 @@ set -ex
...
+# single brackets because of `set -e` option above.
+if [ "$output" -gt "0" ]; then
+  echo -e "Downloading latest third_party proto files for comparison..."
+
+  # Download third_party proto files int build/ directory for comparison against $dir /third_party
+  bash "$dir"/scripts/proto-update-deps.sh build
+
+  echo -e "\nChecking Protobuf files for differences...\n"
+
+  DIFF=$(diff -rq -x '*.yaml' --exclude=google "$dir"/build/third_party "$dir"/third_party || true)
```
dwedul-figure, line 22: So yeah.... This is a significantly better solution than the one I had in mind. My thoughts were overly complicated and would probably be prone to breaking.

👍 🥇 💯

dwedul-figure:
The `third_party breaking` step that is part of the `Protobuf Checks` workflow is currently failing because the `GrantAuthorization` message definition was moved from `proto/cosmos/authz/v1beta1/genesis.proto` to `proto/cosmos/authz/v1beta1/authz.proto`. However, neither file is new, and both are part of the same package, so it's not actually breaking anything.

I double checked this by regenerating the java/kotlin code (`cd protoBindings; ./gradlew assemble`), which resulted in the same generated code as before.

So in this PR, that specific failure can be ignored.

dwedul-figure (approved):

derekadams (approved):

PR #850, opened 2022-06-06
Update the marker module to fix the bank SendEnabled issue.

dwedul-figure:
## Description

1. Bump Cosmos-SDK to `v0.45.4-pio-1`, which contains the bank module's `SendEnabled` fix.
2. Update the marker module to correctly manage the `SendEnabled` entries.
3. Some `Makefile` devtool cleanup with due to `go` changes.
4. Regenerate swagger config. It only includes the change that added the `restricted` field to name queries.
5. Update the `third_party` bank protos.

Related to Cosmos-SDK issue: [11859](https://github.com/cosmos/cosmos-sdk/issues/11859)

---

write a little note why.


x/marker/keeper/marker.go
```diff
@@ -701,20 +701,19 @@ func (k Keeper) RemoveSendEnabledForMissingMarkers(ctx sdk.Context) {
 	ctx.Logger().Info("Removing SendEnabled for Missing Markers...")
```
iramiller, line 701: Not sure this function is safe ... it seemingly only works by convention --- that there will be (or was) a marker with every send enabled entry ... but perhaps there could be one at some point from say an IBC denom...

x/marker/keeper/marker.go
```diff
@@ -787,49 +786,11 @@ func (k Keeper) accountControlsAllSupply(ctx sdk.Context, caller sdk.AccAddress,
...
-	updated := make([]*banktypes.SendEnabled, 0)
-	for i, se := range bankParams.SendEnabled {
-		if bankParams.SendEnabled[i].Denom != denom {
-			updated = append(updated, se)
-		}
-	}
-	if len(updated) != len(bankParams.SendEnabled) {
-		bankParams.SendEnabled = updated
-		k.bankKeeper.SetParams(ctx, bankParams)
+	se, found := k.bankKeeper.GetSendEnabledEntry(ctx, denom)
```
iramiller, line 792: do we want an entry created here in all cases?

dwedul-figure: Hmm. I was thinking we did. But having them for all cases really only prevents a problem if the default is ever changed. And I see now that we didn't create them in all cases previously. So I'll tweak it to only store entries contrary to the default.

x/marker/keeper/marker.go
dwedul-figure: I didn't look into usage of this function. I guessed that it was from a migration. I'll look into removing it.

dwedul-figure: Yeah. It's only called as part of the `lava` migration. I'll create an issue to clean up the upgrades and remove upgrade-only code.

iramiller: Hmm. I wish we had put upgrade in that name then... but yeah, good idea to remove now so it doesn't get used on accident

dwedul-figure: Done.

iramiller (approved):
LGTM

derekadams (approved):
LGTM

Issue #903, opened 2022-06-28
Fix the MarkerTransferAuthorization Accept function and TransferCoin authz handling to prevent problems when other authorization types are used


egaxhaj:
## Summary of Bug

When there is a `MsgTransferRequest` the `Accept` method decrements funds and returns an updated `MarkerTrasnferAuthorization` with the new limit amount when there exists an `msg.administrator`. Then the funds are again decremented in the `authzHandler` function in the Marker module. This may lead to double decrementing when the transfer type is something other than `MsgTransferRequest`.

Line 710:
https://github.com/provenance-io/provenance/blob/1283582fdc559857ce2d5fc1af3a39a904abd471/x/marker/keeper/marker.go#L699-L717

We're deleting here when `limitLeft.IsZero()` and ignoring `accept.Delete` which is set by `shouldDelete` (see below). If we  check for `if accept.Delete { ... }` it helps avoid `type casting` and allows for other authorization types to work as well.

The logic in `authzHandler()` and `MarkerTransferAuthorization.Accept()` are confusing. 

https://github.com/provenance-io/provenance/blob/1283582fdc559857ce2d5fc1af3a39a904abd471/x/marker/types/authz.go#L26-L44

Line 40: Why is the update occurring in the `marker` module?

## Steps to Reproduce

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

dwedul-figure:
After some research, it looks like this is working as intended, except accidentally, and in such a way that would be easy to have things go wrong.

In the `Accept` method, the limit is decremented if the `from == admin`. In the `TransferCoin` endpoint, we only call `Accept` when `from != admin`. But then, in the `TransferCoin` endpoint, if `from != admin`, we then manually decrement the limit.

That decrementing should ONLY be handled in the `Accept` method. In the `TransferCoin` endpoint, where we call `Accept`, we then need to properly handle its result, i.e. deleting if appropriate, or updating with the returned authorization if appropriate.

Basically, the `MarkerTransferAuthorization.Accept` method should always return the updated authorization and delete flag (without the `from`/`admin` comparison). Then the `authzHandler` should not care what type of authorization is being used as long as `Accept` says it's okay. It should also use the result of `Accept` to decide when to update or delete the authorization entry.

dwedul-figure:
I guess there **_is_** still actually an error.

[...]

PR #906, opened 2022-06-29
fix marker module double decrement bug

egaxhaj:
## Description

- Fix double decrement bug in marker module
- Added `MarkerTransferAuthorization` tests
- Added marker module `authzHandler` tests

closes: #903

---

write a little note why.

### DIFF
--- a/x/marker/types/authz_test.go
+++ b/x/marker/types/authz_test.go
@@ -1,0 +1,47 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+
+	"github.com/cosmos/cosmos-sdk/simapp"
+)
+
+var (
+	coin1000 = sdk.NewCoin("stake", sdk.NewInt(1000))
+	coin500  = sdk.NewCoin("stake", sdk.NewInt(500))
+)
+
+func TestMarkerTransferAuthorization(t *testing.T) {
+	app := simapp.Setup(false)
+	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
+	authorization := NewMarkerTransferAuthorization(sdk.NewCoins(coin1000))
+
+	t.Log("verify authorization returns valid method name")
+	require.Equal(t, authorization.MsgTypeURL(), "/provenance.marker.v1.MsgTransferRequest")
+	require.NoError(t, authorization.ValidateBasic())
+	send := &MsgTransferRequest{Amount: coin500}
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Bug Fixes
 
 * Add new `msgfees` `NhashPerUsdMil`  default param to param space store on upgrade (PR [#875](https://github.com/provenance-io/provenance/issues/875))
+* Marker module `authzHandler` function is double decrementing [#903](https://github.com/provenance-io/provenance/issues/903)
```
dwedul-figure, line 43: This is probably a bit too deep for the changelog. Also, it's not actually double decrementing due some fortuitous  interplay involving when the `authzHandler` is called.

A better changlog entry might be along the lines of "Fix the `MarkerTransferAuthorization` Accept function and `TransferCoin` authz handling to prevent problems when other authorization types are used."

I did verify with a user of that authorization that it's not actually double decrementing. But if that function were to be used in other places, it'd probably end up double decrementing.

x/marker/types/authz_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/cosmos/cosmos-sdk/simapp"
+)
+
+var (
+	coin1000 = sdk.NewCoin("stake", sdk.NewInt(1000))
+	coin500  = sdk.NewCoin("stake", sdk.NewInt(500))
+)
+
+func TestMarkerTransferAuthorization(t *testing.T) {
```
dwedul-figure, line 18: Nit: You don't need to change anything in here. This comment is mostly for future stuff.

I prefer to have tests split out by function, and then use `t.Run(...)` to identify specific test cases on that function.
1. It makes it easier to find where a function is being unit tested.
2. When something fails, it's easier to identify what the failure is.
3. If multiple parts break, it tells you all of them rather than just the first failure. That makes fixing things easier since A) It's easier to identify the root problem. B) Even when it only takes 10 seconds to run the unit tests, it's still an interruption of work flow. C) When you don't know everything that's broken, it's impossible to know when you might be done fixing things.

CHANGELOG.md
dwedul-figure: Also, a `State Machine Breaking` entry should be added to the changelog for this since handling of authorizations other than `MarkerTransferAuthorization` is changing (getting fixed).

egaxhaj: Fixed.

x/marker/types/authz_test.go
egaxhaj: Updated to use functions.

dwedul-figure (dismissed):

iramiller (approved):

PR #911, opened 2022-06-29
Update metadata module to properly manage authz grants.

egaxhaj:
## Description

- Calls `Accept()` to run authorization validation logic 
- handles deleting of grant when `Delete = true` in the response from `Accept()` 
- updates grant when `Updated != nil` in the response from `Accept()`

closes: #905 

---

write a little note why.


egaxhaj:
NOTE: awaiting a new release from upstream fork to be able to pull in https://github.com/provenance-io/cosmos-sdk/pull/175 and add some unit tests.

cc: @dwedul-figure

x/metadata/keeper/keeper_test.go
```diff
@@ -2,6 +2,7 @@ package keeper_test
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/require"
```
dwedul-figure, line 5: This should be moved down to 2nd to last (with the other testify entries)

x/metadata/keeper/keeper_test.go
```diff
@@ -312,6 +328,107 @@ func (s *KeeperTestSuite) TestValidateAllOwnerPartiesAreSigners() {
...
+
+	now := s.ctx.BlockHeader().Time
+	s.Require().NotNil(now)
+
+	// Test cases
+	for _, tc := range cases {
+		s.T().Run(tc.name, func(t *testing.T) {
+			a := authz.NewCountAuthorization(tc.msgTypeURL, tc.allowedAuthorizations)
+			err := s.app.AuthzKeeper.SaveGrant(s.ctx, tc.grantee, tc.granter, a, now.Add(time.Hour))
+			s.Require().NoError(err)
```
dwedul-figure, line 408: Need to skip this if the `granter` or `grantee` is nil. It's also best to include a message in the assertions so that, when it breaks, it's easier to identify. Something like `"SaveGrant"` is all that's needed.

x/metadata/keeper/keeper_test.go
```diff
@@ -312,6 +328,107 @@ func (s *KeeperTestSuite) TestValidateAllOwnerPartiesAreSigners() {
...
+		name                  string
+		owners                []types.Party
+		signers               []string
+		msgTypeURL            string
+		allowedAuthorizations int32
+		granter               sdk.AccAddress
+		grantee               sdk.AccAddress
+		errorMsg              string
+	}{
+		// generic authorization test cases
```
dwedul-figure, line 346:
```suggestion
		// count authorization test cases
```

x/metadata/keeper/keeper_test.go
```diff
@@ -388,69 +505,84 @@ func (s *KeeperTestSuite) TestValidateAllOwnersAreSigners() {
...
+		})
+	}
+}
+
+func (s *KeeperTestSuite) TestValidateAllOwnersAreSignersWithGenericAuthorization() {
+
+	tests := map[string]struct {
+		owners     []string
+		signers    []string
+		msgTypeURL string
```
dwedul-figure, line 527: Here too. Can we put the `TestValidateAllOwnersAreSigners` tests back to the way the were before?

x/metadata/keeper/keeper_test.go
```diff
@@ -460,6 +592,106 @@ func (s *KeeperTestSuite) TestValidateAllOwnersAreSigners() {
...
+
+	now := s.ctx.BlockHeader().Time
+	s.Require().NotNil(now)
+
+	// Test cases
+	for _, tc := range tests {
+		s.T().Run(tc.name, func(t *testing.T) {
+			a := authz.NewCountAuthorization(tc.msgTypeURL, tc.allowedAuthorizations)
+			err := s.app.AuthzKeeper.SaveGrant(s.ctx, tc.grantee, tc.granter, a, now.Add(time.Hour))
+			require.NoError(t, err)
```
dwedul-figure, line 671: Here too with the only-if `granter` and `grantee` aren't nil thing and require message thing.

x/metadata/keeper/keeper_test.go
```diff
@@ -460,6 +592,106 @@ func (s *KeeperTestSuite) TestValidateAllOwnersAreSigners() {
...
+					assert.Nil(t, auth)
+				} else {
+					actual := auth.(*authz.CountAuthorization).AllowedAuthorizations
+					assert.Equal(t, tc.allowedAuthorizations-1, actual)
+				}
+			}
+		})
+	}
+}
+
```
dwedul-figure, line 694: I'd like to see a test with two owners (1 & 2), and two different signers (3 & 4), with two authz count authorization grants: granter: 1, grantee: 3, count: 1; and granter: 2, grantee: 4, count: 2. Then a message is executed and we check that both grants have updated appropriately. Then another message is executed and we make sure it fails (because the first grant only had 1 use), and that the second grant still exists with 1 use left.

x/metadata/keeper/keeper_test.go
```diff
@@ -226,40 +227,73 @@ func (s *KeeperTestSuite) TestValidateAllOwnerPartiesAreSigners() {
...
+			if len(tc.errorMsg) == 0 {
+				assert.NoError(t, err, "%s unexpected error", n)
+			} else {
+				assert.EqualError(t, err, tc.errorMsg, "%s error", n)
+			}
+		})
+	}
+}
+
+func (s *KeeperTestSuite) TestValidateAllOwnerPartiesAreSignersWithGenericAuthorization() {
```
dwedul-figure, line 245: Can we put `TestValidateAllOwnerPartiesAreSigners` back the way it was before? That makes it easier to see that all the previous tests still pass, and that dual breaking changes (one in the main code, one in a test) weren't accidentally introduced.

CHANGELOG.md
```diff
@@ -42,6 +42,12 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * Update the swagger files (including third-party changes). [#728](https://github.com/provenance-io/provenance/issues/728)
 * Bump IBC to 2.3.0 and update third-party protos [PR 868](https://github.com/provenance-io/provenance/pull/868)
 
+### Features
+
+* Create authz authorization with limited number of uses. [#807](https://github.com/provenance-io/provenance/issues/807)
+  * Create a new authz authorization type that is limited to a certain number of times [#904](https://github.com/provenance-io/provenance/issues/904)
+  * Update metadata and marker modules to delete authorizations based on Delete flag [#728](https://github.com/provenance-io/provenance/issues/728)
```
dwedul-figure, line 49: The first two of these entries are the same, just with different issue numbers. You can get rid of one of them. Also, it'd be better and more accurate to phrase it with reference to the bumped cosmos-sdk version since that's where that new authz changes were actually made. Furthermore, identifying the new authorization type in here would help readers since the first question they're going to ask is, "What's this new type of authorization?"

```suggestion
* Bump Cosmos-SDK to `v0.45.4-pio-4` (from `v0.45.4-pio-2`) to utilize the new `CountAuthorization` authz grant type. [#807](https://github.com/provenance-io/provenance/issues/807)
* Update metadata module authz handling to properly call `Accept` and delete/update authorizations as they're used [#728](https://github.com/provenance-io/provenance/issues/728)
```

dwedul-figure (commented):
Since `v0.45.4-pio-4` has some new protos, you'll need to update the third-party protos in here too.

x/metadata/keeper/keeper_test.go
egaxhaj: 👍

x/metadata/keeper/keeper_test.go
egaxhaj: Hmm,  just looked at `SaveGrant` and it doesn't check for `nil` `granter` and `grantee`. It doesn't error out either. The goal must be to collect `gas` fees?

CHANGELOG.md
```diff
@@ -42,6 +42,11 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * Update the swagger files (including third-party changes). [#728](https://github.com/provenance-io/provenance/issues/728)
 * Bump IBC to 2.3.0 and update third-party protos [PR 868](https://github.com/provenance-io/provenance/pull/868)
 
+### Features
+
+* Bump Cosmos-SDK to `v0.45.4-pio-4` (from `v0.45.4-pio-2`) to utilize the new `CountAuthorization` authz grant type. [#807](https://github.com/provenance-io/provenance/issues/807)
+* Update metadata module authz handling to properly call `Accept` and delete/update authorizations as they're used [#728](https://github.com/provenance-io/provenance/issues/728)
```
iramiller, line 48: Issue 728 is for swagger

iramiller (changes requested):
This pull request (and the issue it is linked with) are titled as if they resolve this authz update for both metadata and marker modules.  Seemingly issue 905 should be for only the marker module update (there is no reason for the two modules to be updated together) and the marker update should be split into a new issue.

CHANGELOG.md
egaxhaj: I updated the title to reflect the change only for `metadata` module. 

No change required for the `marker` module. It already handles `accept.Delete`.
https://github.com/provenance-io/provenance/blob/a2aebd5971e06fbbb796dd2b4af1306524d8ec3a/x/marker/keeper/marker.go#L699-L717

SpicyLemon:
> This pull request (and the issue it is linked with) are titled as if they resolve this authz update for both metadata and marker modules. Seemingly issue 905 should be for only the marker module update (there is no reason for the two modules to be updated together) and the marker update should be split into a new issue.

The marker module update has already been done with #906.

This PR also bumps the cosmos-SDK version, which should probably be either a) split out into its own PR or b) listed in the description.

SpicyLemon (commented):
The tests still need some attention.

x/metadata/keeper/keeper_test.go
SpicyLemon: I mean put them back to what they were previously (i.e. no changes at all to those two functions).

x/metadata/keeper/keeper_test.go
```diff
@@ -312,12 +316,203 @@ func (s *KeeperTestSuite) TestValidateAllOwnerPartiesAreSigners() {
...
+				},
+				{
+					allowedAuthorizations: 2,
+					granter:               s.user2Addr,
+					grantee:               s.user4Addr,
+					errorMsg:              fmt.Sprintf("missing signatures from [%s (PARTY_TYPE_OWNER) ]", s.user1),
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
egaxhaj: > Also, I'm not sure if this is covered by any of the other tests, but it'd probably be good to have a test where two signers are required (1 & 2), and there are two grants (1 -> 3 for one use and 2 -> 3 for two uses), and only 3 is the signer; make sure both grants are updated/deleted appropriately.

This is very similar to the ^^^ case that we now want to move to `x/metadata/client/cli/cli_test.go`. Instead of:

```
```

x/metadata/client/cli/cli_test.go
```diff
@@ -3399,3 +3411,114 @@ func (s *IntegrationCLITestSuite) TestWriteSessionCmd() {
...
 }
+
+// ---------- tx cmd CountAuthorization tests ----------
+
+func (s *IntegrationCLITestSuite) TestCountAuthorizationIntactTxCommands() {
+	// The scenario being tested is as follows:
+	// There are two owners (1 & 2) and two signers (3 & 4).
+	// 1 & 2 are required signers and 3 & 4 are the actual signers.
+	// It should find a grant (granter: 1 -> grantee: 3) and complain
+	// about 4 not being one of the required signers because grant (granter: 2 -> grantee: 4) does not exist.
```
SpicyLemon, line 3422:
```suggestion
	// There are two owners (1 & 2) and the value owner is also 1.
	// That means that to delete the scope, both 1 and 2 need to be signers.
	// However, we're going to have user 3 sign the `RemoveScopeCmd` calls.
	// We first create the spec, then scope.
	// Then create a `CountAuthorization` grant for granter: 1 -> grantee: 3 with only one use.
	// We attempt a `RemoveScopeCmd` with 3 as the only signer. It should fail because it's not signed by 2.
	// Then we add another `CountAuthorization` grant for granter: 2 -> grantee: 3.
	// We attempt a `RemoveScopeCmd` again with only 3 as the signer. This time it should work.
	//
	// This tests the following:
	// * Makes sure the 1st `CountAuthorization` is not "used" for the failed `TX`. If it were, the 2nd `RemoveScopeCmd` would fail because it's not signed by 1.
	// * Makes sure that authorizations are only used once for the `TX` even if being used for both an owner and the value owner. If that were the case, the 2nd `RemoveScopeCmd` would fail because it wasn't signed by the value owner.
```

iramiller (approved):

dwedul-figure (approved):

SpicyLemon (approved):

Issue #627, opened 2022-01-28
Create an Active Participation and Engagement incentive module


iramiller:
## Summary

Create a new blockchain module that performs epoch based rewards distribution using a configurable share from the community pool to an eligible pool of addresses based on their on-chain activities during the period.  

## Problem Definition

One of the major challenges for a growing blockchain is the distribution of incentives to the right collection of accounts to drive interest and engagement in the platform.  While a proof of stake blockchain typically has a comprehensive distribution system based on delegations, this rewards system primarily benefits passive participation.  The Incentive module design targets active network participants for reward distribution by monitoring their activities and allocating distribution shares accordingly.

## Proposal

Create an Incentive module that supports distribution of rewards per epoch using shares.  The module should be built in a way that is extensible and flexible for different distribution strategies.  

Any account can create a new incentive program by providing funding and a set of configuration that includes the epochs to fund over, eligibility criteria, and a list of activity detectors and their weights.

In addition to direct funding through creation of incentive programs a community funding program can be configured via governance control which uses the community funding pool as a source of funds to distribute.

### End Block Processing
The processing pipeline of the module end block should be based on 4 stages:

1. Determine the incentive opportunity detectors to use for the current epoch
2. Identify events within the event manager history (#626) that represent incentive opportunities for an address, include a distribution weight factor and a source pool
3. For each incentive opportunity, filter based on eligibility criteria required for the incentive.
4. Update the epoch shares distribution for each address

At the end of a given epoch the module should determine the total funds to distribute from each funding source and split these amounts across the total number of shares allocated in the period (`share-value`).  Using this `share-value` the module should send a payment to each address in the amount of `( shares-accrued * share-value )`.  When this is complete the epoch records should be cleared in preparation for the next period.

### Basic Parameters (Governance Control)

- *Epoch duration (blocks)* - This is the length of the repeating epoch period measured in blocks
- *Community pool accrual tax rate* - This is the share of the accrued community funds to collected during the epoch to distribution (as a percentage)

### Incentives
Governance proposals should be used to configure the incentive opportunities available for a given epoch.  Incentive opportunities have a defined period, eligibility criteria, and funding source.

### Eligibility criteria examples

  - Account has transactions during period
  - Account has active delegations (threshold?)
  - Account maintains a balance of funds

## Example 1.

[...]

PR #930, opened 2022-07-12
Active participation and engagement module (#627)

arnabmitra:
## Description

closes: #627 
---
the complete spec lives here https://github.com/provenance-io/provenance/tree/8827dc3e8d0bf09ad793a038fade5dbb22bbf622/x/reward/spec maybe worth reading it to get an idea on the workings of the module higher level.

write a little note why.


proto/provenance/reward/v1/reward.proto
```diff
@@ -0,0 +1,254 @@
...
+  string maximum_active_stake_percentile = 6
+      [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
+}
+
+// ActionTransfer represents the transfer action and its required eligibility criteria.
+message ActionTransfer {
+  option (gogoproto.equal)            = true;
+  option (gogoproto.goproto_stringer) = false;
+
+  // Minimum number of successful delegates.
```
iramiller, line 233: This comment needs to be updated for the transfer action, similar for the next several occurrences of copy/paste

proto/provenance/reward/v1/tx.proto
```diff
@@ -0,0 +1,142 @@
...
+    option (google.api.http).patch = "/provenance/reward/v1/reward_programs/{reward_program_id}";
+  }
+
+  // ClaimRewards is the RPC endpoint for claiming rewards belonging to completed claim periods of a reward program
+  rpc ClaimRewards(MsgClaimRewardsRequest) returns (MsgClaimRewardsResponse) {
+    option (google.api.http).patch =
+        "/provenance/reward/v1/reward_claims/{reward_address}/reward_programs/{reward_program_id}";
+  }
+
+  // ClaimRewards is the RPC endpoint for claiming rewards for completed claim periods of every reward program for the
```
iramiller, line 33: The method name in this comment is incorrect

x/metadata/client/cli/cli_test.go
```diff
@@ -3411,114 +3399,3 @@ func (s *IntegrationCLITestSuite) TestWriteSessionCmd() {
 
 	runTxCmdTestCases(s, testCases)
 }
-
-// ---------- tx cmd CountAuthorization tests ----------
-
-func (s *IntegrationCLITestSuite) TestCountAuthorizationIntactTxCommands() {
```
iramiller, line 3417: Why was this test removed?

x/metadata/keeper/keeper.go
```diff
@@ -209,24 +204,10 @@ func (k Keeper) checkAuthzForMissing(
 			grantee := types.MustAccAddressFromBech32(signer)
 
 			for _, msgType := range msgTypeURLs {
-				authorization, exp := k.authzKeeper.GetCleanAuthorization(ctx, grantee, granter, msgType)
```
iramiller, line 212: These changes do not look like they should be reverted here.

iramiller (commented):
Some initial comments ... it looks like some unintentional changes to the metadata module were included (reverting upstream changes?).  Also some minor clean up in the protobuf comments is required.  Overall this looks like a very high quality release.

Due to its size it will take a while to review... this is just a first pass for me.

x/metadata/keeper/keeper.go
arnabmitra: yeah looks we have a bad merge from main :( .. reverted all metadata files to match main

proto/provenance/reward/v1/reward.proto
arnabmitra: good catch, fixed to reflect transfer action.

proto/provenance/reward/v1/tx.proto
arnabmitra: good catch, updated comment to reflect ClaimAllRewards

proto/provenance/reward/v1/reward.proto
```diff
@@ -0,0 +1,254 @@
...
+  // Grace period after a RewardProgram FINISHED. It is the number of seconds until a RewardProgram enters the EXPIRED
+  // state.
+  uint64 expiration_offset = 20;
+  // Actions that count towards the reward.
+  repeated QualifyingAction qualifying_actions = 21
+      [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"qualifying_actions"];
+}
+
+// EventSetRewardProgram event emitted when a reward program is is updated
+message EventSetRewardProgram {
```
iramiller, line 102: This is an extensive amount of information in an event... typically speaking events are small and indicate something specific occurred. It would make more sense to indicate the program changed and provide its ID for an interested user to query on than to duplicate the entire state of the object into the event...

iramiller (commented):
Looking pretty solid overall. Just a couple minor questions/comments.

proto/provenance/reward/v1/reward.proto
arnabmitra: @Taztingo  i believe deleted it, since it seems a bit too much detail.

x/reward/client/cli/tx.go
```diff
@@ -0,0 +1,227 @@
...
+		GetCmdRewardProgramAdd(),
+		GetCmdEndRewardProgram(),
+		GetCmdClaimReward(),
+	)
+
+	return txCmd
+}
+
+func GetCmdRewardProgramAdd() *cobra.Command {
+	actionsExampleJSON := "{\"qualifying_actions\":[{\"delegate\":{\"minimum_actions\":\"0\",\"maximum_actions\":\"0\",\"minimum_delegation_amount\":{\"denom\":\"nhash\",\"amount\":\"0\"},\"maximum_delegation_amount\":{\"denom\":\"nhash\",\"amount\":\"100\"},\"minimum_active_stake_percentile\":\"0.000000000000000000\",\"maximum_active_stake_percentile\":\"1.000000000000000000\"}}]}"
```
SpicyLemon, line 52: If you delimited this with ticks, you wouldn't have to escape all the quotes.

Alternatively, since it's only used in one place, you could just put it there instead of `%s`ing it in.

x/reward/client/cli/tx.go
```diff
@@ -0,0 +1,227 @@
...
+		"Program Title" "A short description" \
+		--total-reward-pool 580nhash \
+		--max-reward-by-address 10nhash \
+    	--start-time 2022-05-10\
+		--claim-periods 52 \
+		--max-rollover-periods 4 \
+		--claim-period-days 7 \
+		--expire-days 14 \ 
+		--qualifying-actions %s
+		`, version.AppName, actionsExampleJSON),
```
SpicyLemon, line 69: This is all inside single tics, so all that indenting is included.  That's probably also why the `--start-time` entry isn't spaced in the same as the rest.

Since it's all part of a single, multi-line command, having it indented one level would be prudent, but not as far as it is right now.

It's output like this:
```
$ provenanced tx reward add-reward-program \
		"Program Title" "A short description" \
		--total-reward-pool 580nhash \
```

x/reward/client/cli/tx.go
```diff
@@ -0,0 +1,227 @@
...
+		Example: fmt.Sprintf(`$ %[1]s tx reward add-reward-program \
+		"Program Title" "A short description" \
+		--total-reward-pool 580nhash \
+		--max-reward-by-address 10nhash \
+    	--start-time 2022-05-10\
+		--claim-periods 52 \
+		--max-rollover-periods 4 \
+		--claim-period-days 7 \
+		--expire-days 14 \ 
+		--qualifying-actions %s
```
SpicyLemon, line 68: Mixing numbered and non-numbered format identifiers can easily lead to problems.
```suggestion
		--qualifying-actions %[2]s
```

internal/handlers/msg_service_router_test.go
```diff
@@ -406,7 +414,1712 @@ func TestMsgServiceAssessMsgFee(t *testing.T) {
...
+	app.Commit()
+
+	assert.Equal(t, 13, len(res.Events))
+	lookForRewardCreatedEvent, found := contains(res.Events, "reward_program_created")
+	assert.Equal(t, true, found, "event reward_program_created should exist")
+	assert.Equal(t, 1, len(*lookForRewardCreatedEvent), "event reward_program_created should exist")
+	lookForMessageEvent, foundMessage := contains(res.Events, "message")
+	assert.Equal(t, true, foundMessage, "event message should exist")
+	_, foundAttribute := containsAttribute(*lookForMessageEvent, "create_reward_program")
+	assert.Equal(t, true, foundAttribute, "event attribute create_reward_program should exist")
```
SpicyLemon, line 508: Here again with the `contains` thing vs using the `.Contains` test. If one of these were to fail, the only context you would get is the msg and that something was true/false unexpectedly. You'd then have to spend extra time firing up the debugger in order to see what it was and what was missing.

Alternatively, create a function that takes in a `*testing.T` and is a `t.Helper()` that does the checking you want to be done, but make it output all of the thing being checked and what it's being checked for.

internal/handlers/msg_service_router_test.go
```diff
@@ -406,7 +414,1712 @@ func TestMsgServiceAssessMsgFee(t *testing.T) {
...
+		_, res, errFromDeliverTx := app.Deliver(encCfg.TxConfig.TxEncoder(), tx1)
+		require.NoError(t, errFromDeliverTx)
+		assert.Equal(t, true, len(res.GetEvents()) >= 1, "should have emitted an event.")
+		app.EndBlock(abci.RequestEndBlock{Height: height})
+		app.Commit()
+		seq = seq + 1
+	}
+	claimPeriodDistributions, err := app.RewardKeeper.GetAllClaimPeriodRewardDistributions(ctx)
+	require.NoError(t, err)
+	assert.Equal(t, true, len(claimPeriodDistributions) == 1, "claim period reward distributions should exist")
```
SpicyLemon, line 574: There are many better tests than this.
```suggestion
	assert.Len(t, claimPeriodDistributions, 1, "claim period reward distributions should exist")
```

I'm going to stop commenting on all these spots. But please stop using the `.Equal` test with `true`/`false` period, and only use `.True` and `.False` tests when you're specifically testing a function that returns a boolean. You're just shooting yourself (and me) in the foot.

x/reward/keeper/claim_reward_distribution.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return rewardDistributions, nil
+}
+
+// ClaimPeriodRewardDistributionIsValid Checks if an Epoch Reward Distribution is valid
+func (k Keeper) ClaimPeriodRewardDistributionIsValid(claimPeriodReward *types.ClaimPeriodRewardDistribution) bool {
+	return claimPeriodReward.RewardProgramId != 0
+}
```
SpicyLemon, line 62: Is there a reason this isn't a method on the `ClaimPeriodRewardDistribution` type? Usually something like this would go in a `ValidateBasic()` method. I know it's not a `Msg`, but it's still pretty standard to do that.

x/reward/keeper/claim_reward_distribution_test.go
```diff
@@ -0,0 +1,198 @@
...
+				types.NewClaimPeriodRewardDistribution(1, 1, sdk.NewInt64Coin("nhash", 100), sdk.NewInt64Coin("nhash", 100), 1, false),
+				types.NewClaimPeriodRewardDistribution(1, 2, sdk.NewInt64Coin("nhash", 100), sdk.NewInt64Coin("nhash", 100), 1, false),
+				types.NewClaimPeriodRewardDistribution(1, 3, sdk.NewInt64Coin("nhash", 100), sdk.NewInt64Coin("nhash", 100), 1, false),
+			},
+			true,
+			1,
+		},
+	}
+
+	for _, tt := range tests {
```
SpicyLemon, line 79: Nit: Usually we use `tc` for this variable name.

SpicyLemon (commented):
I'm only about 1/3 of the way through, but wanted to submit what I've got because it's late and I'm tired.

Will pick it back up again tomorrow.

x/reward/keeper/claim_reward_distribution_test.go
Taztingo: Good catch, will change.

x/reward/keeper/claim_reward_distribution.go
Taztingo: This was primarily done to keep the ClaimPeriodRewardDistribution a pod, but it looks like we added other methods to it. Thanks for finding this, and I'll remove it to use the Validate method we have instead.

x/reward/client/cli/tx.go
Taztingo: Agreed, will change.

internal/handlers/msg_service_router_test.go
Taztingo: Agreed, we should be using the other methods that assert exposes. Went through and changed the ones that make use of length.

app/app.go
```diff
@@ -717,6 +720,8 @@ func New(
 		name.NewAppModule(appCodec, app.NameKeeper, app.AccountKeeper, app.BankKeeper),
 		attribute.NewAppModule(appCodec, app.AttributeKeeper, app.AccountKeeper, app.BankKeeper, app.NameKeeper),
 		msgfeesmodule.NewAppModule(appCodec, app.MsgFeesKeeper, app.interfaceRegistry),
+		wasm.NewAppModule(appCodec, &app.WasmKeeper, app.StakingKeeper),
```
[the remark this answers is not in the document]
arnabmitra: yeah idk either i will delete it... 🤷

cmd/provenanced/cmd/config_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	provenanceconfig "github.com/provenance-io/provenance/internal/pioconfig"
```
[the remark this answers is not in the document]
arnabmitra: good point..there is no need for that alias..removing it and fixed the import ordering.

x/reward/abci_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	app = simapp.Setup(false)
+	ctx = app.BaseApp.NewContext(false, tmproto.Header{})
+	ctx = ctx.WithBlockHeight(1).WithBlockTime(now)
+
+	ctx = ctx.WithBlockHeight(2)
+	require.NotPanics(t, func() {
+		reward.BeginBlocker(ctx, app.RewardKeeper)
+	})
+}
```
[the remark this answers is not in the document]
arnabmitra: i think the Begin blocker and end blocker is checked in keeper_test.go quite extensively..for e.g ```	suite.app.RewardKeeper.StartRewardProgram(suite.ctx, &rewardProgram)
	suite.app.RewardKeeper.SetRewardProgram(suite.ctx, rewardProgram)

	// We want to set the events here
	validators := getTestValidators(6, 6)
	delegates := suite.createDelegateEvents(validators[0].OperatorAddress, "1000000000nhash", "cosmos15ky9du8a2wlstz6fpx3p4mqpjyrm5cgxwpuzvh", "50000000000000.000000000000000000")
	delegates = delegates.AppendEvents(suite.createDelegateEvents(validators[0].OperatorAddress, "1000000000nhash", "cosmos15ky9du8a2wlstz6fpx3p4mqpjyrm5cgxwpuzvh", "50000000000000.000000000000000000"))
	SetupEventHistory(suite, delegates)
	reward.EndBlocker(suite.ctx, suite.app.RewardKeeper)```
Additionally the msg_servicerouter_test also do this because they exercise this via ```app.BeginBlock(abci.RequestBeginBlock{Header: tmproto.Header{Height: height, Time: time.Now().UTC()}})``` &&& ```app.EndBlock(abci.RequestEndBlock{Height: height})
		app.Commit()``` ..so i would defer to those tests imo..i would almost vote to delete `abci_test.go` because it give the impression that those were not tested more elaborately but 🤷‍♂️

proto/provenance/reward/v1/query.proto
```diff
@@ -0,0 +1,148 @@
...
+  }
+  // A filter on the types of reward programs.
+  QueryType query_type = 1;
+}
+
+// QueryRewardProgramsResponse contains the list of RewardPrograms matching the query
+message QueryRewardProgramsResponse {
+  // List of RewardProgram objects matching the query_type.
+  repeated RewardProgram reward_programs = 1 [(gogoproto.nullable) = false];
+}
```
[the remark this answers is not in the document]
nullpointer0x00: We should probably default to adding pagination to any query with variable results in the future.  I added them for this query.

internal/handlers/msg_service_router_test.go
Taztingo: Agreed, this can be simplified. Added two additional test helper functions.

cmd/provenanced/cmd/testnet.go
```diff
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"time"
 
+	provenanceconfig "github.com/provenance-io/provenance/internal/pioconfig"
+
```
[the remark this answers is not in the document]
arnabmitra: changed

internal/handlers/msg_service_router_test.go
```diff
@@ -2,10 +2,18 @@ package handlers_test
 
 import (
 	"fmt"
+	"github.com/provenance-io/provenance/internal/pioconfig"
```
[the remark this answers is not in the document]
arnabmitra: changed..moved it with other provenance imports.

x/reward/client/cli/cli_test.go
```diff
@@ -0,0 +1,1023 @@
...
+		expectedCode uint32
+	}{
+		{"claim rewards tx - valid",
+			[]string{
+				"1",
+			},
+			false,
+			"",
+			0,
+		},
```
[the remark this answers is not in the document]
nullpointer0x00: Yep, good catch.  We were doing a lot of copy and paste for our tests.  Passing in a []string or 1 argument doesn't make sense in these cases.  Also, I will keep an eye out for patterns like this in other modules (since they were a copy and paste).  Thanks, good catches.

x/reward/client/cli/cli_test.go
```diff
@@ -0,0 +1,1023 @@
...
+func (s *IntegrationTestSuite) TestQueryClaimPeriodRewardDistributionAll() {
+	defaultMaxQueryIds := make([]uint64, 100)
+	for i := 0; i < 100; i++ {
+		defaultMaxQueryIds[i] = uint64(i + 1)
+	}
+	testCases := []struct {
+		name         string
+		args         []string
+		byId         bool
+		expectErr    bool
```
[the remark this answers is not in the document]
nullpointer0x00: Yep good point.  I will keep my eye out for this pattern else where in our project.  Since, we copied and pasted the structure from some other module.

proto/provenance/reward/v1/tx.proto
```diff
@@ -0,0 +1,142 @@
...
+message MsgCreateRewardProgramRequest {
+  option (gogoproto.equal)            = true;
+  option (gogoproto.goproto_stringer) = false;
+
+  // title for the reward program.
+  string title = 1;
+  // description for the reward program.
+  string description = 2;
+  // provider address for the reward program funds and signer of message.
+  string distribute_from_address = 3;
```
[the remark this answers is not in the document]
Taztingo: The jsontag isn't needed and it serializes to snake_case by default.

proto/provenance/reward/v1/reward.proto
```diff
@@ -0,0 +1,219 @@
...
+  // Minimum amount of coins for a program to rollover.
+  cosmos.base.v1beta1.Coin minimum_rollover_amount = 9 [(gogoproto.nullable) = false];
+  // Number of seconds that a claim period lasts.
+  uint64 claim_period_seconds = 10;
+  // Time that a RewardProgram should start and switch to STARTED state.
+  google.protobuf.Timestamp program_start_time = 11 [
+    (gogoproto.stdtime)  = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.jsontag)  = "program_start_time",
+    (gogoproto.moretags) = "yaml:\"program_start_time\""
```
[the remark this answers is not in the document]
Taztingo: Removed yaml tags.

x/reward/keeper/time_management.go
```diff
@@ -0,0 +1,245 @@
...
+
+	claimPeriodReward, err := k.GetClaimPeriodRewardDistribution(ctx, rewardProgram.GetCurrentClaimPeriod(), rewardProgram.GetId())
+	if err != nil || claimPeriodReward.GetClaimPeriodId() == 0 {
+		ctx.Logger().Error(fmt.Sprintf("Missing ClaimPeriodRewardDistribution for RewardProgram %d ", rewardProgram.GetId()))
+		return fmt.Errorf("a ClaimPeriodRewardDistribution does not exist for RewardProgram %d with claim period %d", rewardProgram.GetId(), rewardProgram.GetId())
+	}
+
+	totalClaimPeriodRewards, err := k.CalculateRewardClaimPeriodRewards(ctx, rewardProgram.GetMaxRewardByAddress(), claimPeriodReward)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("Unable to calculate claim period rewards for RewardProgram %d ", rewardProgram.GetId()))
```
SpicyLemon, line 123:
```suggestion
		ctx.Logger().Error(fmt.Sprintf("Unable to calculate claim period rewards for RewardProgram %d ", rewardProgram.GetId()), "err", err)
```

x/reward/module/module.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// ConsensusVersion implements AppModule/ConsensusVersion.
+func (AppModule) ConsensusVersion() uint64 { return 1 }
+
+func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
+	rewardModule.BeginBlocker(ctx, am.keeper)
+}
+
+// EndBlock does nothing
```
SpicyLemon, line 189: Not true.

x/reward/simulation/decoder.go
```diff
@@ -0,0 +1,29 @@
...
+func NewDecodeStore(cdc codec.Codec) func(kvA, kvB kv.Pair) string {
+	return func(kvA, kvB kv.Pair) string {
+		switch {
+		case bytes.Equal(kvA.Key[:1], types.RewardProgramIDKey):
+			var attribA, attribB types.RewardProgram
+
+			cdc.MustUnmarshal(kvA.Value, &attribA)
+			cdc.MustUnmarshal(kvB.Value, &attribB)
+
+			return fmt.Sprintf("%v\n%v", attribA, attribB)
```
SpicyLemon, line 24: Don't we need a `case` for each type of key we put in the store?

Also, you can use `bytes.HasPrefix(kva.Key, types.RewardProgramIDKey)` to be a bit more robust (e.g. if `RewardProgramIDKey` ever changes to have more than one byte).

x/reward/spec/08_genesis.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Reward Genesis
+
+In this section we describe the processing of the reward messages and the corresponding updates to the state.
+
+<!-- TOC 2 -->
+  - [Msg/CreateRewardProgramRequest](#msg-createrewardprogramrequest)
+  - [Msg/EndRewardProgramRequest](#msg-endrewardprogramrequest)
+  - [Msg/ClaimRewardRequest](#msg-claimrewardrequest)
+  - [Msg/ClaimAllRewardsRequest](#msg-claimallrewardsrequest)
```
SpicyLemon, line 13: Outdated TOC.

x/reward/types/action_builder.go
```diff
@@ -0,0 +1,234 @@
+package types
+
+import (
+	"fmt"
+
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/cosmos/cosmos-sdk/x/gov/types"
```
SpicyLemon, line 10:
```suggestion
	sdk "github.com/cosmos/cosmos-sdk/types"
	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
	"github.com/cosmos/cosmos-sdk/x/gov/types"
	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
```

x/reward/types/action_builder.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
```
SpicyLemon, line 12:
```suggestion

var (
	_ ActionBuilder = TransferActionBuilder{}
	_ ActionBuilder = DelegateActionBuilder{}
	_ ActionBuilder = VoteActionBuilder{}
)

```

x/reward/types/reward.go
```diff
@@ -0,0 +1,583 @@
...
+type (
+	// RewardAction defines the interface that actions need to implement
+	RewardAction interface {
+		ActionType() string
+		Evaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState, event EvaluationResult) bool
+		PreEvaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState) bool
+		PostEvaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState) bool
+		GetBuilder() ActionBuilder
+	}
+)
```
SpicyLemon, line 48:
```suggestion
// RewardAction defines the interface that actions need to implement
type RewardAction interface {
	ActionType() string
	Evaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState, event EvaluationResult) bool
	PreEvaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState) bool
	PostEvaluate(ctx sdk.Context, provider KeeperProvider, state RewardAccountState) bool
	GetBuilder() ActionBuilder
}
```

SpicyLemon (commented):
Just finished looking through the rest.

x/reward/simulation/decoder.go
arnabmitra: i am not sure 
very specific to to sim tests for sure ``````// AppModuleSimulation defines the standard functions that every module should expose
// for the SDK blockchain simulator``````
`metadata` and `marker` both have it commented out ```// RegisterStoreDecoder registers a decoder for metadata module's types
func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
	// sdr[types.StoreKey] = simulation.NewDecodeStore(am.cdc)
}
```
 marker has a lot of sims tests, so not sure how it is used, but i added the rest of the keys(actually there were a couple of unused keys, so a helpful comment :) )

File: x/reward/keeper/reward_account_state.go [hunk 30]
```diff
@@ -0,0 +1,188 @@
...
+// ParseRewardAccountLookUpKey parse key generated like via GetRewardAccountStateAddressLookupKey(address, rewardProgramId, claimPeriodId)
+func ParseRewardAccountLookUpKey(accountStateAddressLookupKey []byte, addr sdk.AccAddress) (RewardAccountLookup, error) {
+	// address length is stored in first byte
+	lengthOfAddress := int64(accountStateAddressLookupKey[1:2][0])
+	address := sdk.AccAddress(accountStateAddressLookupKey[2 : lengthOfAddress+2])
+	if !addr.Equals(address) {
+		return RewardAccountLookup{}, fmt.Errorf("addresses do not match up")
+	}
+	rewardID := binary.BigEndian.Uint64(accountStateAddressLookupKey[lengthOfAddress+2 : lengthOfAddress+2+types.RewardIDKeyLength])
+	claimID := binary.BigEndian.Uint64(accountStateAddressLookupKey[lengthOfAddress+2+types.RewardIDKeyLength : lengthOfAddress+2+types.RewardIDKeyLength+types.ClaimPeriodIDLength])
```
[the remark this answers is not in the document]
arnabmitra: moved

x/reward/module/module.go
arnabmitra: fixed

x/reward/types/action_builder.go
Taztingo: Sure, I can add this.

x/reward/keeper/reward_claim_test.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+var (
+	minDelegation = sdk.NewInt64Coin("nhash", 4)
+	maxDelegation = sdk.NewInt64Coin("nhash", 40)
+)
+
+func (suite *KeeperTestSuite) TestClaimRewards() {
+	suite.SetupTest()
```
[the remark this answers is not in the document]
Taztingo: Removed.

x/reward/keeper/reward_claim_test.go
```diff
@@ -0,0 +1,321 @@
...
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+var (
+	minDelegation = sdk.NewInt64Coin("nhash", 4)
+	maxDelegation = sdk.NewInt64Coin("nhash", 40)
+)
+
+func (suite *KeeperTestSuite) TestClaimRewards() {
```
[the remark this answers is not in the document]
Taztingo: Works for me.

x/reward/keeper/rules.go
```diff
@@ -0,0 +1,241 @@
+package keeper
+
+import (
+	"fmt"
+
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
```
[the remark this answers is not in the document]
Taztingo: Makes sense.

x/reward/keeper/rules.go
```diff
@@ -0,0 +1,241 @@
...
+
+	return nil
+}
+
+// IterateABCIEvents Iterates through all the ABCIEvents that match the eventCriteria.
+// Nil criteria means to iterate over everything.
+func (k Keeper) IterateABCIEvents(ctx sdk.Context, criteria *types.EventCriteria, action func(string, *map[string][]byte) error) error {
+	for _, event := range ctx.EventManager().GetABCIEventHistory() {
+		event := event
+		ctx.Logger().Info(fmt.Sprintf("events type is %s", event.Type))
```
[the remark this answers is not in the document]
Taztingo: This was accidentally left it. Removed.

x/reward/keeper/rules.go
```diff
@@ -0,0 +1,241 @@
...
+		// Process actions and get the results
+		actions = k.ProcessQualifyingActions(ctx, program, action, actions)
+		results = append(results, actions...)
+	}
+
+	return results, nil
+}
+
+func (k Keeper) ProcessQualifyingActions(ctx sdk.Context, program *types.RewardProgram, processor types.RewardAction, actions []types.EvaluationResult) []types.EvaluationResult {
+	successfulActions := []types.EvaluationResult(nil)
```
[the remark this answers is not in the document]
Taztingo: This change would lead to an unallocated array. We currently have a slice, but we can allocate space upfront if we think slice appending is a bottleneck.

x/reward/keeper/rules.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	for _, action := range actions {
+		state, err := k.GetRewardAccountState(ctx, program.GetId(), program.GetCurrentClaimPeriod(), action.Address.String())
+		if err != nil {
+			continue
+		}
+		if state.Validate() != nil {
+			state = types.NewRewardAccountState(program.GetId(), program.GetCurrentClaimPeriod(), action.Address.String(), 0, map[string]uint64{})
+		}
```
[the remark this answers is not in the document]
Taztingo: We are creating a new state because of the pattern we have in place. When a key is not found we are returning a null/empty object. The id 0 is not considered valid so it fails validation, and we create the missing object. We are creating state on the fly because we won't know upfront how many people will be participating.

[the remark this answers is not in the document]
Taztingo: In a nutshell it's valid, but we just haven't created the state yet for that user.

x/reward/keeper/reward_claim.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if len(rewards) == 0 {
+		return amount, nil
+	}
+
+	for i := 0; i < len(rewards); i++ {
+		reward := rewards[i]
+		amount.Denom = reward.GetClaimPeriodReward().Denom
+		amount = amount.Add(reward.GetClaimPeriodReward())
+	}
```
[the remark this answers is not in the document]
Taztingo: Nope, we can use for each here. Updated to use for each.

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	enforceMaxMinPageLimit(pageRequest)
+	return pageRequest
+}
+
+func enforceMaxMinPageLimit(pageRequest *query.PageRequest) {
+	if pageRequest.Limit == 0 || pageRequest.Limit > defaultPerPageLimit {
+		pageRequest.Limit = defaultPerPageLimit
+	}
+}
```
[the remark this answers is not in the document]
nullpointer0x00: We modified an example of the metadata modules pagination pattern here.  Yes, it was giving us a few problems.  I just stripped everything out and followed how the sdk does things instead of adding this custom query methods.  The sdk will set the default max limit to `100` if it isn't set.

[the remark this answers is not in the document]
SpicyLemon: But this sets a **MAX** of 100 on the limit, which isn't very user friendly. E.g. If I do a query with `--limit 500`, this will then only return 100 entries (even if there are more than that).

With the SDK's logic, providing `Limit = 0` (or not providing a limit argument) is the same as providing `Limit = 100` and `CountTotal = true`. By defaulting it to `100` here (before it gets to the SDK logic), you lose the `CountTotal = true` behavior. If that's what's intended here, there should be a comment about it with a bit about why.

But even if the change to `CountTotal` is intended, there shouldn't be a max on the limit.

x/reward/keeper/rules.go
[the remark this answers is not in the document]
SpicyLemon: `successfulActions := []types.EvaluationResult(nil)` casts `nil` into a `[]types.EvEvaluationResult`.
`var successfulActions []types.EvaluationResult` creates a variable of type `[]types.EvaluationResult` with a value of `nil`.

They're almost equivalent. The former requires a cast operation. It's also syntax that is rarely used for defining a slice variable.

x/reward/keeper/query_server.go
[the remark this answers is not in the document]
nullpointer0x00: Sorry, I wasn't clear.  I saw that as a problem as well.  I removed all that code and just let the sdk handle the pagination.  We are no longer calling the problematic method (`enforceMaxMinPageLimit`)  that set the **MAX** as 100.  Also, the sdk will set the limit to 100 if the `Limit = 0`.  Therefore, checking if the limit is set to 0 ourself and setting it to 100 is redundant.

x/reward/keeper/msg_server.go
```diff
@@ -0,0 +1,146 @@
...
+		sdk.NewEvent(
+			types.EventTypeRewardProgramEnded,
+			sdk.NewAttribute(types.AttributeKeyRewardProgramID, fmt.Sprintf("%d", rewardProgram.Id)),
+		),
+	)
+
+	return &types.MsgEndRewardProgramResponse{}, nil
+}
+
+func (s msgServer) ClaimRewards(goCtx context.Context, req *types.MsgClaimRewardsRequest) (*types.MsgClaimRewardsResponse, error) {
```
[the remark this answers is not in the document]
arnabmitra: updated comments

x/reward/spec/02_state.md
```diff
@@ -0,0 +1,67 @@
...
+
+<!-- TOC -->
+- [State](#state)
+  - [Reward Program](#reward-program)
+  - [Claim Period Reward Distribution](#claim-period-reward-distribution)
+  - [Reward Account State](#reward-account-state)
+  - [Qualifying Actions](#qualifying-actions)
+    - [Action Delegate](#action-delegate)
+    - [Action Transfer](#action-transfer)
+    - [Action Vote](#action-vote)
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/time_management.go
```diff
@@ -0,0 +1,245 @@
...
+		// called from the beginblocker, not much we can do here but return
+		return
+	}
+
+	for index := range rewardPrograms {
+		switch {
+		case rewardPrograms[index].IsStarting(ctx):
+			err = k.StartRewardProgram(ctx, &rewardPrograms[index])
+			if err != nil {
+				return
```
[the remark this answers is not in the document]
arnabmitra: agreed ..`On iterating through reward programs, logging error and continuing rather than returning error,`

Taztingo (approved):
LGTM

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,203 @@
...
+	return &response, nil
+}
+
+func (k Keeper) RewardDistributionsByAddress(ctx context.Context, request *types.QueryRewardDistributionsByAddressRequest) (*types.QueryRewardDistributionsByAddressResponse, error) {
+	if request == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	address, err := sdk.AccAddressFromBech32(request.Address)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, err.Error())
```
SpicyLemon, line 140: The `sdkerrors.Wrap` and `Wrapf` functions are deprecated. When you've got a specific error (like here), you should use the `.Wrap` or `.Wrapf` off that error.
```suggestion
		return nil, sdkerrors.ErrInvalidAddress.Wrap(err.Error())
```
There are many other places where it's used, but I'll just leave this one comment.

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,203 @@
...
+			// think ignoring the error maybe ok here since it's just another lookup
+			if errFromGetRewardAccount != nil {
+				return false, errFromGetRewardAccount
+			}
+			if result.GetSharesEarned() == 0 || (request.ClaimStatus != result.ClaimStatus && request.ClaimStatus != types.RewardAccountState_CLAIM_STATUS_UNSPECIFIED) {
+				return false, nil
+			}
+
+			states = append(states, result)
+		}
```
SpicyLemon, line 162: Did this get changed recently? I could have sworn it used to only have the `states = append(states, result)` inside the `if accumulate {` block.

If there's an entry where `errFromGetRewardAccount` is nil or the other `if` stuff is true, then the counting will be off when using an offset for pagination. This will specially be noticeable if filtering the results by claim status. I know it sucks that all the extra work needs to be done when not accumulating, but in this case, there's no way around it; it all needs to be done before we know whether or not it should count as an entry.

x/reward/keeper/msg_server_test.go
```diff
@@ -0,0 +1,563 @@
...
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/simapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+func (s *KeeperTestSuite) TestCreateRewardProgramTransaction() {
+	s.SetupTest()
```
SpicyLemon, line 13: I see this `s.SetupTest()` call in a lot of tests. That's being called by the `suite.Suite` stuff automatically. So calling it explicitly is calling it a second time.

x/reward/keeper/time_management.go
```diff
@@ -0,0 +1,256 @@
...
+// UpdateUnexpiredRewardsProgram called from begin blocker, starts/ends or expires rewards programs.
+func (k Keeper) UpdateUnexpiredRewardsProgram(ctx sdk.Context) {
+	rewardPrograms, err := k.GetUnexpiredRewardPrograms(ctx)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("error iterating reward programs: %v ", err))
+		// called from the beginblocker, not much we can do here but return
+		return
+	}
+
+	for index := range rewardPrograms {
```
SpicyLemon, line 21: Iterating over maps = :badtime:

x/reward/keeper/query_server.go
arnabmitra: >  I could have sworn it used to only have the states = append(states, result) inside the if accumulate { block.

still seeing it happen within the loop, unless i am misunderstanding your comment.

x/reward/keeper/time_management.go
```diff
@@ -0,0 +1,256 @@
...
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+// UpdateUnexpiredRewardsProgram called from begin blocker, starts/ends or expires rewards programs.
+func (k Keeper) UpdateUnexpiredRewardsProgram(ctx sdk.Context) {
+	rewardPrograms, err := k.GetUnexpiredRewardPrograms(ctx)
```
SpicyLemon, line 14: I'm a little worried that we have to iterate over all reward programs in every begin blocker. It's probably not going to be an issue until there are lots of rewards programs, but I don't know what "lots" counts as.

Maybe as a separate issue, we should create an index/queue similar to what was done with the `x/authz` module.

x/reward/spec/02_state.md
```diff
@@ -0,0 +1,75 @@
...
+
++++ https://github.com/provenance-io/provenance/blob/4e354a9fd554a420f7970522d2e8b0b749baad9d/proto/provenance/reward/v1/reward.proto#L114-L130
+
+---
+## Reward Account State
+
+The purpose of `RewardAccountState` is to track state at the address level of a claim period. It counts the number of claim period shares the user obtained, the status of their `RewardClaim`, and other stateful information that assists the system in properly granting rewards.
+
+* AccountStateAddressLookupKeyPrefix: `0x04 | Account Address (8 bytes) | Reward Program ID (8 bytes) | Claim Period ID (8 bytes) -> ProtocolBuffers(RewardAccountState)`
+* AccountStateKeyPrefix: `0x05 | Reward Program ID (8 bytes) | Claim Period ID (8 bytes) | Account Address (8 bytes) -> ProtocolBuffers(RewardAccountState)`
```
SpicyLemon, line 42: Is the `Account Address (8 bytes)` part right? Usually there's a length byte followed by the address.

x/reward/keeper/query_server.go
arnabmitra: changed

x/reward/keeper/msg_server_test.go
arnabmitra: agreed

x/reward/keeper/query_server.go
nullpointer0x00: I moved the accumulate method back.  I understand it better now.  Thanks.

x/reward/spec/02_state.md
arnabmitra: agreed ..i'll fix the spec.. code wise looks like ^^ comment ```	
lengthOfAddress := int64(accountStateAddressLookupKey[1:2][0])
	address := sdk.AccAddress(accountStateAddressLookupKey[2 : lengthOfAddress+2])
```

File: x/reward/keeper/query_server.go [hunk 48]
```diff
@@ -0,0 +1,202 @@
...
+					if rewardProgram.GetState() == state {
+						response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
+						break
+					}
+				}
+			default:
+				return false, vErr
+			}
+		}
+		return true, nil
```
SpicyLemon, line 61: This still has the problem of not properly counting things (same with the `FilteredPaginate` in `ClaimPeriodRewardDistributions`). It needs to return `true` or `false` the same, regardless of what `accumulate` is.

Lets say there are 100 entries, and we're doing limit 2. But we're filtering for `state` = `expired`. There are two expired entries, at indexes 1, 94, and 95.

Offset 0:
* `accumulate` will start out as `true`.
* Index 0 returns `false`, so it  isn't counted.
* Index 1 returns `true`, and is counted/added.
* Indexes 2-93 return `false`.
* Index 94, returns `true` and is counted/added.
* There have been two `true`s, so the limit has been reached, and `accumulate` is now `false`.
* If still counting, indexes 95, 96, 97, 98, and 99 all return true
* It has returned `true` 7 times, so that's how many total results it will claim to have.

Then, Offset = 2:
* `accumulate` will start out as `false`.
* Indexes 0, and 1 both return `true`, and are counted (but not added to the results).
* The count now equals the offset, so `accumulate` becomes `true`.
* Indexes 2-93 return `false`.
* Index 94 returns `true` and is counted/added.
* Index 95 returns `true` and is counted/added.
* There have been two `true`s, so the limit has been reached, and `accumulate` is `false` again.
* If still counting, indexes 96, 97, 98, and 99 all return `true`.
* It has returned `true` 8 times (2 from the start, 2 from accumulation, and 4 for after), and will report that there are now 8 results total.

Later, at Offset = 50:
* `accumulate` will start out as `false`.
* Indexes 0 to 49 all return `true`, and are counted (but not added to the results).
* The count now equals the offset, so `accumulate` becomes `true`.
* Indexes 50-93 return `false`.
* Index 94 returns `true` and is counted/added.
* Index 95 returns `true` and is counted/added.
* There have been two `true`s, so the limit has been reached, and `accumulate` is `false` again.
* If still counting, indexes 96, 97, 98, and 99 all return `true`.
* It has returned `true` 56 times (50 from the start, 2 from accumulation, and 4 for after), and will report that there are now 8 results total.
* It also returns the exact same two results that were returned with Offset = 2.

x/reward/keeper/time_management.go
arnabmitra: i agree on the concern . @Taztingo  has some results on it https://docs.google.com/spreadsheets/d/1dsqdnI999a7N3tsDAF7yrjpTd1vFI3bzJ8BOIINx2n0/edit#gid=0 with 100 rewards program running concurrently.

x/reward/keeper/query_server.go
nullpointer0x00: Good catch!  Thanks!

x/reward/keeper/query_server.go
arnabmitra: great explanation! @SpicyLemon 

moved all accumulate to only where `append` is used

x/reward/keeper/time_management.go
arnabmitra: I'll create a separate issue, similar to authz

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,204 @@
...
+						response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
+					}
+					break
+				}
+			}
+		default:
+			return false, vErr
+		}
+		return true, nil
+	})
```
SpicyLemon, line 64: This will never return `true`.

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,204 @@
...
+	kvStore := sdkCtx.KVStore(k.storeKey)
+	prefixStore := prefix.NewStore(kvStore, types.ClaimPeriodRewardDistributionKeyPrefix)
+	pageRes, err := query.FilteredPaginate(prefixStore, req.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
+		var claimPeriodRewardDist types.ClaimPeriodRewardDistribution
+		vErr := claimPeriodRewardDist.Unmarshal(value)
+		if accumulate {
+			if vErr == nil {
+				response.ClaimPeriodRewardDistributions = append(response.ClaimPeriodRewardDistributions, claimPeriodRewardDist)
+			} else {
+				return false, vErr
```
SpicyLemon, line 103: You shouldn't be returning inside the `accumulate` block.

x/reward/keeper/query_server.go
arnabmitra: i suck.. you're right..i consulted with @nullpointer0x00  and this is the whole function now 
```
	pageResponse, err := query.FilteredPaginate(prefixStore, req.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
		var rewardProgram types.RewardProgram
		vErr := rewardProgram.Unmarshal(value)
		switch {
		case vErr == nil && len(rewardProgramStates) == 0:
			if accumulate {
				response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
			}
			return true, nil
		case vErr == nil:
			for _, state := range rewardProgramStates {
				if rewardProgram.GetState() == state {
					if accumulate {
						response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
					}
					return true, nil
				}
			}
		default:
			return false, vErr
		}
		return false, nil
	})
```

x/reward/keeper/query_server.go
arnabmitra: fixed along with the ^^ comment

SpicyLemon: Still returning false in the `accumulate`.

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,205 @@
...
+					if accumulate {
+						response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
+					}
+					return true, nil
+				}
+			}
+		default:
+			return false, vErr
+		}
+		return false, nil
```
SpicyLemon, line 64: This is a mess.
```suggestion
		if vErr != nil {
			return false, vErr
		}
		count := len(rewardProgramStates) == 0
		for _, state := range rewardProgramStates {
			if rewardProgram.GetState() == state {
				count = true
				break
			}
		}
		if accumulate && count {
			response.RewardPrograms = append(response.RewardPrograms, rewardProgram)
		}
		return count, nil
```

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,205 @@
...
+		var claimPeriodRewardDist types.ClaimPeriodRewardDistribution
+		vErr := claimPeriodRewardDist.Unmarshal(value)
+		if accumulate {
+			if vErr == nil {
+				response.ClaimPeriodRewardDistributions = append(response.ClaimPeriodRewardDistributions, claimPeriodRewardDist)
+			} else {
+				return false, vErr
+			}
+		}
+		return true, nil
```
SpicyLemon, line 107:
```suggestion
		if vErr != nil {
			return false, vErr
		}
		if accumulate {
			response.ClaimPeriodRewardDistributions = append(response.ClaimPeriodRewardDistributions, claimPeriodRewardDist)
		}
		return true, nil
```

x/reward/spec/02_state.md
SpicyLemon: Still not fixed.

arnabmitra: my bad.. i didn't see the second place where size was mentioned..not sure why we are including size there anyways, but 🤷 https://github.com/provenance-io/provenance/pull/930/commits/fc42d1c5ad18be3d68563dda4c7d5d0a3cb436e4 that's the commit for the fix..

x/reward/keeper/query_server.go
Taztingo: Thanks for the suggestion. I changed it to be similar to this.

x/reward/keeper/query_server.go
Taztingo: It should be changed to something similar to this.

x/reward/client/cli/query.go
```diff
@@ -0,0 +1,281 @@
...
+	const active = "active"
+	const completed = "completed"
+	const outstanding = "outstanding"
+	cmd := &cobra.Command{
+		Use:     "reward-program {reward_program_id|\"all\"|\"pending\"|\"active\"\"completed\"|\"outstanding\"}",
+		Aliases: []string{"rp", "rewardprogram"},
+		Short:   "Query the current reward programs",
+		Long: fmt.Sprintf(`%[1]s reward-program {reward_program_id} - gets the reward program for a given id.
+%[1]s reward-program all - gets all the reward programs
+%[1]s reward-program active - gets all active the reward programs`, cmdStart),
```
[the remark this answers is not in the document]
Taztingo: Thanks, added a fix for this.

SpicyLemon (changes requested):
Still some general housekeeping to do.

Please don't resolve the comments. I'll resolve them when they've been resolved.

proto/provenance/reward/v1/reward.proto
```diff
@@ -0,0 +1,188 @@
...
+    CLAIM_STATUS_EXPIRED = 4;
+  }
+  // The id of the reward program that this share belongs to.
+  uint64 reward_program_id = 1;
+  // The id of the claim period that the share belongs to.
+  uint64 claim_period_id = 2;
+  // Owner of the reward account state.
+  string address = 3;
+  // The number of actions done by this account, mapped by action type.
+  map<string, uint64> action_counter = 4;
```
SpicyLemon, line 118: I'm pretty sure we cannot use maps in state stuff due to non-deterministic ordering.

x/reward/keeper/reward_claim.go
```diff
@@ -0,0 +1,170 @@
...
+
+// ClaimAllRewards calls ClaimRewards, however differs from ClaimRewards in that it claims all the rewards that the address
+// is eligible for across all reward programs.
+func (k Keeper) ClaimAllRewards(ctx sdk.Context, addr string) ([]*types.RewardProgramClaimDetail, sdk.Coins, error) {
+	var allProgramDetails []*types.RewardProgramClaimDetail
+	allRewards := sdk.Coins{}
+	err := k.IterateRewardPrograms(ctx, func(rewardProgram types.RewardProgram) (stop bool, err error) {
+		// ignore expired reward programs from all claim( i.e do not error on them)
+		if rewardProgram.State != types.RewardProgram_STATE_EXPIRED {
+			details, reward, err := k.ClaimRewards(ctx, rewardProgram.GetId(), addr)
```
SpicyLemon, line 144: It's generally bad practice to update state stuff while you're iterating over it. Weird things can happen to the iterator. A lot of times you can get away with it, specially when state is small, but it'll come back to bite you later.

x/reward/types/reward.go
```diff
@@ -0,0 +1,595 @@
...
+		return true
+	}
+
+	// If we don't have the event then we don't match it
+	_, exists := ec.Events[eventType]
+	return exists
+}
+
+func (ec *ABCIEvent) MatchesAttribute(name string, value []byte) bool {
+	attribute, exists := ec.Attributes[name]
```
SpicyLemon, line 85: Probably need to make sure `ec.Attributes` isn't nil first.

x/reward/spec/03_state_transitions.md
```diff
@@ -0,0 +1,63 @@
...
+-->
+
+# State Transitions
+
+This document describes the state transition operations pertaining to:
+
+1. [Reward Programs](./03_state_transitions.md#reward-programs)
+2. [Reward Claims](./03_state_transitions.md#reward-claims)
+
+<!-- TOC 2 2 -->
```
SpicyLemon, line 12: Need to `make update-tocs`.

SpicyLemon (commented):
Some final comments on the docs.

x/reward/spec/03_state_transitions.md
```diff
@@ -0,0 +1,63 @@
...
+A user may force a Reward Program in this state to end with the `end-reward-program` transaction. In this case the Reward Program will be deleted and not progress.
+
+### Started 
+The Reward Program has started, and users can participate by performing qualifying actions. Participants can claim their rewards at the end of the claim period that the qualifying action was performed in.
+
+### Finished 
+The Reward Program has ended, and participants can no longer make qualifying actions. Participants have a limited amount of time to collect their remaining rewards.
+
+#### Note
+A user may force a Reward Program in this state to end with the `end-reward-program` transaction. The Reward Program will transition to the `Finished` state on the next `BeginBlock`.
```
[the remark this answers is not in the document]
Taztingo: Good catch, it took me a while to understand the confusion. This Note should actually be under PENDING and STARTED. We may also want to make distinction between ending and "early ending". Maybe like canceled or something.

x/reward/spec/01_concepts.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Reward For Claim Period**
+
+$$\left( ClaimPeriodRewardPool \right) \times \left( EarnedShares \over ClaimPeriodShares \right) $$
+
+## Rollover
+It is possible that not all of the `Claim Period Reward Pool` will be distributed. This can happen when their is not enough activity, and participants are gated by the `max_reward_per_address`. The `Reward Program` will attempt to move these funds into a `Rollover Claim Period`. A `Rollover Claim Period` behaves exactly like any other `Claim Period`, but it is not guaranteed to have an equal portion of the original `Reward Program Reward Pool`. A `Reward Program` may run up to `max_rollover_claim_periods`, but is not guaranteed to run any of them. This is dependent on user activity, `program_end_time_max` field, and the `minimum_rollover_amount` field. Currently, the `minimum_rollover_amount` is set to 10% of the `Claim Period Reward Pool`.
```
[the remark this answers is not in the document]
Taztingo: Good catch.

x/reward/spec/03_state_transitions.md
```diff
@@ -0,0 +1,63 @@
+<!--
+order: 3
+-->
+
+# State Transitions
+
+This document describes the state transition operations pertaining to:
+
+1. [Reward Programs](./03_state_transitions.md#reward-programs)
+2. [Reward Claims](./03_state_transitions.md#reward-claims)
```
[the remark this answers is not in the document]
Taztingo: Agreed

x/reward/spec/03_state_transitions.md
Taztingo: I think this was covered in the conversation below.

cmd/provenanced/cmd/config_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/provenance-io/provenance/internal/pioconfig"
```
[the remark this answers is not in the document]
Taztingo: Moved it down with the others.

cmd/provenanced/cmd/init.go
```diff
@@ -8,6 +8,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/provenance-io/provenance/internal/pioconfig"
```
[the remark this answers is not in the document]
Taztingo: Moved it down with the others.

cmd/provenanced/cmd/root.go
```diff
@@ -8,6 +8,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/provenance-io/provenance/internal/pioconfig"
```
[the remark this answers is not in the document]
Taztingo: Moved it down with the others.

internal/handlers/msg_service_router_test.go
```diff
@@ -6,9 +6,18 @@ import (
...
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/golang/protobuf/proto"
+	rewardtypes "github.com/provenance-io/provenance/x/reward/types"
+
 	piosimapp "github.com/provenance-io/provenance/app"
 	"github.com/provenance-io/provenance/internal/antewrapper"
 	"github.com/provenance-io/provenance/internal/handlers"
+	"github.com/provenance-io/provenance/internal/pioconfig"
+
 	msgfeestypes "github.com/provenance-io/provenance/x/msgfees/types"
```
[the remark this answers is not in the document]
Taztingo: Adjusted all imports to match suggestions.

proto/provenance/reward/v1/tx.proto
```diff
@@ -0,0 +1,145 @@
...
+  uint64 claim_periods = 7;
+  // number of days a claim period will exist.
+  uint64 claim_period_days = 8;
+  // maximum number of claim periods a reward program can rollover.
+  uint64 max_rollover_claim_periods = 9;
+  // number of days before a reward program will expire after it has ended.
+  uint64 expire_days = 10;
+  // actions that count towards the reward.
+  repeated QualifyingAction qualifying_actions = 11
+      [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"qualifying_actions"];
```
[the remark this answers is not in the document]
Taztingo: LGTM

x/reward/keeper/reward_account_state_test.go
```diff
@@ -0,0 +1,424 @@
...
+	suite.Assert().Equal(uint64(1), accountState.GetRewardProgramId(), "reward program id must match")
+	suite.Assert().Equal(uint64(2), accountState.GetClaimPeriodId(), "reward claim period id must match")
+	suite.Assert().Equal(uint64(3), accountState.GetSharesEarned(), "earned shares must match")
+	suite.Assert().Equal("cosmos1ffnqn02ft2psvyv4dyr56nnv6plllf9pm2kpmv", accountState.GetAddress(), "address must match")
+	suite.Assert().Equal(types.RewardAccountState_CLAIM_STATUS_UNCLAIMABLE, accountState.GetClaimStatus(), "should be set to unclaimable initially")
+	suite.Assert().Equal(map[string]uint64{}, accountState.GetActionCounter(), "action counter must match")
+}
+
+func (suite *KeeperTestSuite) TestGetSetRewardAccountState() {
+	suite.SetupTest()
```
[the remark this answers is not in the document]
Taztingo: Agreed, removing now.

x/reward/keeper/reward_account_state_test.go
```diff
@@ -0,0 +1,424 @@
...
+	suite.Assert().Nil(err, "must not have error")
+	suite.Assert().Equal(expectedState.GetRewardProgramId(), actualState.GetRewardProgramId(), "reward program id must match")
+	suite.Assert().Equal(expectedState.GetClaimPeriodId(), actualState.GetClaimPeriodId(), "reward claim period id must match")
+	suite.Assert().Equal(expectedState.GetAddress(), actualState.GetAddress(), "address must match")
+	suite.Assert().Equal(expectedState.GetSharesEarned(), actualState.GetSharesEarned(), "shares earned must match")
+	suite.Assert().Equal(expectedState.GetClaimStatus(), actualState.GetClaimStatus(), "should be set to unclaimed initially")
+	suite.Assert().Equal(expectedState.GetActionCounter(), actualState.GetActionCounter(), "action counter must match")
+}
+
+func (suite *KeeperTestSuite) TestGetInvalidAccountState() {
```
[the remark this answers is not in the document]
Taztingo: Agreed and updated because they are just test suite receivers.

x/reward/keeper/query_server_test.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/cosmos/cosmos-sdk/types/query"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+func TestQueryServerTestSuite(t *testing.T) {
+	suite.Run(t, new(KeeperTestSuite))
+}
+
```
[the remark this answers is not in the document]
Taztingo: Verified that they are in the same package. Removed duplicate.

x/reward/types/reward.go
Taztingo: If ec.Attributes is nil then exists will be false.

https://go.dev/play/p/K26sK9wUrlZ?v=goprev

x/reward/keeper/reward_claim.go
Taztingo: Agreed, the documentation even mentions this.

x/reward/types/reward.go
SpicyLemon: Oh! Nice! I could have sworn I've gotten nil pointer errors when trying to access a nil map before. Maybe I'm misremembering that or maybe they fixed it in 1.18. Either way. Nice.

x/reward/spec/02_state.md
```diff
@@ -0,0 +1,75 @@
...
+* Reward Program ID: `0x02 -> uint64(RewardProgramID)`
+
++++ https://github.com/provenance-io/provenance/blob/cba0b20229234bd655d7aefdb5aa5195e1c01214/proto/provenance/reward/v1/reward.proto#L13-L99
+
+---
+## Claim Period Reward Distribution
+
+A `ClaimPeriodRewardDistribution` is created for each claim period of every `RewardProgram`. Its purpose is to track live claim period specific information. Examples of this include the total number of granted shares in the claim period, sum of of all its rewards given out as claims, and the amount of reward allocated to it from the `RewardProgram`.
+
+* Claim Period Reward Distribution: `0x03 | Claim Period ID (8 bytes) | Reward Program ID (8 bytes) -> ProtocolBuffers(ClaimPeriodRewardDistribution)`
```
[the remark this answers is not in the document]
Taztingo: Updated the documentation and the actual key to reflect these suggestions.

proto/provenance/reward/v1/reward.proto
SpicyLemon: Yeah.
https://developers.google.com/protocol-buffers/docs/proto3#maps
> Wire format ordering and map iteration ordering of map values are undefined, so you cannot rely on your map items being in a particular order.
So this will need to be changed to a `repeated` new message with a `string` and `int64`.

x/reward/client/cli/cli_test.go
```diff
@@ -0,0 +1,916 @@
...
+	}{
+		{"query all reward by address",
+			s.accountAddr.String(),
+			"all",
+			false,
+			"",
+			0,
+			[]uint64{1, 2, 3, 4},
+			100,
+		},
```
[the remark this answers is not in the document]
Taztingo: Added named params to make tests easier to follow.

x/reward/client/cli/cli_test.go
```diff
@@ -0,0 +1,916 @@
...
+		expectErrMsg       string
+		expectedCode       uint32
+		signer             string
+	}{
+		{"end reward program - valid",
+			"1",
+			"",
+			0,
+			s.accountAddresses[0].String(),
+		},
```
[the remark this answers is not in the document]
Taztingo: Added named params to make tests easier to follow.

x/reward/types/action_builder.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (b *TransferActionBuilder) CanBuild() bool {
+	return !b.Sender.Empty() && !(len(b.Action) == 0) && !b.Recipient.Empty()
+}
+
+func (b *TransferActionBuilder) BuildAction() (EvaluationResult, error) {
+	if !b.CanBuild() {
+		return EvaluationResult{}, nil
+	}
```
SpicyLemon, line 91: The other action builders return an error in this case. Should this one too or is there something special for this one? If special, a comment would help.

x/reward/keeper/msg_server.go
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/msg_server.go
```diff
@@ -0,0 +1,146 @@
...
+	return &types.MsgClaimRewardsResponse{
+		ClaimDetails: types.RewardProgramClaimDetail{
+			RewardProgramId:            req.GetRewardProgramId(),
+			TotalRewardClaim:           reward,
+			ClaimedRewardPeriodDetails: details,
+		},
+	}, nil
+}
+
+func (s msgServer) ClaimAllRewards(goCtx context.Context, req *types.MsgClaimAllRewardsRequest) (*types.MsgClaimAllRewardsResponse, error) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,202 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+var _ types.QueryServer = Keeper{}
+
+func (k Keeper) RewardPrograms(ctx context.Context, req *types.QueryRewardProgramsRequest) (*types.QueryRewardProgramsResponse, error) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("unable to query all reward programs: %v", err))
+	}
+	response.Pagination = pageResponse
+	return &response, nil
+}
+
+func (k Keeper) RewardProgramByID(ctx context.Context, req *types.QueryRewardProgramByIDRequest) (*types.QueryRewardProgramByIDResponse, error) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/query_server.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	if ClaimPeriodReward.Validate() == nil {
+		response.ClaimPeriodRewardDistribution = &ClaimPeriodReward
+	}
+
+	return &response, nil
+}
+
+func (k Keeper) RewardDistributionsByAddress(ctx context.Context, request *types.QueryRewardDistributionsByAddressRequest) (*types.QueryRewardDistributionsByAddressResponse, error) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/reward_account_state.go
```diff
@@ -0,0 +1,168 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/provenance-io/provenance/x/reward/types"
+)
+
+func (k Keeper) GetRewardAccountState(ctx sdk.Context, rewardProgramID, rewardClaimPeriodID uint64, addr string) (state types.RewardAccountState, err error) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/keeper/reward_account_state.go
```diff
@@ -0,0 +1,168 @@
...
+	// an empty map and nil. It's solution is to always make
+	// it nil because of the spec.
+	if state.ActionCounter == nil {
+		state.ActionCounter = make(map[string]uint64)
+	}
+
+	return state, err
+}
+
+func (k Keeper) SetRewardAccountState(ctx sdk.Context, state types.RewardAccountState) {
```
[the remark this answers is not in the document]
Taztingo: Added

x/reward/module/module.go
```diff
@@ -0,0 +1,194 @@
+package reward
+
+import (
+	"context"
+	"encoding/json"
+	"math/rand"
+
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
```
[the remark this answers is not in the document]
Taztingo: Reordered imports.

x/reward/keeper/reward_program.go
```diff
@@ -0,0 +1,194 @@
...
+	return k.getRewardProgramByState(ctx, types.RewardProgram_STATE_PENDING)
+}
+
+// GetAllExpiredRewardPrograms gets all expired RewardPrograms
+func (k Keeper) GetAllExpiredRewardPrograms(ctx sdk.Context) ([]types.RewardProgram, error) {
+	return k.getRewardProgramByState(ctx, types.RewardProgram_STATE_EXPIRED)
+}
+
+// GetUnexpiredRewardPrograms gets all RewardPrograms that are not expired
+func (k Keeper) GetUnexpiredRewardPrograms(ctx sdk.Context) ([]types.RewardProgram, error) {
```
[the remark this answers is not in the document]
Taztingo: Changed name to match other functions.

x/reward/keeper/reward_program.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		if stop {
+			break
+		}
+	}
+	return nil
+}
+
+// GetOutstandingRewardPrograms Gets all RewardPrograms that have not expired
+func (k Keeper) GetOutstandingRewardPrograms(ctx sdk.Context) ([]types.RewardProgram, error) {
```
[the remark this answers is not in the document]
Taztingo: Agreed, updated name to be consistent with others.

x/reward/types/action_builder.go
arnabmitra: should return error..changed

SpicyLemon:
Really close. The only outstanding thing I see is https://github.com/provenance-io/provenance/pull/930#discussion_r966239701 about having a `map` in the proto.

arnabmitra:
> Really close. The only outstanding thing I see is [#930 (comment)](https://github.com/provenance-io/provenance/pull/930#discussion_r966239701) about having a `map` in the proto.

Agreed on the map change, we talked about it this morning, but i am not sure where we landed on all the changes needed to remove it, will definitely get it into the PR by Monday.. good catch

SpicyLemon (approved):
Looks good!

proto/provenance/reward/v1/reward.proto
```diff
@@ -0,0 +1,197 @@
...
+
+  // Minimum number of successful votes.
+  uint64 minimum_actions = 1;
+  // Maximum number of successful votes.
+  uint64 maximum_actions = 2;
+  // Minimum delegation amount the account must have across all validators, for the vote action to be counted.
+  cosmos.base.v1beta1.Coin minimum_delegation_amount = 3 [(gogoproto.nullable) = false];
+}
+
+// The number of actions done by this account, key --> action type, value --> actions performed count.
```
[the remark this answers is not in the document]
arnabmitra: comment looks already updated

Issue #934, opened 2022-07-14
Please add support for Arm64 


Alien901:
## Summary of Bug

Getting errors on Arm64

## Version

v1.11.1

## Steps to Reproduce

make install

`provenance$ git checkout tags/v1.11.1 -b v1.11.1
Switched to a new branch 'v1.11.1'
ubuntu@provenance-20220714-1627:~/provenance$ make clean
rm -rf /home/ubuntu/provenance/build/*
ubuntu@provenance-20220714-1627:~/provenance$ make install
--> Ensure dependencies have not been modified
all modules verified
go: downloading github.com/golang/mock v1.6.0
go: downloading github.com/frankban/quicktest v1.14.3
go: downloading github.com/cosmos/gorocksdb v1.2.0
go: downloading github.com/dgraph-io/badger/v2 v2.2007.4
go: downloading github.com/jmhodges/levigo v1.0.0
go: downloading go.etcd.io/bbolt v1.3.6
go: downloading github.com/inconshreveable/mousetrap v1.0.0
go: downloading github.com/otiai10/mint v1.3.3
go: downloading github.com/pascaldekloe/goe v0.1.0
go: downloading github.com/google/go-cmp v0.5.8
go: downloading github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
go: downloading github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
go: downloading gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
go: downloading github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
go: downloading github.com/fortytw2/leaktest v1.3.0
go: downloading github.com/kr/pretty v0.3.0
go: downloading github.com/sasha-s/go-deadlock v0.2.1-0.20190427202633-1595213edefa
go: downloading github.com/facebookgo/ensure v0.0.0-20200202191622-63f1cf65ac4c
go: downloading github.com/dgraph-io/ristretto v0.0.3
go: downloading github.com/dgryski/go-farm v0.0.0-20200201041132-a6ae2369ad13
go: downloading github.com/dustin/go-humanize v1.0.0
go: downloading github.com/onsi/ginkgo v1.16.4
go: downloading github.com/onsi/gomega v1.13.0
go: downloading github.com/jhump/protoreflect v1.9.0
go: downloading github.com/cosmos/ledger-cosmos-go v0.11.1
[...]

PR #938, opened 2022-07-14
Taztingo/934 arm64 linux support

Taztingo:
## Description

The purpose of this PR is to make it possible to build on Linux ARM64 out of the box. Validation scripts were added to look for needed dependencies, during the build process.Additionally, the librdkafka target was updated to include Linux ARM64.

closes: #934

---

write a little note why.


scripts/m1-dependency-check.sh
```diff
@@ -23,14 +23,15 @@ do
...
     fi
 done
 
 # Check that PKG_CONFIG_PATH is set
-case ":$PKG_CONFIG_PATH:" in
-  *:"$( brew --prefix openssl )/lib/pkgconfig":*) : ;;
-  *) echo "PKG_CONFIG_PATH is missing openssl. Recommended to run: export PKG_CONFIG_PATH=\"\$( brew --prefix openssl )\"/lib/pkgconfig\"\${PKG_CONFIG_PATH:+:\$PKG_CONFIG_PATH}\""; exit 1;
-esac
+op="$( brew --prefix openssl )/lib/pkgconfig"
+if ! tr ':' '\n' <<< $PKG_CONFIG_PATH | grep -xFq "$op"; then
```
SpicyLemon, line 33: Need some extra quotes in here:
```suggestion
if ! tr ':' '\n' <<< "$PKG_CONFIG_PATH" | grep -xFq "$op"; then
```

scripts/linux_arm64_librdkafka_install.sh
```diff
@@ -0,0 +1,30 @@
...
+    make install
+    cd ${PWD}
+
+    # Cleanup
+    rm -rf $TEMP
+fi
+
+# Check if we have the env varibale set
+if ! tr ':' '\n' <<< "$LD_LIBRARY_PATH" | grep -xFq "$LIB_PATH"; then
+    echo 'LD_LIBRARY_PATH is not set. Please set it with export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib'
```
SpicyLemon, line 27: Tiny nit: The `:` is only needed if there are things already in `LD_LIBRARY_PATH`. That's why the other stuff uses this construct: `${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}`.

That uses variable expansion to do some magic. Basically, if `PKG_CONFIG_PATH` isn't empty (the `:+` part), return `:` then `$PKG_CONFIG_PATH`. But if it's empty, it still just returns empty. Where this is being used, the entry is being added to the front of the variable, so that's why the `:` comes first. In this case, you could do `${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}$LIB_PATH`.

```suggestion
    echo "LD_LIBRARY_PATH is not set. Please set it with export LD_LIBRARY_PATH=\${LD_LIBRARY_PATH:+\$LD_LIBRARY_PATH:}$LIB_PATH"
```

scripts/linux_arm64_librdkafka_install.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+TEMP=/tmp/librdkafka
+LIB_PATH=/usr/local/lib
```
SpicyLemon, line 4: If you use `LIB_PATH=${LIB_PATH:-/usr/local/lib}`, it allows this to be set from outside the script.

This is variable expansion. The `:-` is "If LIB_PATH is empty, expand to the following."
So basically, if someone has `export LIB_PATH='/usr/bin/'` in their environment, and runs this script, that's the path that would get looked at.

If you made that change here, you'd want to make the same change to `scripts/linux-arm64-dependency-check.sh`.

scripts/m1-dependency-check.sh
Taztingo: Good catch thanks! Should be resolved now.

scripts/linux_arm64_librdkafka_install.sh
Taztingo: That makes sense, added the change.

dwedul-figure (approved):

arnabmitra (approved):

iramiller (approved):
Happy to see that other fix for the path check in here too

Issue #940, opened 2022-07-14
ABCI Request Response Streaming


egaxhaj:
## Summary

Enable listening of `BeginBlocker` and `EndBlocker` request/response events to external consumers.

## Problem Definition

Currently block data is accessed through [Queries](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/messages-and-queries.md#queries) which proceed either through Tendermint and the ABCI API, or through the gRPC server. In addition to these queries, it would be beneficial to have a means of streaming ABCI  req/res to external system(s) as they occur in real-time.

**Why do we need this feature?**
 - To provide a means of consuming `BeginBlocker` and `EndBlocker` req/res events in real-time.
 
**What problems may be addressed by introducing this feature?**
- Allow consuming applications of blockchain data to move away from using the query API and gRPC server and into real-time processing from an external resource. 

**What benefits does the Provenance Network stand to gain by including this feature?**
- Consumer applications no longer need to query nodes directly through the query API or through the gRPC server. Instead, events can be consumed from and external system (i.e: Kafka) in real-time. This frees up resources on he node to do block processing. 

**Should this feature be implemented in the Cosmos SDK directly?**
- This feature is already being worked in [ADR-038](https://docs.cosmos.network/master/architecture/adr-038-state-listening.html). However, the current implementation leads to an `AppHash` mismatch who's cause it undetermined at this time. The implementation suggested in this feature is a lightweight version of ADR-038 but **without** the state-listening component.

**Are there any disadvantages of including this feature?**
- No known disadvantages.

Subtasks:
- [ ] #951

## Proposal

Based on the above requirements we will introduce a new `StreamingService` interface and change `app.BeginBlocker` and `app.EndBlocker` implementations.

#### Streaming Service interface
In a new file, `internal/streaming/streaming.go`, we will introduce `StreamService` and `StreamServiceInitializer` interfaces that plug into the `app` and stream out ABCI request and responses.

```go
// StreamingService interface used to hook into the ABCI message processing of the BaseApp
type StreamingService interface {
	// StreamBeginBlocker updates the streaming service with the latest BeginBlock messages
	StreamBeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, res abci.ResponseBeginBlock)
	// StreamEndBlocker updates the steaming service with the latest EndBlock messages
	StreamEndBlocker(ctx sdk.Context, req abci.RequestEndBlock, res abci.ResponseEndBlock)
}

[...]

PR #952, opened 2022-07-19
add api support for abci streaming to external systems

egaxhaj:
## Description

Adds support for streaming `BeginBlocker` and `EndBlocker` request and responses to external systems. 

closes: #940

---

write a little note why.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -945,12 +945,7 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 
 // RegisterStreamingService is used to register a streaming service into the App
 func (app *App) registerStreamingService(appOpts servertypes.AppOptions) {
-	app.Logger().Info(fmt.Sprintf("Streaming Service Config options: "+
-		"\n%s/config/app.toml"+
-		"\n----------------------------------"+
-		"\nset streaming.enable true"+
-		"\nset streaming.service to one of %s\n",
-		os.Getenv("PIO_HOME"), appstreaming.ConfigOptions()))
+	app.Logger().Info(fmt.Sprintf("Available streaming services: %s", appstreaming.ConfigOptions()))
 	enableStreaming := cast.ToBool(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.EnableParam)))
 	if enableStreaming {
 		serviceTomlKey := fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.ServiceParam)
--- a/app/app.go
+++ b/app/app.go
@@ -286,8 +286,8 @@ type App struct {
 	// module configurator
 	configurator module.Configurator
 
-	// ABCI streaming services
-	streamingServices []streaming.StreamService
+	// ABCI streaming service
+	streamingService streaming.StreamService
 }
 
 func init() {
@@ -834,14 +834,15 @@ func New(
 	app.ScopedTransferKeeper = scopedTransferKeeper
 
 	// register streaming service
-	enabledServices := cast.ToStringSlice(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.EnabledParam)))
-	for _, key := range enabledServices {
-		ssi, found := appstreaming.StreamServiceInitializers[key]
+	streamingEnabled := cast.ToBool(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.EnabledParam)))
+	if streamingEnabled {
+		service := cast.ToString(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.ServiceParam)))
+		ssi, found := appstreaming.StreamServiceInitializers[service]
 		if found {
 			app.RegisterStreamingService(ssi.Init(appOpts, app.AppCodec()))
-			logger.Info("registered streaming service", "service", ssi)
+			logger.Info("Starting ABCI external streaming", "service", service)
 		} else {
-			logger.Error("unsupported streaming service", "service", ssi)
+			logger.Error("unsupported ABCI streaming service", "service", service)
 		}
 	}
 
@@ -854,8 +855,8 @@ func (app *App) Name() string { return app.BaseApp.Name() }
 // BeginBlocker application updates every begin block
 func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
 	res := app.mm.BeginBlock(ctx, req)
-	for _, s := range app.streamingServices {
-		s.StreamBeginBlocker(ctx, req, res)
+	if app.streamingService != nil {
+		app.streamingService.StreamBeginBlocker(ctx, req, res)
 	}
 	return res
 }
@@ -863,8 +864,8 @@ func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.R
 // EndBlocker application updates every end block
 func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.ResponseEndBlock {
 	res := app.mm.EndBlock(ctx, req)
-	for _, s := range app.streamingServices {
-		s.StreamEndBlocker(ctx, req, res)
+	if app.streamingService != nil {
+		app.streamingService.StreamEndBlocker(ctx, req, res)
 	}
 	return res
 }
@@ -984,7 +985,7 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 // RegisterStreamingService is used to register a streaming service into the App
 func (app *App) RegisterStreamingService(s streaming.StreamService) {
 	// App will pass BeginBlocker and EndBlocker requests and responses to the streaming services
-	app.streamingServices = append(app.streamingServices, s)
+	app.streamingService = s
 }
 
 // RegisterSwaggerAPI registers swagger route with API Server
--- a/app/app.go
+++ b/app/app.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
--- a/app/app.go
+++ b/app/app.go
@@ -3,8 +3,6 @@ package app
 import (
 	"encoding/json"
 	"fmt"
-	appStreaming "github.com/provenance-io/provenance/app/streaming"
-	"github.com/provenance-io/provenance/internal/streaming"
 	"io"
 	"net/http"
 	"os"
@@ -104,11 +102,13 @@ import (
 
 	// PROVENANCE
 	appparams "github.com/provenance-io/provenance/app/params"
+	appstreaming "github.com/provenance-io/provenance/app/streaming"
 	_ "github.com/provenance-io/provenance/client/docs/statik" // registers swagger-ui files with statik
 	"github.com/provenance-io/provenance/internal/antewrapper"
 	piohandlers "github.com/provenance-io/provenance/internal/handlers"
 	"github.com/provenance-io/provenance/internal/provwasm"
 	"github.com/provenance-io/provenance/internal/statesync"
+	"github.com/provenance-io/provenance/internal/streaming"
 	"github.com/provenance-io/provenance/x/attribute"
 	attributekeeper "github.com/provenance-io/provenance/x/attribute/keeper"
 	attributetypes "github.com/provenance-io/provenance/x/attribute/types"
@@ -836,7 +836,7 @@ func New(
 	// register streaming service
 	enabledServices := cast.ToStringSlice(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.EnabledParam)))
 	for _, key := range enabledServices {
-		ssi, found := appStreaming.StreamServiceInitializers[key]
+		ssi, found := appstreaming.StreamServiceInitializers[key]
 		if found {
 			app.RegisterStreamingService(ssi.Init(appOpts, app.AppCodec()))
 			logger.Info("registered streaming service", "service", ssi)
--- a/app/app.go
+++ b/app/app.go
@@ -834,17 +834,7 @@ func New(
 	app.ScopedTransferKeeper = scopedTransferKeeper
 
 	// register streaming service
-	enableStreaming := cast.ToBool(appOpts.Get(fmt.Sprintf("%s.%s", streaming.TomlKey, streaming.EnableParam)))
-	if enableStreaming {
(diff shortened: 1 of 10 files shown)


internal/streaming/streaming.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// TOML configuration parameter keys
+const (
+	// TomlKey is the top-level TOML key for streaming configuration
+	TomlKey = "streaming"
+
+	// EnabledParam is the flag that enables ABCI request response streaming services
+	EnabledParam = "enabled"
+)
```
SpicyLemon, line 17: If this is expected to be a custom addition to one of the config files, it's not going to jive well with our `config` commands. Specifically, the `set`, `pack`, and `unpack` commands will stomp out any such custom entry (since they're not part of the template).

Annoyingly, for the `plugins` config, it looks like they went with the key `on`, where their other sections use `enabled`, and then `plugins.enabled` as a slice of strings with the names of the plugins to use.

In my opinion, `enabled` should be a boolean indicating whether or not that feature should be active. Then something like `services` or `names` would be a slice of strings with the names of the plugins to use.

Regardless, as far as I can tell, this `streaming` config section is new. ADR-048 mentions a `plugins.streaming` section, but no `plugins.streaming.enabled` or `streaming.enabled` entries and no `streaming` section.

If any of this is new, and specific to provenanced (i.e. not generic or better defined by Cosmos-SDK), we should consider creating our own config file to house these things. If that's the case, we'll want to do that as a separate issue though. But either way, if this is a new config thing, there needs to be documentation on it.

app/streaming/streaming.go
```diff
@@ -0,0 +1,9 @@
+package streaming
+
+// StreamServiceInitializers contains a map of predefined StreamServiceInitializer implementations
```
SpicyLemon, line 3: Misplaced/duplicate comment.
```suggestion
```

app/streaming/streaming.go
```diff
@@ -0,0 +1,9 @@
...
+import (
+	"github.com/provenance-io/provenance/internal/streaming"
+)
+
+// StreamServiceInitializers contains a map of supported StreamServiceInitializer implementations
+var StreamServiceInitializers = map[string]streaming.StreamServiceInitializer{}
```
SpicyLemon, line 9: I'm guessing that this is something that's supposed to be added to during the `init` function for each plugin.

Do we not have any examples or plugins that we want available right now?

It'd be great if there were some unit/integration tests on this stuff too.

app/app.go
```diff
@@ -828,6 +833,18 @@ func New(
 	app.ScopedIBCKeeper = scopedIBCKeeper
 	app.ScopedTransferKeeper = scopedTransferKeeper
 
+	// register streaming service
```
SpicyLemon, line 836: There should be an `Info` log line here that lists all the keys available for streaming configuration. That way, people can more easily see what's available and know what keys to use when setting their config.

app/app.go
```diff
@@ -836,12 +853,20 @@ func (app *App) Name() string { return app.BaseApp.Name() }
 
 // BeginBlocker application updates every begin block
 func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
-	return app.mm.BeginBlock(ctx, req)
+	res := app.mm.BeginBlock(ctx, req)
+	for _, s := range app.streamingServices {
+		s.StreamBeginBlocker(ctx, req, res)
+	}
```
SpicyLemon, line 859: Should this (and the similar new lines in `EndBlocker` be fired off as separate `go` routines? E.g.:
```golang
go func() {
	for _, s := range app.streamingServices {
		s.StreamBeginBlocker(ctx, req, res)
	}
}()
```

Pro: The `BeginBlocker` function doesn't have to wait on any of the calls to `StreamBeginBlocker`.
Con: Possible memory concerns about holding onto the `req` and `res` values outside the running of the `BeginBlocker` or `EndBlocker` functions. Also possible concurrency concerns if one isn't done yet, but `BeginBlocker` or `EndBlocker` is called again.

SpicyLemon (commented):
This needs unit/integration tests.

I'm also unsure what the difference between this stuff and the plugins stuff is. An actual use case and some unit tests would probably help with that respect though.

internal/streaming/streaming.go
egaxhaj: I borrowed from the plugin system in ADR-038. However, it is different because in this implementation we're not able to capture any state change events. The reason for this lighter imlementation is because adr-038 has an `AppHash` error that's being caused either how state-change events are pushed to the listeners or how the plugin system is injected. In a discussion with @iramiller I came up with this version.

This implementation and its configuration is custom to Provenance. So we can change the config names. The `streaming` section is new and has no relation to `plugins.streaming` from ADR-038. Per @iramiller we want to disable ADR-038 implementation we have in our Cosmos SDK fork and wait for upstream to merge ADR-038 (This has seen some progress as of the last two days).

For the config. I actually think its better we remove `enabled` all together. I don't think we should actually have more than one service enabled at a time. 
We would just have `[streaming]` and `streaming.service = kafka | postgresql, trace' etc. But only one enabled.  So the `BeginBlocker` and `EndBlocker` only push out to one external service. Than clients can consumer from that external service. I'll change the design to reflect this. 

I would love some help on figuring out the config so it does work with `provenance config set`.

app/streaming/streaming.go
egaxhaj: I'll add a lightweight example that logs to `trace`.

app/app.go
egaxhaj: The intent is to stay in sync with the external service. Otherwise, the node may jump a head blocks when the external service fails. Ideally, streaming to external service would run on a non-validator node. This way there's no concern of performance issues. We should actually document this in the config. We can put a `goroutine` in there but it will need to be with a `wg` (with Kafka we'll produce [asynchronously](https://github.com/confluentinc/confluent-kafka-go#examples)) .

I'm actually changing the design to only have one service enabled at a time. We should support multiple services. I don't know what I was thinking there? We only need to push out to one external service and have consumer applications `fan-out` from the external service.

```toml
[stremaing]

# trace, kafka, postgresql
service = ""
```

internal/streaming/streaming.go
egaxhaj: To add to this a little more I also have a `Kafka` implementation on as separate [branch](https://github.com/provenance-io/provenance/tree/951-abci-beginblocker-and-endblocker-streaming-to-kafka). I wanted to separate the API support from service implementations.

SpicyLemon: > I would love some help on figuring out the config so it does work with provenance config set.

The `config` command stuff all works off of reflection on the config structs. Writing the config files is then done by providing the appropriate config struct to the appropriate toml template. Since none of the templates have these fields, they'll disappear any time the config file is programmatically updated.

I've been thinking for quite some time that it'd be nice if we had our own `provenance.toml` config file (and associated object). But the only thing I could think to put in there so far is the `testnet` flag. If we want this `streaming` config to be in a file, that's probably the best way to go.

We'd want to do that as a separate issue/PR though, so for this, we just need documentation about what the config fields are, and what's expected. That way, we know what to make our template/struct look like.

Also, I mentioned this in the Cosmos PR, but the standard boolean on/off field name is `enable`. I like having such a field because it allows you to turn something off without losing how you've configured it. So there'd be a `streaming.enable` boolean and `streaming.service` string.

iramiller: It would be interesting to see if there could be a way an app like provenance using the sdk/tendermint could create its own consolidated and documented toml file to replace all the others and simplify...

SpicyLemon: Summary:

1. I've created #962 for the new config file.
2. The on/off field should be `enable` (not `enabled`).
3. Make sure the documentation (e.g. in `app/streaming/README.md`) is accurate.

Until #962 is completed, we'll just have to live with the knowledge that the `config` commands a) won't find these new fields, and b) might unconfigure this stuff.

egaxhaj: 👍 I'll get started on the `proveance.toml`.

app/app.go
egaxhaj: Now that we're going to create our own `provenance.config` => #962, do we still want to do this here, somewhere else or not at all?

dwedul-figure: This isn't really all that related to that issue. Regardless of where the config lives, people still need to know what strings are expected so that they can be properly configured.

Also, maybe this should panic if the configured key doesn't exist (rather than just logging an error and moving on).

app/streaming/README.md
```diff
@@ -0,0 +1,92 @@
...
+    // StreamEndBlocker updates the steaming service with the latest EndBlock messages
+    StreamEndBlocker(ctx sdk.Context, req abci.RequestEndBlock, res abci.ResponseEndBlock)
+}
+```
+
+Take a look at [app/streaming/trace/service/service.go](./trace/service/service.go) for an implementation example of this interface.
+
+## Configuration
+
+Streaming services are configured in the `streaming` TOML mapping in the App's `app.toml` file. There are two parameters
```
SpicyLemon, line 59: With the addition of the `custom.toml` option, we should say to put this in there.

```suggestion
Streaming services are configured in the `streaming` TOML mapping which can be placed in the App's `custom.toml` file. There are two parameters
```

app/streaming/streaming.go
```diff
@@ -0,0 +1,21 @@
...
+var StreamServiceInitializers = map[string]streaming.StreamServiceInitializer{
+	"trace": trace.StreamServiceInitializer,
+}
+
+func ConfigOptions() string {
+	keys := make([]string, 0, len(StreamServiceInitializers))
+	for k := range StreamServiceInitializers {
+		keys = append(keys, k)
+	}
+	return strings.Join(keys, ",")
```
SpicyLemon, line 20: Probably a good idea to sort the keys before returning them. The returned value might not be used anywhere to cause determinism problems, but it's probably best to be on the safe side.

app/streaming/trace/service/service.go
```diff
@@ -0,0 +1,87 @@
...
+func (tss *TraceStreamingService) StreamBeginBlocker(
+	ctx sdk.Context,
+	req abci.RequestBeginBlock,
+	res abci.ResponseBeginBlock,
+) {
+	errMsg := "BeginBlocker listening hook failed"
+
+	// write req
+	if err := tss.write(ctx, &req); err != nil {
+		ctx.Logger().Error(errMsg, "height", req.Header.Height, "err", err)
```
SpicyLemon, line 43: There needs to be something here and in the other one that allows someone to look at the log line and distinguish between the two errors. Maybe add the args `, "action", "write request"` and `, "action", "write response"` to each log statement respectively.

Same goes for the two log messages in the endblocker listener hook.

app/streaming/trace/service/service_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/tendermint/tendermint/libs/log"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	abci "github.com/tendermint/tendermint/abci/types"
+	types1 "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+var (
```
SpicyLemon, line 16: nit: Global variables are an easy way to end up with broken things.

I admit that It's probably safe in this limited instance, but is something to avoid in most cases (in favor of the `"github.com/stretchr/testify/suite"` approach).

app/streaming/trace/trace.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/provenance-io/provenance/internal/streaming"
+	"github.com/spf13/cast"
+)
+
+const (
+	// TomlKey is the top-level TOML key for TraceStreamingService configuration
+	TomlKey = "trace"
+
+	// PrintDataToStdoutParam is the TraceStreamingService flag that logs full streamed data at debug level
+	PrintDataToStdoutParam = "print_data_to_stdout"
```
SpicyLemon, line 18: In Tendermint `v0.35`, they changed all config field names to use dashes instead of underscores. Cosmos-SDK already uses only dashes in their config entries. Unless there's a good reason to use underscores here, we probably want to go with dashes instead:
```suggestion
	PrintDataToStdoutParam = "print-data-to-stdout"
```
The documentation would also need updating.

app/app.go
```diff
@@ -926,6 +943,30 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
...
 }
 
+// RegisterStreamingService is used to register a streaming service into the App
+func (app *App) registerStreamingService(appOpts servertypes.AppOptions) {
+	app.Logger().Info(fmt.Sprintf("Streaming Service Config options: "+
+		"\n%s/config/app.toml"+
+		"\n----------------------------------"+
+		"\nset streaming.enable true"+
+		"\nset streaming.service to one of %s\n",
+		os.Getenv("PIO_HOME"), appstreaming.ConfigOptions()))
```
SpicyLemon, line 953: This is probably too much, and I find it confusing. It took me a while to realize that it was logging what the config would need to be in order to enable streaming as opposed to what it is. That's best left for documentation though. All that's really desired here is the list of keys (services) that you could set `streaming.service` to. Just something like this:
```suggestion
	app.Logger().Info(fmt.Sprintf("Available streaming services: %s", appstreaming.ConfigOptions()))
```

Other notes about this:
1. The `app.toml` file isn't necessarily where this is configured. It could be any of the config files. We should be directing users to the `custom.toml` for this now anyway.
2. `os.Getenv("PIO_HOME")` only gets the home directory when it's defined by an environment variable. But it can also be set by either a `--home` flag or the default. I'm not sure you can properly get the home directory from the info you have in here.

SpicyLemon (commented):
Just a couple more small things.

SpicyLemon (approved):

app/streaming/trace/service/service.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx sdk.Context,
+	req abci.RequestEndBlock,
+	res abci.ResponseEndBlock,
+) {
+	errMsg := "EndBlocker listening hook failed"
+
+	// write req
+	if err := tss.write(ctx, &req); err != nil {
+		ctx.Logger().Error(errMsg, "action", "write request", "height", ctx.BlockHeight(), "err", err)
+		panic(err)
```
arnabmitra, line 65: do we really want to panic on a node, if streaming fails?

arnabmitra: i guess if it's a validator it probably doesn't matter much, but still leaving comment out here..

arnabmitra (approved):
i think this PR looks good to me, the "omitted" use case i am not sure i understand but it being controlled by the flag makes it sound code wise..

egaxhaj:
closing this for now in favor of a `grpc` approach discussed [here](https://github.com/cosmos/cosmos-sdk/pull/11691#issuecomment-1201875170). Will keep the repo open for reference.

Issue #1010, opened 2022-08-15
Replace use of deprecated sdkResponse.Data field


SpicyLemon:
## Summary

The `sdkResponse.Data` field is deprecated and the MsgResponses should be used instead.

## Problem Definition

In `internal/provwasm/simulation.go` we use the `sdkResponse.Data` field that is now deprecated.

## Proposal

Get what's needed from the `MsgResponses` instead of `sdkResponse.Data`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #995, opened 2022-08-05
Upgrade provenance to use our Cosmos-SDK fork's `v0.46.1-pio-1-rc1`.


SpicyLemon:
Bump Cosmos-SDK to `v0.46.0-pio-1-rc4` (from `v0.45.5-pio-1`).

SpicyLemon:
A couple issues were found with `rc1`. They were fixed and are now in `rc2`. So now that's the goal of this ticket.
Fixes:
* [222](https://github.com/provenance-io/cosmos-sdk/pull/222): Create the authz GetAuthorization function
* [223](https://github.com/provenance-io/cosmos-sdk/pull/223): Create IMsgServiceRouter interface and baseapp.SetMsgServiceRouter function

SpicyLemon:
I've currently got it pulling in a local `wasmd` that has been updated to use Cosmos v0.46. But that's just a draft PR and isn't tagged yet or anything.

The issue: https://github.com/CosmWasm/wasmd/issues/927
The Draft PR: https://github.com/CosmWasm/wasmd/pull/930

We might have to pull that into our own `wasmd` fork and tag it so we can start moving forward.

SpicyLemon:
Okay. I've got most of the unit tests ironed out. There's still a couple stragglers, and then one giant mess with the msg fees antehandler. Cosmos moved a lot of stuff around and as I've got things now, it looks like things are sometimes being charged twice. I'll have to really dig in and see how our stuff meshes with the new stuff and what changes need to be made.

SpicyLemon:
In our fork of [wasmd](https://github.com/provenance-io/wasmd), I copied the draft PR branch from https://github.com/CosmWasm/wasmd/pull/930 and created branch `prov/v0.28.0-w-sdk-0.46.0` from it, and tagged it as `v0.28.0-pio-1-rc1`. That will let us at least move forward without everyone needed a local copy of that.

SpicyLemon:
Needed to inject history and update swagger. So `rc4` was cut.

kwtalley:
There was a bug in the wasmd version where contract instantiation would cause an overflow with gas calculation. There's a version from notional-labs that is being used by the community since 0.46 sdk isn't officially supported

source: https://github.com/CosmWasm/wasmd/pull/938
project status: https://github.com/CosmWasm/wasmd/issues/941

Issue #1015, opened 2022-08-16
Upgrade wasmd to an official release


SpicyLemon:
## Summary

Update `wasmd` to an official release that supports Cosmos-SDK v0.46.

## Problem Definition

When bumping Cosmos-SDK to v0.46, we had to tag `v0.28.0-pio-1-rc1` in our own fork. It's a copy of a draft PR to enable support for v0.46 in wasmd. But it's pulled from just a draft PR. We want something more reviewed/accepted before releasing our own new version of provenance.

## Proposal

Once [wasmd](github.com/CosmWasm/wasmd) releases a new version that's compatible with Cosmos v0.46, bump us to that new version.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

kwtalley:
wasmd is not likely to upgrade to cosmos-sdk v0.46 until 0.30.

0.29 is sitll in RC and went to the latest cosmos-sdk 0.45.x so this issue should be moved to backlog for hopefully 1.14 release

iramiller:
How are we going to release a 0.46 based provenanced without wasmd support? We might have to make our own release with this support similar to 0.42 days...

SpicyLemon:
> How are we going to release a 0.46 based provenanced without wasmd support? We might have to make our own release with this support similar to 0.42 days...

That's what we're currently doing. From `go.mod`:
```
replace github.com/CosmWasm/wasmd => github.com/provenance-io/wasmd v0.28.0-pio-1-rc1
```

That version was created from this draft PR: https://github.com/CosmWasm/wasmd/pull/930

kwtalley:
There was a bug in the wasmd version where contract instantiation would cause an overflow with gas calculation. There's a version from notional-labs that is being used by the community since 0.46 sdk isn't officially supported

source: https://github.com/CosmWasm/wasmd/pull/938
project status: https://github.com/CosmWasm/wasmd/issues/941

Issue #1008, opened 2022-08-15
Upgrade IBC to v5


SpicyLemon:
## Summary

Follow the upgrade documentation to get us to IBC v5.

## Problem Definition

As part of the update to v0.46.0, `go.mod` was updated to pull in IBC v5.0.0-beta1 (from v2.3.0). Compilation errors were fixed, but no migration steps were taken.

## Proposal

Follow the migration guides to get us from v2 to v5:
* v2 to v3: https://github.com/cosmos/ibc-go/blob/v5.0.0-beta1/docs/migrations/v2-to-v3.md
* v3 to v4: https://github.com/cosmos/ibc-go/blob/v5.0.0-beta1/docs/migrations/v3-to-v4.md
* v4 to v5: No documentation yet exists, so you'll need to dig into the repo a bit more.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1016, opened 2022-08-16
Bump Cosmos-SDK to v0.46.1-pio-1-rc1 (from v0.45.5-pio-1).

SpicyLemon:
## Description

This PR bumps Cosmos-SDK to `v0.46.1-pio-1-rc1` and gets it to the point where it compiles and tests pass. However, it should be pointed out that the fee decorator stuff is NOT working correctly, and the tests have been deactivated; #1006 is for fixing that.

Other changes of note in this PR:
* Added a `ochre-rc1` upgrade handler.
* Bump IBC to `v5.0.0-beta1` (from `v2.3.0`). `v5.0.0-beta1` is the only one that's compatible with Cosmos `v0.46` (at least right now). I got it to compile, but there's still migration work to do. Followup: #1008 
* Bump `wasmd` to `v0.28.0-pio-1-rc1`, which is a branch that is a draft PR to the main `wasmd` repo that I pulled into our fork and tagged. Followup: #1015
* Refactor the `x/marker` `Holding` query to utilize the SDK's new `x/bank` `DenomOwners` query. They added an index so it should be a pretty snappy (usable) query now.
* Removed the config index-events workaround.
* Removed the custom/legacy rest endpoints from `x/attribute`, `x/marker`, and `x/name` since the SDK removed all that rest support stuff.
* Removed all the state-listening/kafka stuff.
* Refactored some of the test setup stuff to resemble changes to similar stuff in the SDK.
* Replaced use of a few deprecated things (that were simple changes).
* Replaced use of deprecated `sdkResponse.Data` field since it's no longer populated, causing tests/sims to fail.
* Replaced use of deprecated `sdk.Int` with the new `cosmossdk.io/math` library.

closes: #995
closes: #1010

Note: The proto checking github actions fail for this PR, and will continue to fail for any PR after this until #1017 is merged in. It's being done separately since it's a large change, and this PR is already very large.

---

write a little note why.


x/reward/types/action_builder.go
```diff
@@ -228,6 +230,13 @@ func (v *VoteActionBuilder) GetEventCriteria() *EventCriteria {
 	})
 }
 
+var govVoteMsgUrls = []string{
```
Taztingo, line 233: I recommend moving govVoteMsgUrls into the AddEvent method since nothing else uses it.

app/test_helpers.go
```diff
@@ -88,20 +121,44 @@ func Setup(isCheckTx bool) *App {
...
+) GenesisState {
 	// set genesis accounts
 	authGenesis := authtypes.NewGenesisState(authtypes.DefaultParams(), genAccs)
 	genesisState[authtypes.ModuleName] = app.AppCodec().MustMarshalJSON(authGenesis)
 
 	validators := make([]stakingtypes.Validator, 0, len(valSet.Validators))
 	delegations := make([]stakingtypes.Delegation, 0, len(valSet.Validators))
 
-	bondAmt := sdk.NewInt(1000000)
+	bondAmt := sdk.DefaultPowerReduction
```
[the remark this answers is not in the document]
SpicyLemon: This test stuff was mostly copied out of the SDK and so it still uses `sdk.DefaultBondDenom` (`stake`) as the denomination.

app/prefix_test.go
```diff
@@ -0,0 +1,59 @@
...
+			"m/44'/118'/0'/0/0",
+			true,
+			false,
+			false,
+			"",
+		},
+		{
+			"has correct bip44th path for mainnet",
+			"m/44'/118'/0'/0/0",
+			true,
```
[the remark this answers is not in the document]
SpicyLemon: Good catch.

Looks like these tests weren't actually doing anything. `118` is the `atom` coin type. We want it to be `505` for mainnet and `1` for testnet. The test was erroneously just getting the BIP44 of the default config each time rather than the config being set by our `SetConfig`.

I've corrected these tests.

iramiller (approved):

x/reward/types/action_builder.go
SpicyLemon: Having it in `AddEvent` means it has to be recreated every single time it's needed. Those calls to `sdk.MsgTypeURL` aren't just simple look-ups. But it probably is better to set it the first time `AddEvent` is called because of `MsgTypeURL`'s possible reliance on reflection and/or proto registration. So I left the variable scoped to the package, but set it if it hasn't been set yet.

Taztingo: I like the new approach, and it makes sense from a performance perspective.

iramiller (approved):

iramiller (approved):

app/app.go
```diff
@@ -707,7 +719,7 @@ func New(
 		capability.NewAppModule(appCodec, *app.CapabilityKeeper),
 		feegrantmodule.NewAppModule(appCodec, app.AccountKeeper, app.BankKeeper, app.FeeGrantKeeper, app.interfaceRegistry),
 		gov.NewAppModule(appCodec, app.GovKeeper, app.AccountKeeper, app.BankKeeper),
-		mint.NewAppModule(appCodec, app.MintKeeper, app.AccountKeeper),
+		mint.NewAppModule(appCodec, app.MintKeeper, app.AccountKeeper, nil),
```
arnabmitra, line 722: is there any specific reason why in `mint.NewAppModule` the BankKeeper is being passed in as `nil`, probably good to document it so that no one passes in a non nil BankKeeper later if there is a reason.

sims.mk
```diff
@@ -7,7 +7,8 @@
 ### The strings in there leave off the "test-sim" prefix.
 ###
 ### Environment Variables:
```
Taztingo, line 9: Should we also add SIMAPP?

app/app.go
SpicyLemon: That `nil` is actually an `InflationCalculationFn`:

`type InflationCalculationFn func(ctx Context, minter Minter, params Params, bondedRatio Dec) Dec`
> InflationCalculationFn defines the function required to calculate inflation rate during BeginBlock. It receives the minter and params stored in the keeper, along with the current bondedRatio and returns the newly calculated inflation rate. It can be used to specify a custom inflation calculation logic, instead of relying on the default logic provided by the sdk.

sims.mk
SpicyLemon: The `SIMAPP` variable is defined with a `=` (not a `?=`), so it's not an environment variable that can be defined outside the makefile.

arnabmitra (approved):

internal/antewrapper/tx_gas_limit_decorator.go
```diff
@@ -17,13 +18,39 @@ func NewTxGasLimitDecorator() TxGasLimitDecorator {
 	return TxGasLimitDecorator{}
 }
 
+// govMsgURLs the MsgURLs of all the governance module's messages.
+// Use getGovMsgURLs() instead of using this variable directly.
+var govMsgURLs []string
+
+// getGovMsgURLs returns govVoteMsgUrls, but first sets it if it hasn't yet been set.
+func getGovMsgURLs() []string {
+	if len(govMsgURLs) != 8 {
```
Taztingo, line 27: This is really minor, but would it be safer to reverse your condition to govMsgURLs == nil? That makes it a little bit easier to read, and in case the contents of govMsgURLs ever change then this doesn't have to also be updated.

x/reward/types/action_builder.go
```diff
@@ -228,6 +230,26 @@ func (v *VoteActionBuilder) GetEventCriteria() *EventCriteria {
 	})
 }
 
+// govVoteMsgURLs the MsgURLs of all the governance module's vote messages.
+// Use getGovVoteMsgURLs() instead of using this variable directly.
+var govVoteMsgURLs []string
+
+// getGovVoteMsgURLs returns govVoteMsgURLs, but first sets it if it hasn't yet been set.
+func getGovVoteMsgURLs() []string {
+	if len(govVoteMsgURLs) != 4 {
```
Taztingo, line 239: Same comment as what I left before. Is it possible to switch the condition to govVoteMsgURLs == nil?

internal/antewrapper/tx_gas_limit_decorator.go
SpicyLemon: Yeah... I waffled on that and even messed it up the first time because I had `4` there instead of `8`.

Both in here and in the x/reward action_builder, I changed the test to ` == nil` and put a comment over it about it.

x/reward/types/action_builder.go
SpicyLemon: Yup. Just changed it.

Taztingo (approved):
LGTM. Nice job!

iramiller (approved):

Issue #1099, opened 2022-09-28
Gov prop gives consensus failure when voting ends.


SpicyLemon:
## Summary of Bug

When the voting period ends on a governance proposal (e.g. an upgrade propsoal), the chain halts with: `ERR CONSENSUS FAILURE!!! err="GasMeter is not of type FeeGasMeter"`.

## Version

`dwedul/1006-fix-msg-fees` (in a state that shouldn't need any changes before being merged into `main`), commit `0baff2f3`.

## Steps to Reproduce

1. `git checkout dwedul/1006-fix-msg-fees`
2. If #1097 isn't merged yet: `git checkout dwedul/add-upgrade-test-script scripts/upgrade-test.sh`
3. `go mod tidy && go mod vendor`
4. `make clean build run-config`
5. `./scripts/upgrade-config.sh`
6. `make run`
7. In another terminal: `scripts/upgrade-test.sh brown`
8. Back in the first terminal, wait for voting to end on the newly submitted upgrade proposal.

Result (formatting changed to make it easier to read):
```
ERR CONSENSUS FAILURE!!!
err="GasMeter is not of type FeeGasMeter"
module=consensus
stack=
    goroutine 119 [running]:
    runtime/debug.Stack()
      runtime/debug/stack.go:24 +0x64
    github.com/tendermint/tendermint/consensus.(*State).receiveRoutine.func2()
      github.com/tendermint/tendermint@v0.34.21/consensus/state.go:727 +0x44
    panic({0x105251960, 0x1055ffb00})
      runtime/panic.go:884 +0x204
    github.com/provenance-io/provenance/internal/handlers.(*PioMsgServiceRouter).RegisterService.func2({{0x105626830, 0x140000540e8}, {0x105639f40, 0x14000d55680}, {{0xb, 0x0}, {0x140056bca67, 0x7}, 0xf0, {0x183e1da0, ...}, ...}, ...}, ...)
      github.com/provenance-io/provenance/internal/handlers/msg_service_router.go:129 +0xbd8
    github.com/cosmos/cosmos-sdk/x/gov.EndBlocker.func2({0x1, {0x1400131b948, 0x1, 0x1}, 0x2, 0x14000d54ac0, 0x140016af368, 0x140016af3b0, {0x140016af3c8, 0x1, ...}, ...})
      github.com/cosmos/cosmos-sdk@v0.46.0/x/gov/abci.go:73 +0xca0
[...]

PR #1097, opened 2022-09-27
Add a script to make testing upgrades easier.

SpicyLemon:
## Description

Add a script to make testing upgrades easier.
Update the documentation to direct people to the script instead of the several steps the script takes care of.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #1102, opened 2022-09-29
Fix gov proposal end-of-voting-period consensus failure

SpicyLemon:
## Description

In the begin and end blockers, wrap the gas meter as a fee gas meter if it isn't already a fee gas meter.

In the new gov stuff, when voting ends on a gov proposal, the end block triggers handles the sending of successfully voted-on proposals, and does so through the message service router. Our custom message service router expects the gas meter to be a fee gas meter though. So, in order for the new gov stuff to have the right gas meter, it needs to be wrapped in our end block so that it's wrapped in the gov module's end block which can then call our message service router without blowing up.

closes: #1099

Note: I put together this PR without any integration tests on it because writing such an integration test is going to take quite a bit longer. I am working on it though.

I did test this by `make clean build run-config && ./scripts/upgrade-config.sh && make run`, then in another window, I ran the script from #1097 with the color "brown". Before this PR, the chain halted at the end of the voting period (with the not-a-fee-gas-meter error). After this PR, it got past the end of the voting period and instead (expectedly) halted at the upgrade height.

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Bug Fixes
 
 * Remove the workaround for the index-events configuration field (now fixed in the SDK). [#995](https://github.com/provenance-io/provenance/issues/995)
+* Fix consensus failure at the end of gov voting periods. [#1099](https://github.com/provenance-io/provenance/issues/1099)
 
 ### Client Breaking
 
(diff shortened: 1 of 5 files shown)


SpicyLemon:
I'm going to cancel this PR and do it a different way.

After discussions in our channel, I (we) think it's better to bypass the fee consumption (in the message service router) if the gas meter is not a fee gas meter. If it gets there without the gas meter being a fee gas meter, just skip the msg fee consumption.

Issue #1092, opened 2022-09-26
Make the `fee` event correct before v1.13 goes to prod


arnabmitra:
## Summary of Bug

Make the `fee` event correct before v1.13 goes to prod.
Currently 
```
events := sdk.Events{sdk.NewEvent(sdk.EventTypeTx,
		sdk.NewAttribute(sdk.AttributeKeyFee, feeTx.GetFee().String()),
	),
	
```
Which is correct in case of `success`

But in failure case, this is not right because the fee charged is only the 
```
sdk.NewEvent(sdk.EventTypeTx,
			sdk.NewAttribute(AttributeKeyMinFeeCharged, baseFeeToConsume.String()),
		),
```
so proposed fix is to set this value to ` baseFeeToConsume.String(` initially but to swap it out to `total fee charged` value
if the `runMsgs` is a `success`, will require changes to `baseapp.go` to do that.

We also need to emit the suppressed sdk `SendCoin` events from the FeeInvoker distribution of funds.  These include the rollup of fees to the FeeModule and Recipients of msgfees and event for the Swept Fee amount.  

## Version
v.1.12 (current production)

## Steps to Reproduce
Send a tx which will fail, fee charged will be only the `min_fee_charged` event but `fee` event will be whatever is sent in.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1107, opened 2022-09-30
Update fee event to contain total charge and emit send coin events

nullpointer0x00:
## Description

Versions <= v1.12 have a fee event that is dispatched at the beginning of the transaction that is unreliable.  This is due to the fact that it is emitting the total amount of fee that is sent with the transaction.  The case where this fails to be accurate is when the transaction fails at the ante handler level.  This is because only gas_wanted * min_gas_cost is charged.
  
This PR also dispatches the events from the distribution of msg based fees and sweeping of the remained fees.  These where suppressed in versions <= 1.12
 
closes: #1092

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,11 +54,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
   * Refactored the `ProvenanceDeductFeeDecorator`. It now makes sure the payer has enough in their account to cover the additional fees. It also now deducts/consumes the `floor gas price * gas`.
   * Added the `fee_payer` attribute to events of type `tx` involving fees (i.e. the ones with attributes `fee`, `min_fee_charged`, `additionalfee` and/or `baseFee`).
   * Moved the additional fees calculation logic into the msgfees keeper.
-<<<<<<< HEAD
-  * Update `fee` event with amount charged even on failure and emit SendCoin events from `DeductFeesDistributions` [#1092](https://github.com/provenance-io/provenance/issues/1092)
-=======
+* Update `fee` event with amount charged even on failure and emit SendCoin events from `DeductFeesDistributions` [#1092](https://github.com/provenance-io/provenance/issues/1092)
 * Bump IBC to `5.0.0-pio-1` (from `v2.3.0`) to add a check for SendEnabled [#1100](https://github.com/provenance-io/provenance/issues/1100)
->>>>>>> main
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
   * Refactored the `ProvenanceDeductFeeDecorator`. It now makes sure the payer has enough in their account to cover the additional fees. It also now deducts/consumes the `floor gas price * gas`.
   * Added the `fee_payer` attribute to events of type `tx` involving fees (i.e. the ones with attributes `fee`, `min_fee_charged`, `additionalfee` and/or `baseFee`).
   * Moved the additional fees calculation logic into the msgfees keeper.
+  * Update `fee` event with amount charged even on failure and emit SendCoin events from `DeductFeesDistributions` [#1092](https://github.com/provenance-io/provenance/issues/1092)
 
 ### Bug Fixes
 
--- a/internal/handlers/aggregate_events.go
+++ b/internal/handlers/aggregate_events.go
@@ -1,0 +1,40 @@
+package handlers
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+// AggregateEvents is called in the baseapp after a transaction has completed
+// This is used to modify the events that are emitted for a transaction.
+// anteEvents will be populated on failure and success
+// resultEvents will only be populated on success
+func AggregateEvents(anteEvents []abci.Event, resultEvents []abci.Event) ([]abci.Event, []abci.Event, error) {
+	if len(resultEvents) == 0 { // tx failed...fix fee event to have the exact fee charged
+		var err error
+		var txFee sdk.Coins
+		var feeIndex int
+		var feeFound, spenderFound bool
+		for i, event := range anteEvents {
+			if !feeFound && event.Type == sdk.EventTypeTx && string(event.Attributes[0].Key) == sdk.AttributeKeyFee {
+				feeFound = true
+				feeIndex = i
+			}
+			// first spent coin event is the coin sent to fee module for tx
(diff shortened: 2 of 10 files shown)


internal/handlers/aggregate_events.go
```diff
@@ -0,0 +1,40 @@
...
+		var feeIndex int
+		var feeFound, spenderFound bool
+		for i, event := range anteEvents {
+			if !feeFound && event.Type == sdk.EventTypeTx && string(event.Attributes[0].Key) == sdk.AttributeKeyFee {
+				feeFound = true
+				feeIndex = i
+			}
+			// first spent coin event is the coin sent to fee module for tx
+			if !spenderFound && event.Type == banktypes.EventTypeCoinSpent && string(event.Attributes[0].Key) == banktypes.AttributeKeySpender && len(anteEvents) >= i+3 {
+				txFee, err = sdk.ParseCoinsNormalized(string(event.Attributes[1].Value))
```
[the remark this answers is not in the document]
nullpointer0x00: That was an artifact from an old impl.  It is now removed.

internal/handlers/aggregate_events.go
```diff
@@ -0,0 +1,39 @@
...
+// This is used to modify the events that are emitted for a transaction.
+// anteEvents will be populated on failure and success
+// resultEvents will only be populated on success
+func AggregateEvents(anteEvents []abci.Event, resultEvents []abci.Event) ([]abci.Event, []abci.Event) {
+	if len(resultEvents) == 0 { // tx failed...fix fee event to have the exact fee charged
+		var txFee []byte
+		var feeIndex int
+		var feeFound, spenderFound bool
+		for i, event := range anteEvents {
+			if !feeFound && event.Type == sdk.EventTypeTx && string(event.Attributes[0].Key) == sdk.AttributeKeyFee {
```
SpicyLemon, line 20: I feel like hard-coding the indexes in here is a recipe for future problems (talking about all the spots in this file where that's done).

internal/handlers/aggregate_events.go
```diff
@@ -0,0 +1,39 @@
...
+				feeFound = true
+				feeIndex = i
+			}
+			// first spent coin event is the coin sent to fee module for tx
+			if !spenderFound && event.Type == banktypes.EventTypeCoinSpent && string(event.Attributes[0].Key) == banktypes.AttributeKeySpender {
+				txFee = event.Attributes[1].Value
+				spenderFound = true
+			}
```
SpicyLemon, line 28: Is there any reason not to use the `min_fee_charged` value? Relying on the first spender having the fee amount seems delicate.

internal/handlers/msg_fee_invoker.go
```diff
@@ -79,10 +79,12 @@ func (afd MsgFeeInvoker) Invoke(ctx sdk.Context, simulate bool) (sdk.Coins, sdk.
 
 		// If there's fees left to collect, or there were consumed fees, deduct/distribute them now.
 		if !unchargedFees.IsZero() || !consumedFees.IsZero() {
+			ctx = ctx.WithEventManager(sdk.NewEventManager())
```
SpicyLemon, line 82: The new context should be put into a new variable since we only want the fresh event manager for this if block.

x/msgfees/keeper/keeper.go
```diff
@@ -176,10 +176,12 @@ func (k Keeper) DeductFeesDistributions(bankKeeper bankkeeper.Keeper, ctx sdk.Co
 	if neg {
 		return sdkerrors.ErrInsufficientFunds.Wrapf("negative balance after sending coins to accounts and fee collector: remainingFees: %q, sentCoins: %q, distribution: %v", remainingFees, sentCoins, fees)
 	}
-	// sweep the rest of the fees to module
-	err := bankKeeper.SendCoinsFromAccountToModule(ctx, acc.GetAddress(), k.feeCollectorName, unsentFee)
-	if err != nil {
-		return sdkerrors.ErrInsufficientFunds.Wrap(err.Error())
+	if unsentFee.IsAllPositive() {
```
SpicyLemon, line 179: Long story short, you should use `!unsentFee.IsZero()` for this test.

Long story:

Reading this, `unsentFee.IsAllPositive()` doesn't look right since `0hotdog,5stake` would return false. In that case `!unsentFee.IsZero()` would be better.

I did a quick test though, and (at least right now), it looks like `SafeSub` doesn't include entries that end up being 0. So this ends up working. But since that's not obvious, and depends on external functionality, a comment about why `IsAllPositive` is used would be helpful for future maintainers.

Then, all that being said, if `SafeSub` resulted in a negative amount, we've returned an error. Combine that with knowing that `SafeSub` doesn't return entry with `0` amounts. What you're left with is that either `unsentfee` is zero (has nothing in it), or it has one or more positive amounts in it. So again, it'd just be better to use `! .IsZero`.

nullpointer0x00: Yeah, good call.   I originally had it as that.

internal/handlers/aggregate_events.go
nullpointer0x00: I was thinking of using that initially, but I was under the impression that we might remove that at some point.  I also thought that the first spender seemed like a delicate thing to do...
I did review the anteevents and currently the first spender will only be this event.  However, I just realized we could add another ante handler earlier in the chain that could add an event before this one and it would break.  
I will use use the `min_fee_charged` event for now.

internal/handlers/aggregate_events.go
nullpointer0x00: I agree.

internal/handlers/msg_fee_invoker.go
nullpointer0x00: Done.

SpicyLemon (approved):
Looks good. Just need to fix conflicts.

CHANGELOG.md
```diff
@@ -54,7 +54,11 @@ Ref: https://keepachangelog.com/en/1.0.0/
   * Refactored the `ProvenanceDeductFeeDecorator`. It now makes sure the payer has enough in their account to cover the additional fees. It also now deducts/consumes the `floor gas price * gas`.
   * Added the `fee_payer` attribute to events of type `tx` involving fees (i.e. the ones with attributes `fee`, `min_fee_charged`, `additionalfee` and/or `baseFee`).
   * Moved the additional fees calculation logic into the msgfees keeper.
+<<<<<<< HEAD
+  * Update `fee` event with amount charged even on failure and emit SendCoin events from `DeductFeesDistributions` [#1092](https://github.com/provenance-io/provenance/issues/1092)
+=======
 * Bump IBC to `5.0.0-pio-1` (from `v2.3.0`) to add a check for SendEnabled [#1100](https://github.com/provenance-io/provenance/issues/1100)
+>>>>>>> main
```
SpicyLemon, line 61: Oops

SpicyLemon (approved):

CHANGELOG.md
nullpointer0x00: Oops is right... Fixed.

iramiller (approved):

Issue #957, opened 2022-07-20
Expiration Module Epic


derekadams:
## Summary

This epic adds the concept of asset expiration to Provenance. When assets are added to the chain, expiration metadata is added so that the assets may be removed at a future date, allowing the head state of the chain to be minimized. The expiration process is designed to recover gas costs of the expiration operations and not require any per-block processing overhead in handling asset expiration.

## Problem Definition

When storing assets on-chain, there is an expense in terms of node memory and processing that is associated with each added asset that is maintained in the head state. Without some method of expiration, the head state of the system would continue growing until it eventually becomes too large to be efficiently managed and system performance would degrade.

In many cases, asset lifespan may be estimated at creation so that the asset may be expired and potentially removed from active state. The expiration process uses node processing resources to prune the state, so the solution should offset gas costs if possible. Also, since a large number of assets may be monitored for expiration over a long lifespan, care must be taken to prevent expiration processing from adding too much overhead to the default processing loop for nodes.

## Proposal

* [x] #961 
* [x] #1056 
* [x] https://github.com/provenance-io/provwasm/issues/93
* [x] #1063
* [x] #1075

[Child issues to be added here to support implementation]

## Supporting Documentation
[Draft ADR](https://hackmd.io/G3duqVfnQnysYgrfq5AeZA)
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

egaxhaj:
@derekadams We have a duplicate epic created by @iramiller #689 with subtasks that we need to fold into one epic.

Issue #1075, opened 2022-09-16
Add spec documentation for expiration module


egaxhaj:
## Summary

Update `x/expiration/spec/*.md` files to include documentation

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

egaxhaj:
fixed by #1153

PR #1115, opened 2022-10-03
Expiration module

derekadams:
## Description

closes: #957

---

write a little note why.


proto/provenance/expiration/v1/query.proto
```diff
@@ -0,0 +1,101 @@
+// Since: provenance 1.12
+syntax = "proto3";
+package provenance.expiration.v1;
+
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "provenance/expiration/v1/expiration.proto";
+// import "cosmos/base/v1beta1/coin.proto";
```
arnabmitra, line 8: remove comment

Taztingo:
Are the specs supposed to be in this, or are we adding them at a later time? It says they are all empty files.

x/name/legacy/v042/types.go
```diff
@@ -13,7 +13,7 @@ var (
 	NameAddressLengthLegacy = 20
 )
 
-// GetAddressKeyPrefix returns a store key for a name record address
+// GetOwnerKeyPrefix returns a store key for a name record address
```
Taztingo, line 16: I think this was changed by mistake.

proto/provenance/expiration/v1/query.proto
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // Query Expirations by owner
+  rpc AllExpirationsByOwner(QueryAllExpirationsByOwnerRequest) returns (QueryAllExpirationsByOwnerResponse) {
+    option (google.api.http).get = "/provenance/expiration/v1/all/{owner}";
+  }
+
+  // Query all expired Expirations
+  rpc AllExpiredExpirations(QueryAllExpiredExpirationsRequest) returns (QueryAllExpiredExpirationsResponse) {
+    option (google.api.http).get = "/provenance/expiration/v1/all/expired";
```
SpicyLemon, line 39: You might need to rename this so it doesn't conflict with the `AllExpirationsByOwner` endpoint. I'm guessing that it will think that `expired` is a just an `{owner}` value.

SpicyLemon, line 39: Have you checked that all of these REST endpoints work? I have a feeling it will be confused and think `params` and `all` are just `{module_asset_id}` values, and `expired` is an `{owner}` value. Sometimes things work better than I expect though.

proto/provenance/expiration/v1/query.proto
```diff
@@ -0,0 +1,101 @@
...
+
+// QueryExpirationResponse response for querying for a specific expiration
+message QueryExpirationResponse {
+  Expiration expiration = 1;
+}
+
+// QueryAllExpirationsRequest queries all expirations
+message QueryAllExpirationsRequest {
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
```
SpicyLemon, line 65: Please change all the `pagination` indexes in here to `99`. Doing so keeps them at the end in the future, when something needs to be added to either a request or response.

proto/provenance/expiration/v1/tx.proto
```diff
@@ -0,0 +1,67 @@
...
+message MsgExtendExpirationResponse {}
+
+// MsgInvokeExpirationRequest represents a message to invoke expiration logic for an asset
+message MsgInvokeExpirationRequest {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.stringer)        = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // Module asset ID which will have its expiration logic triggered
+  string module_asset_id = 1 [(gogoproto.moretags) = "yaml:\"module_asset_id\""];
```
SpicyLemon, line 61: We don't need the `yaml` tags anymore.

proto/provenance/expiration/v1/query.proto
```diff
@@ -0,0 +1,101 @@
...
+
+// QueryParamsResponse is the response type for the Query/Params RPC method.
+message QueryParamsResponse {
+  // params defines the parameters of the module.
+  Params params = 1 [(gogoproto.nullable) = false];
+}
+
+// QueryExpirationRequest queries for a specific expiration
+message QueryExpirationRequest {
+  string module_asset_id = 1 [(gogoproto.moretags) = "yaml:\"module_asset_id\""];
```
SpicyLemon, line 54: Don't need the `yaml` tag.

proto/provenance/metadata/v1/tx.proto
```diff
@@ -111,6 +113,10 @@ message MsgWriteScopeRequest {
   // If there is a value in scope.specification_id that is different from the one created from this uuid, an error is
   // returned.
   string spec_uuid = 4 [(gogoproto.moretags) = "yaml:\"spec_uuid\""];
+
+  // expiration is an optional expiration timespan string, e.g. "1y" (see expiration module for format)
+  // If provided, it overrides the default scope expiration with the given time period.
+  string expiration = 5 [(gogoproto.moretags) = "yaml:\"expiration\""];
```
SpicyLemon, line 119: I thought we were keeping the expiration stuff separate from the things being expired.

proto/provenance/metadata/v1/tx.proto
```diff
@@ -21,6 +21,8 @@ service Msg {
   rpc WriteScope(MsgWriteScopeRequest) returns (MsgWriteScopeResponse);
   // DeleteScope deletes a scope and all associated Records, Sessions.
   rpc DeleteScope(MsgDeleteScopeRequest) returns (MsgDeleteScopeResponse);
+  // ExpireScope expires a scope and all associated Records, Sessions.
+  rpc ExpireScope(MsgExpireScopeRequest) returns (MsgExpireScopeResponse);
```
SpicyLemon, line 25: What's the difference between `DeleteScope` and `ExpireScope`? Why do we need an `ExpireScope`?

x/expiration/client/cli/cli_test.go
```diff
@@ -0,0 +1,828 @@
...
+	s.user2Addr, s.acctErr = s.keyringAccounts[2].GetAddress()
+	s.Require().NoError(s.acctErr, "getting keyringAccounts[2] address")
+	s.user2AddrStr = s.user2Addr.String()
+
+	// A third user account
+	s.user3Addr, s.acctErr = s.keyringAccounts[3].GetAddress()
+	s.Require().NoError(s.acctErr, "getting keyringAccounts[3] address")
+	s.user3AddrStr = s.user3Addr.String()
+
+	// A third user account
```
SpicyLemon, line 147: Probably just remove these comments. Or if keeping them, update the rest to the correct numbers.

x/expiration/client/cli/query.go
```diff
@@ -0,0 +1,233 @@
...
+		Aliases: []string{"a"},
+		Short:   "Query all expirations",
+		Args:    cobra.NoArgs,
+		Example: strings.TrimSpace(
+			fmt.Sprintf(`
+				$ %[1]s query expiration all
+				$ %[1]s query expiration all --page=2 --limit=100
+				`,
+				version.AppName,
+			)),
```
SpicyLemon, line 111: Because this string is inside tics, all characters are included. That means that the initial newline after the first tic is in the string as well as the 4 leading tabs on each line.

```suggestion
			fmt.Sprintf(`$ %[1]s query expiration all
$ %[1]s query expiration all --page=2 --limit=100`,
				version.AppName,
			)),
```

Also, maybe consider creating a variable to hold the string `version.AppName + " query expiration"`, and providing that in all these commands instead of `version.AppName` and repeating those next two pieces each time.

x/expiration/client/cli/query.go
```diff
@@ -0,0 +1,233 @@
...
+		Aliases: []string{"o"},
+		Short:   "Query all expirations by owner",
+		Args:    cobra.ExactArgs(1),
+		Example: strings.TrimSpace(
+			fmt.Sprintf(`
+				$ %[1]s query expiration owner pb1skjwj5whet0lpe65qaq4rpq03hjxlwd9nf39lk
+				$ %[1]s query expiration owner pb1skjwj5whet0lpe65qaq4rpq03hjxlwd9nf39lk --page=2 --limit=100
+				`,
+				version.AppName,
+			)),
```
SpicyLemon, line 152: Here again with the spacing.

x/expiration/client/cli/query.go
```diff
@@ -0,0 +1,233 @@
...
+		Aliases: []string{"e"},
+		Short:   "Query all expired expirations",
+		Args:    cobra.NoArgs,
+		Example: strings.TrimSpace(
+			fmt.Sprintf(`
+				$ %[1]s query expiration expired
+				$ %[1]s query expiration expired --page=2 --limit=100
+				`,
+				version.AppName,
+			)),
```
SpicyLemon, line 193: Spacing here too.

x/expiration/client/cli/tx.go
```diff
@@ -0,0 +1,138 @@
...
+		Aliases: []string{"e"},
+		Short:   "Extend expiration metadata for an asset on the provenance blockchain",
+		Long: `Extend expiration metadata for an asset on the provenance blockchain
+
+Example:
+$ %s tx expiration extend pb1skjwj5whet0lpe65qaq4rpq03hjxlwd9nf39lk duration 1y
+
+module-asset-id   - module asset address
+duration          - the duration period for which the module asset will continue to remain on chain.
+					Valid time units are "y", "w", "d", "h"`,
```
SpicyLemon, line 52: This line is probably spaced in too much. Also, if you want it to line up nicely with the beginning of the line above it, you'll need to use spaces (not tabs).

SpicyLemon (commented):
Just getting started on reviewing this (11/56 files down). I need to stop for a bit and wanted to get my initial comments in now rather than waiting to have all of them at once.

x/expiration/client/cli/query.go
```diff
@@ -0,0 +1,233 @@
...
+		Short:                      "Querying commands for the expiration module",
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+	queryCmd.AddCommand(
+		GetQueryParamsCmd(),
+		GetExpirationCmd(),
+		GetAllExpirationsCmd(),
+		GetAllExpirationsByOwnerCmd(),
```
Taztingo, line 33: You are missing GetAllExpiredExpirationsCmd.

x/expiration/client/cli/tx.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return cmd
+}
+
+// InvokeExpirationCmd creates a command for invoking expiration logic on an asset
+func InvokeExpirationCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "invoke [module-asset-id]",
+		Aliases: []string{"i"},
+		Short:   "Invoke expiration logic for an asset on the provenance blockchain",
```
Taztingo, line 88: I haven't looked at the code for Invoking yet, but one question that I have as a new user is "will invoke logic be called again when the timer expires?". Might be nice to clarify that either here or in the specs.

arnabmitra:
if you are adding a module, i believe needs an upgrade handler addition for new module also.

egaxhaj:
> > Are the specs supposed to be in this, or are we adding them at a later time? It says they are all empty files.
> 
> 
> 
> There's an outstanding issue (https://github.com/provenance-io/provenance/issues/1075) for adding those specs. I think we're ok to go to RC without them, but we definitely want them there before the release.

@arnabmitra

x/expiration/module/module.go
```diff
@@ -0,0 +1,173 @@
...
+func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, _ sdkclient.TxEncodingConfig, bz json.RawMessage) error {
+	var data types.GenesisState
+	if err := cdc.UnmarshalJSON(bz, &data); err != nil {
+		return sdkerrors.Wrapf(err, "failed to unmarshal %q genesis state", types.ModuleName)
+	}
+
+	return types.ValidateGenesis(data)
+}
+
+func (b AppModuleBasic) RegisterRESTRoutes(_ sdkclient.Context, _ *mux.Router) {}
```
SpicyLemon, line 75: This isn't used or needed anymore.

x/expiration/module/module.go
```diff
@@ -0,0 +1,173 @@
...
+	return cli.GetQueryCmd()
+}
+
+// AppModule implements the sdk.AppModule interface
+type AppModule struct {
+	AppModuleBasic
+
+	keeper      keeper.Keeper
+	authzKeeper authzkeeper.Keeper
+	// TODO we will want access to other keepers
```
SpicyLemon, line 97: Unaddressed TODO, but also, you don't need the `authzKeeper` in here.

x/expiration/simulation/genesis.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/provenance-io/provenance/x/expiration/types"
+)
+
+// RandomizedGenState generates a random GenesisState for marker
```
SpicyLemon, line 11: Not for a marker.

x/expiration/types/codec.go
```diff
@@ -0,0 +1,43 @@
...
+	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
+}
+
+// whiteListedInterfaceRegistry whitelist of sdk.Msg types
+// invoked when a module asset expiration has expired
+func whiteListedInterfaceRegistry() types.InterfaceRegistry {
+	ir := types.NewInterfaceRegistry()
+	ir.RegisterImplementations((*sdk.Msg)(nil),
+		&metadatatypes.MsgDeleteScopeRequest{},
+		&metadatatypes.MsgDeleteRecordRequest{},
```
SpicyLemon, line 29: I'd like the expiration module to not have anything referencing modules that will be using it. It'd be pretty easy to end up with circular references. Maybe that's something to address later though.

x/expiration/types/codec.go
```diff
@@ -0,0 +1,43 @@
...
+		// todo: what other messages do we need to whitelist?
```
SpicyLemon, line 30: Unaddressed TODO.

x/expiration/types/expiration.go
```diff
@@ -0,0 +1,109 @@
...
+		Owner:         owner,
+		Time:          time,
+		Deposit:       deposit,
+		Message:       message,
+	}
+}
+
+// ValidateBasic basic format checking of the data
+func (e *Expiration) ValidateBasic() error {
+	now := time.Now()
```
SpicyLemon, line 32: Nope. Nope. NoNoNoNoNoNo.

You cannot use `time.Now()` anywhere, ever.

This `ValidateBasic` function simply cannot do any checking on the expiration time other than maybe making sure it's not nil or 0. That validation will need to be done elsewhere where the block time can be accessed.

x/expiration/types/expiration.go
```diff
@@ -0,0 +1,109 @@
...
+		return fmt.Errorf("expecting non nil value to validate an Any message")
+	}
+	if e.Message.TypeUrl == "" {
+		return fmt.Errorf("expecting message type URL to unpack Any")
+	}
+	// validate message is a whitelisted sdk.Msg during a BroadcastTx
+	var msg sdk.Msg
+	if err := ModuleCdc.UnpackAny(&e.Message, &msg); err != nil {
+		return err
+	}
```
SpicyLemon, line 68: Is there even a reason we need to whitelist message types?

x/expiration/types/expiration.go
```diff
@@ -0,0 +1,109 @@
...
+	var duration time.Duration
+	switch period {
+	case "y":
+		duration = time.Duration(digit) * 24 * 365 * time.Hour
+	case "w":
+		duration = time.Duration(digit) * 24 * 7 * time.Hour
+	case "d":
+		duration = time.Duration(digit) * 24 * time.Hour
+	case "h":
+		duration = time.Duration(digit) * time.Hour
```
SpicyLemon, line 102: I don't think I saw anywhere documenting the duration strings. It should probably be documented that `1d` = `24h`, `1w` = `7d` = `168h`, `1y` = `365d` = `8760h`. There are several definitions of the word "year", and even "day". Also, that it's not like a `time.Duration` in that you can only provide one unit, e.g. you can't do `1y5w`.

x/expiration/types/msgs.go
```diff
@@ -0,0 +1,161 @@
...
+// Compile time interface checks
+var (
+	_ sdk.Msg = &MsgAddExpirationRequest{}
+	_ sdk.Msg = &MsgExtendExpirationRequest{}
+	_ sdk.Msg = &MsgInvokeExpirationRequest{}
+)
+
+// MustAccAddressFromBech32 converts a Bech32 address to sdk.AccAddress
+// Panics on error
+func MustAccAddressFromBech32(s string) sdk.AccAddress {
```
SpicyLemon, line 20: The `sdk` now has this function: `sdk.MustAccAddressFromBech32`.

x/expiration/types/msgs.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// private method that checks an address string is bech32 or MetadataAddress type
+func validateAddress(s string) error {
+	if _, err := sdk.AccAddressFromBech32(s); err != nil {
+		// check if we're dealing with a MetadataAddress
+		if _, err2 := metadatatypes.MetadataAddressFromBech32(s); err2 != nil {
+			return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress,
+				fmt.Sprintf("invalid module asset id: %s", err.Error()))
+		}
```
SpicyLemon, line 56: Again, `sdk.AccAddressFromBech32(s)` will return an error if given a metadata address.

x/expiration/types/params.go
```diff
@@ -0,0 +1,98 @@
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
+
+const (
+	NhashDenom = "nhash"
+)
+
+// DefaultDuration is the default duration a module asset
+// will live on chain before expiring. Defaults to 1 year
+var DefaultDuration = 24 * 365 * time.Hour
```
SpicyLemon, line 19: This isn't use anywhere.

x/expiration/types/params.go
```diff
@@ -0,0 +1,98 @@
...
+
+var DefaultDeposit = sdk.Coin{
+	Denom: NhashDenom,
+	// todo: what should this be? should this be a configurable property in app.toml?
+	Amount: sdk.NewInt(1905),
+}
```
SpicyLemon, line 25: 1) Should probably use `sdk.NewIng64Coin(NhashDenom, 1905)`
2) It should probably be something other than `1905` since that's the gas price, and this has nothing to do with that, and is probably way too low.
3) Unaddressed TODO. And Yeah. The denom should come from the config. We'll need to discuss the amount, but I feel like it'd be closer to `1000000000nhash` (1 hash). Alternatively, we could just define it as `0nhash` here so that it's only meaningful if set in params.

x/expiration/types/params.go
```diff
@@ -0,0 +1,98 @@
...
+			that1 = &that2
+		} else {
+			return false
+		}
+	}
+	if that1 == nil {
+		return p == nil
+	} else if p == nil {
+		return false
+	}
```
SpicyLemon, line 78: You've already handled the case where one is nil and the other isn't. Here, you should be actually checking for equality in the values.

x/expiration/types/params.go
```diff
@@ -0,0 +1,98 @@
...
+func validateDepositParam(i interface{}) error {
+	coin, ok := i.(sdk.Coin)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// validate appropriate Coin
+	if coin.Validate() != nil {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
```
SpicyLemon, line 95: The result from `coin.Validate()` should be included in this error message, and the error message should be changed to reflect the different error.

proto/provenance/metadata/v1/tx.proto
SpicyLemon: Nevermind this. I was a bit confused.

SpicyLemon (commented):
Some more comments. I'm just starting on the metadata changes and felt this was a good split point. I also haven't looked at the unit test files yet.

x/metadata/keeper/msg_server.go
```diff
@@ -48,6 +59,44 @@ func (k msgServer) WriteScope(
...
 
+func (k msgServer) createExpirationForScope(ctx sdk.Context, msg *types.MsgWriteScopeRequest) (*exptypes.Expiration, error) {
+	// create expire scope request.
+	expireMsg := types.NewMsgExpireScopeRequest(msg.Scope.ScopeId, msg.Signers)
+	wrapper, anyErr := codec.NewAnyWithValue(expireMsg)
+	if anyErr != nil {
+		return nil, anyErr
+	}
+
+	// use default expiration, but override with cmd line arg if provided
```
SpicyLemon, line 70: Nit: We don't care about command line args in here. Most usage isn't through CLI anyway. There's a similar comment above (line 42 as of writing this). In here, we're just using values provided in the message.

x/metadata/keeper/msg_server.go
```diff
@@ -48,6 +59,44 @@ func (k msgServer) WriteScope(
...
+	expShort := k.GetDefaultScopeExpiration(ctx, msg.Scope)
+	if len(msg.Expiration) > 0 {
+		expShort = msg.Expiration
+	}
```
SpicyLemon, line 74: No need to make this keeper call if we aren't using the default.

```suggestion
	expShort := msg.Expiration
	if len(expShort) == 0 {
		expShort = k.GetDefaultScopeExpiration(ctx, msg.Scope)
	}
```

x/metadata/keeper/msg_server.go
```diff
@@ -48,6 +59,44 @@ func (k msgServer) WriteScope(
...
+
+	// parse expiration into a duration
+	duration, durErr := exptypes.ParseDuration(expShort)
+	if durErr != nil {
+		return nil, durErr
+	}
+
+	// REVIEW THIS: for now, using first owner for scope as expiration depositor
```
SpicyLemon, line 82: Without adding a field to the `MsgWriteScopeRequest`, I think using the first owner is the best way to go.

x/metadata/keeper/msg_server.go
```diff
@@ -48,6 +59,44 @@ func (k msgServer) WriteScope(
 	return types.NewMsgWriteScopeResponse(msg.Scope.ScopeId), nil
 }
 
+func (k msgServer) createExpirationForScope(ctx sdk.Context, msg *types.MsgWriteScopeRequest) (*exptypes.Expiration, error) {
```
SpicyLemon, line 62: This not only creates the expiration, but also sets it and charges the deposit. It would be better named `setExpirationForScope` and only returning an `error`. The `Expiration` return value is never used anyway.

x/metadata/keeper/msg_server.go
```diff
@@ -73,6 +122,31 @@ func (k msgServer) DeleteScope(
 	return types.NewMsgDeleteScopeResponse(), nil
 }
 
+func (k msgServer) ExpireScope(
```
SpicyLemon, line 125: This `ExpireScope` msg and endpoint is exactly the same as the `DeleteScope` msg endpoint.

But there's an issue to address here. When a scope is deleted, any expiration entry for it should also processed. That should happen when invoked via the expiration module, or through the `DeleteScope` endpoint. In the end, the scope (and all it's stuff) should be deleted, it's deposit should be refunded, and the expiration record should also be deleted.

I _think_ this might be solved by updating the `DeleteScope` endpoint to first check for an expiration entry for the given id. If it exists, just invoke that entry. If there isn't an entry, then delete the scope as normal. Doing it this way would be a reason to have different `ExpireScope` and `DeleteScope` endpoints, because it would prevent an infinite loop. A drawback to this is that calling the `DeleteScope` endpoint might then emit the `ExpireScope` events.

x/metadata/keeper/msg_server.go
```diff
@@ -254,7 +328,7 @@ func (k msgServer) DeleteRecord(
 	goCtx context.Context,
 	msg *types.MsgDeleteRecordRequest,
 ) (*types.MsgDeleteRecordResponse, error) {
-	defer telemetry.MeasureSince(time.Now(), types.ModuleName, "tx", "DeleteRecord")
+	defer telemetry.MeasureSince(time.Now(), types.ModuleName, "tx", "DeleteExpiration")
```
SpicyLemon, line 331: Bad change. Old string was correct.

x/metadata/types/events.go
```diff
@@ -103,7 +104,7 @@ const (
 	TxEndpoint_WriteSession TxEndpoint = "WriteSession"
 
 	TxEndpoint_WriteRecord  TxEndpoint = "WriteRecord"
-	TxEndpoint_DeleteRecord TxEndpoint = "DeleteRecord"
+	TxEndpoint_DeleteRecord TxEndpoint = "DeleteExpiration"
```
SpicyLemon, line 107: Bad change here too. Old string was correct.

x/metadata/types/msg.go
```diff
@@ -244,6 +247,56 @@ func (msg MsgDeleteScopeRequest) ValidateBasic() error {
...
+// NewMsgExpireScopeRequest creates a new msg instance
+func NewMsgExpireScopeRequest(scopeID MetadataAddress, signers []string) *MsgExpireScopeRequest {
+	return &MsgExpireScopeRequest{
+		ScopeId: scopeID,
+		Signers: signers,
+	}
+}
+
+func (msg MsgExpireScopeRequest) String() string {
+	out, _ := yaml.Marshal(msg)
```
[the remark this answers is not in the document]
SpicyLemon: I was thinking the same thing. And the SDK still does it in a lot of `String()` stuff, but I think they're trying to get away from anything `yaml` specific. When `yaml` is desired, they want to get it by creating the `json` first, then converting that to `yaml`.

x/metadata/wasm/encode.go
```diff
@@ -59,7 +59,8 @@ func (params *WriteScope) Encode() ([]sdk.Msg, error) {
 		return nil, err
 	}
 
-	msg := types.NewMsgWriteScopeRequest(*scope, params.Signers)
+	// TODO: should the structures be updated with expiration settings?
+	msg := types.NewMsgWriteScopeRequest(*scope, params.Signers, "1y")
```
SpicyLemon, line 63: Unaddressed TODO.
Also, duplication of the default expiration string.

SpicyLemon (commented):
Done with the metadata stuff. I just have some test files left to look through.

x/expiration/types/expiration.go
egaxhaj: Yes. I noticed that tests fail where fail after a `tx` broadcast because it could not find the interfaces to do the unmarshalling. I didn't know how else to handle it.

x/expiration/types/msgs.go
egaxhaj: It first checks to see if it is an AccAddress and when that fails it checks to see if we are dealing with a metadata address.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,458 @@
...
+	bankKeeper bankkeeper.Keeper,
+	router baseapp.IMsgServiceRouter,
+) Keeper {
+	// set KeyTable if it has not already been set
+	if !paramSpace.HasKeyTable() {
+		paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
+	}
+
+	// ensure governance module account is set
+	if addr := acctKeeper.GetModuleAddress(types.ModuleName); addr == nil {
```
Taztingo, line 73: It doesn't look like addr is being used. Can you just change this to `if acct.Keeper.GetModuleAddress(types.ModuleName) == nil`

proto/provenance/expiration/v1/query.proto
egaxhaj: will change to 99

x/expiration/client/cli/tx.go
egaxhaj: >  "will invoke logic be called again when the timer expires?". Might be nice to clarify that either here or in the specs.

Invocation is performed by a user. Will specify that in the specs.

x/expiration/keeper/grpc_query.go
```diff
@@ -0,0 +1,141 @@
...
+	goCtx context.Context,
+	req *types.QueryAllExpirationsByOwnerRequest,
+) (*types.QueryAllExpirationsByOwnerResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	pageRes, expirations, err := k.filteredPaginate(ctx, req.Pagination, func(expiration types.Expiration) bool {
+		return strings.TrimSpace(req.Owner) == expiration.Owner
```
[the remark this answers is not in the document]
egaxhaj: Yes, we do need to and an index. I did not attempt to add it in on this iteration of the module. I felt it would be best to add it later on after a first release was out. I created an issue for it #1123

x/expiration/types/keys.go
```diff
@@ -0,0 +1,39 @@
...
+
+// GetModuleAssetKeyPrefix returns the key prefix used by encoded objects stored in the kv store.
+func GetModuleAssetKeyPrefix(moduleAssetID string) ([]byte, error) {
+	key := ModuleAssetKeyPrefix
+	accAddress, err := sdk.AccAddressFromBech32(moduleAssetID)
+	if err != nil {
+		// check if module asset ID is a MetadataAddress
+		if _, err2 := metadatatypes.MetadataAddressFromBech32(moduleAssetID); err2 != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
egaxhaj: > 2. The expiration module shouldn't have any references to the metadata module.

I agree. However, I didn't see another way of checking for a metadata address type with referencing the metadata types.

@iramiller mentioned that we need to move away from using our own HRP types in the metadata module  (if that is possible) to avoid this exact problem but that's further down the road.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+	store.Set(key, b)
+
+	// emit Extend event
+	extendEvent := types.NewEventExpirationExtend(expiration.ModuleAssetId)
+	return k.emitEvent(ctx, extendEvent)
+}
+
+// deleteExpiration removes an expiration record from the kvstore.
+func (k Keeper) deleteExpiration(ctx sdk.Context, moduleAssetID string) error {
```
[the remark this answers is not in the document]
egaxhaj: Renamed!

> Also, is there a reason this isn't public? 

Yes. Previously it was used by both `DeleteExpiration` and `InvokeExpiration`. We later removed `DeleteExpiration` from CLI in favor of `InvokeExpiration`. I decided to keep it as a separate function to keep `InvokeExpiration` less cluttered.

x/expiration/types/codec.go
egaxhaj: Good catch! I had this in place when we were supporting `add` expiration on the cli to be able to run `ValidateBasic()` on the message that would be invoked during a tx broadcast (it would result in an error with ModuleCdc not having those registered). We have since removed the `add` functionality from the cli and these can be removed. However, we should restrict storing only `delete` message types in an expiration record. Do you have a recommendation on a way to do that without having to reference other modules?

x/metadata/keeper/scope.go
```diff
@@ -135,6 +135,12 @@ func (k Keeper) RemoveScope(ctx sdk.Context, id types.MetadataAddress) {
 	defer types.GetIncObjFunc(types.TLType_Scope, types.TLAction_Deleted)
 }
 
+// GetDefaultScopeExpiration gets default expiration for scope.
+func (k Keeper) GetDefaultScopeExpiration(ctx sdk.Context, scope types.Scope) string {
+	// hard-coded value for now, but may be calculated in the future based on scope info..
+	return "1y"
```
Taztingo, line 141: Minor nit, but would be nice if this was a constant.

x/expiration/types/params.go
egaxhaj: I set this as `1000000000nhash`.

> The denom should come from the config.

If this value is moved to the config, would that not lead to inconsistent state when set to a different value from node to node? I am not certain on this but it may also effect `state sync`.

x/expiration/types/codec.go
SpicyLemon: I think the way to do this is to have a field and registration function to the expiration keeper that stores and adds-to such a whitelist. Then, in a module that is going to use the expiration module (e.g. `x/metadata/module.go`), create a function that takes in an expiration keeper and registers the allowed messages. It wouldn't allow for use during basic validation, but could still be enforced in the other validation.

Then, in `app/app.go`, we'd add a function `initExpirationKeeper` (and call it in `New`) that calls all the module functions that register the whitelisted messages. The easy way to do it is to just hard-code all needed calls (similar to `initParamsKeeper`). The fun way would be to define an interface that a module would satisfy if they want to register whitelisted messages, then go through each module in the module manager and (if it implements that interface), call the method.

x/expiration/types/expiration.go
SpicyLemon: Unresolving this until I see it fixed in the diff since it's such a necessity to have fixed.

x/expiration/types/expiration.go
SpicyLemon: I was asking why we should have a whitelist and this piece of validation at all. In other words, why do we need to restrict messages used in an expiration? What's the downside to not having this piece of validation?

The codec in the keeper should have all message types registered. So it should be able to unpack any message just fine. But this validation here is making sure that the message type is in a smaller whitelist of message types, and it's not clear to me why we need that.

The stuff in the `types` package should be mostly static stuff. The `keeper` package is for stuff that can change based on setup. We really really don't want the expiration module to have dependencies on modules that are going to use it. Specifically, in this case, the `x/expiration` module should not have any imports to anything in the `x/metadata` module. And `types` is not the place to store a whitelist that is built as part of app setup. So this piece of validation isn't really suitable for a `ValidatBasic()`.

This piece of validation should be split out to it's own function that takes in a codec. Then, in the keeper (where we've got a codec and maybe a whitelist codec), make a call to that function as part of further validation. Then, if we decide we still want a whitelist, provide the whitelist codec, otherwise, just provide the full codec.

x/expiration/types/expiration.go
SpicyLemon: Please also document it in the proto. Spec documentation is good too, but the protos are the primary source of truth and the most readily available source of documentation that a consumer has.

x/expiration/types/msgs.go
SpicyLemon: Ah. I see it now. It can be either a regular chain address or a metadata address.

This presents another problem though. We don't want the expiration module to import anything from the metadata module. As such, we probably can't have this check in the `types` package since there will need to be some sort of function registry that the expiration module can use to check if an address is usable. The interface would be something like `IsExpirableAddress(string) bool`. And the expiration module's `validateAddress` function would then loop through those, returning false only if all of them return false (maybe also have the `sdk.AccAddressFromBech32(s)` check, but maybe not). Or maybe it'd be more like an antehandler setup where you chain them together, each one calling the next. I'm not sure which is best.

For now though, maybe just use the `bech32.DecodeAndConvert` function from `"github.com/cosmos/cosmos-sdk/types/bech32"` library directly, e,.g. `hrp, bz, err := bech32.DecodeAndConvert(address)`.

x/metadata/keeper/msg_server.go
```diff
@@ -37,6 +39,15 @@ func (k msgServer) WriteScope(
 	//nolint:errcheck // the error was checked when msg.ValidateBasic was called before getting here.
 	msg.ConvertOptionalFields()
 
+	// temporarily only create expiration if an expiration value is specified (e.g. via flag on command)
```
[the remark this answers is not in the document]
SpicyLemon: In testing, you might be providing the expiration via CLI with the `--expires` flag, but in here (the msg server), you've no idea how they generated the msg that got you to that point. So it's weird to specifically reference a CLI flag in here, even if as an example; you really don't need the `e.g.` in my opinion.

SpicyLemon:
I noticed a lot of my comments marked as resolved without actually being addressed. I understand resolving them if you're just about to push an update that fixes them, but even that doesn't seem to be the case.

x/expiration/keeper/msg_server.go
```diff
@@ -0,0 +1,92 @@
...
+	expiration, err := m.Keeper.GetExpiration(ctx, msg.ModuleAssetId)
+	if err != nil {
+		return nil, err
+	}
+
+	duration, err := types.ParseDuration(msg.Duration)
+	if err != nil {
+		return nil, err
+	}
+	expiration.Time = ctx.BlockTime().Add(*duration)
```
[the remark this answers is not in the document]
egaxhaj: will change this to extend the existing period and comment it on the proto and function.

x/expiration/types/codec.go
egaxhaj: 👍 I'm going to attack this in a new ticket -> #1132.

x/expiration/types/expiration.go
egaxhaj: comment added to the proto:
```
// the duration period by which to extend the existing expiration period
// the accepted values for a duration are n{h,d,w,y}
// where 1h = 60m, 1d = 24h, 1w = 7d (or 168h), 1y = 365d (or 8760h)
string duration = 2;
  ```

File: x/expiration/types/expiration.go (hunk 22 above)
Comment (egaxhaj) [reply x/expiration/types/expiration.go:68]: marking this as resolved since we talked about a different approach in https://github.com/provenance-io/provenance/pull/1115#discussion_r988299600. 

We're safe to remove the `UnpackAny` check because new expirations are only created from other modules.

File: proto/provenance/expiration/v1/expiration.proto [hunk 46]
```diff
@@ -0,0 +1,35 @@
...
+option java_package        = "io.provenance.expiration.v1";
+option java_multiple_files = true;
+
+// Params defines the set of params for the expiration module.
+message Params {
+  option (gogoproto.equal)            = false;
+  option (gogoproto.goproto_stringer) = false;
+
+  // the deposit amount for holding module assets on chain
+  cosmos.base.v1beta1.Coin deposit = 2 [(gogoproto.nullable) = false];
```
SpicyLemon, line 20: Is there a reason to limit this to a single `Coin` instead of using `Coins`?

egaxhaj:
> I noticed a lot of my comments marked as resolved without actually being addressed. I understand resolving them if you're just about to push an update that fixes them, but even that doesn't seem to be the case.

Sorry about the confusion. I mark them as resolved once I've made a change locally. It helps keep track of what is left to resolve before I submit a PR. Because of the scope of the work that was being done to merge 0.46 we avoided submitting PRs to `main` and instead @derekadams created `epic/expiration` branch to keep track of the different items through PRs agains it vs. pushing to it. All the resolved items are handled in #1135

proto/provenance/expiration/v1/query.proto
egaxhaj: updated paths to no longer conflict.

x/metadata/types/msg.go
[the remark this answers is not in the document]
egaxhaj: Removed `String()` and using `stringer` in the proto.

x/expiration/types/params.go
egaxhaj: This has been updated to use `sdk.NewInt64Coin(NhashDenom, 1000000000)`

x/metadata/client/cli/tx.go
```diff
@@ -1000,6 +1004,20 @@ func parseSigners(cmd *cobra.Command, client *client.Context) ([]string, error)
...
+	cmd.Flags().String(FlagExpires, "", "expiration period (e.g. '1y')")
+}
+
+// parseExpires checks expires flag, else uses the default expiration period
+func parseExpires(cmd *cobra.Command) string {
+	flagSet := cmd.Flags()
+	if flagSet.Changed(FlagExpires) {
+		expires, _ := flagSet.GetString(FlagExpires)
+		return expires
+	}
```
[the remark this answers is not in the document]
egaxhaj: removed `.Changed` and used `.GetString` directly.

x/metadata/keeper/msg_server.go
egaxhaj: removed references to CLI from comment.

x/metadata/keeper/msg_server.go
egaxhaj: updated as suggested.

x/metadata/keeper/msg_server.go
egaxhaj: renamed to `setExpirationForScope` and changed method signature to return only `error`.

x/metadata/keeper/msg_server.go
egaxhaj: fixed.

proto/provenance/expiration/v1/expiration.proto
egaxhaj: Yes, my understating is that we're only supporting `nhash` deposits. 

@derekadams is the ^^^ assumption correct or should this be a repeated field of type `Coins`?

x/metadata/keeper/msg_server.go
egaxhaj: updated comment to `using first owner for scope as expiration depositor`

x/metadata/keeper/scope.go
egaxhaj: this is now using the `expirationtypes.DefaultDuration`

x/metadata/keeper/msg_server.go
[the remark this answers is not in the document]
egaxhaj: removed references to `cli` from comment and added additions comments about 'temporary`

x/metadata/wasm/encode.go
egaxhaj: added a new field `Expiration string` to `WriteScope` and passing that in. This will eventually default to `expiration types.DefaultDuration` based on https://github.com/provenance-io/provenance/pull/1115#discussion_r987949444

x/expiration/types/params.go
egaxhaj: Changed to default sdk.NewInt64Coin(NhashDenom, 0) it's only meaningful when set in params.

x/metadata/keeper/msg_server.go
egaxhaj: I added `RemoveExpiration` to the expiration keeper for modules to call and removed `ExpireScope` from metadata keeper. When `DeleteScope` is called it calls `RemoveExpiration` (returns `nil` if no expiration exists) and deletes the scope.

proto/provenance/expiration/v1/expiration.proto
derekadams: I agree that `Coins` makes more sense 👍

egaxhaj: Switched to using `Coins`

x/expiration/spec/01_concepts.md
```diff
@@ -0,0 +1,101 @@
...
+the module asset ID is a bech32 address string or a [MetadataAddress](../../metadata/spec/01_concepts.md) 
+that assigned to each asset on chain.
+
+### Owner
+
+The owner is a bech32 address string that determines the rightful owner to a particular expiring asset on chain.
+
+### Expiration Time
+
+The expiration time determines when an asset on chain will expire and is up for removal.
```
Taztingo, line 44: > up for removal

You can improve the clarity by changing this to "ready for removal".

x/expiration/spec/01_concepts.md
```diff
@@ -0,0 +1,101 @@
...
+### Message
+
+When expiration logic is executed, the module asset message stored in the expiration metadata is dispatched to the associated module
+through the [Msg Service Router](https://docs.cosmos.network/main/core/baseapp.html#msg-service-router).
+
+## Creating Expirations on Assets
+
+Creating an expiration is done through the module that owns the asset. When a new asset is created for a particular module,
+the user creating the asset has the option to specify if the asset in question will expire at some time in the future.
+Note that certain module assets can never expire. Therefore, asset expiration is an opt-in feature that asset owners can
```
Taztingo, line 63: This seems pretty important. You might want to increase the visibility by making this a separate subsection.

```
### Note
Certain module assets will never expire. Therefore, asset expiration is an opt-in feature that asset owners can leverage for their on-chain assets.
```

x/expiration/spec/01_concepts.md
```diff
@@ -0,0 +1,101 @@
...
+leverage for their on-chain assets.
+
+## Extending Expirations on Assets
+
+Extending an expiration can be done through the CLI or through auto-updates from the modules that own them. 
+Note that auto-updates are left up to the module that owns the asset.
```
Taztingo, line 69: Same as above.

x/expiration/spec/01_concepts.md
```diff
@@ -0,0 +1,101 @@
...
+### Expiration Time
+
+The expiration time determines when an asset on chain will expire and is up for removal.
+
+### Deposit
+
+When an expiration record is added to the system, funds, in the form of a deposit, will be moved from the owner address
+associated with the module asset and stored in the account associated with the expiration module. The deposit can be returned
+when the owner later executes the expiration logic which frees up the funds held for the underlying asset.Note that if the
+expiration logic is not triggered and the expiration time passes, external actors may take action to execute the expiration
```
Taztingo, line 51: Missing a space before the word Note. You may also want to make this a separate subsection.

`###Note`

x/expiration/spec/02_state.md
```diff
@@ -0,0 +1,31 @@
+# State
+
+The expiration module maintains a simple state collection. The `Expiration` is the data structure stored on the blockchain.
+It consists of five parts.
```
Taztingo, line 4: May want to rephrase and use a colon instead.

`It consists of the following five parts:`

x/expiration/types/msgs.go
```diff
@@ -0,0 +1,151 @@
...
+	addresses := make([]sdk.AccAddress, len(signers))
+	for i, address := range signers {
+		addresses[i] = sdk.MustAccAddressFromBech32(address)
+	}
+	return addresses
+}
+
+// private method that does a simple validation check that
+// doesn't require access to any other information.
+func validateBasic(expiration Expiration, signers []string) error {
```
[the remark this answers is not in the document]
egaxhaj: I can't recall why I made this private. I folded logic to `ValidateBasic`.

x/expiration/spec/01_concepts.md
egaxhaj: fixed.

x/expiration/spec/01_concepts.md
egaxhaj: updated.

x/expiration/spec/01_concepts.md
egaxhaj: updated.

x/expiration/spec/01_concepts.md
egaxhaj: moved to `### Note`

x/expiration/spec/02_state.md
egaxhaj: fixed.

proto/provenance/attribute/v1/tx.proto
```diff
@@ -42,6 +42,9 @@ message MsgAddAttributeRequest {
   string account = 4;
   // The address that the name must resolve to.
   string owner = 5;
+  // expiration is an optional expiration timespan string, e.g. "1y" (see expiration module for format)
```
nullpointer0x00, line 45: You have this in another comment: ```  // the accepted values for a duration are n{h,d,w,y}
  // where 1h = 60m, 1d = 24h, 1w = 7d (or 168h), 1y = 365d (or 8760h) ```
Maybe add that instead?

nullpointer0x00 (commented):
I'm noticing a lot of missed coverage with tests in the expiration keeper.go and a few Validate methods in the types.  Should we add a bit more coverage?

app/app.go
```diff
@@ -216,6 +220,8 @@ var (
 		markertypes.ModuleName: {authtypes.Minter, authtypes.Burner},
 		wasm.ModuleName:        {authtypes.Burner},
 		rewardtypes.ModuleName: nil,
+
```
nullpointer0x00, line 223: NIT: remove new line

x/expiration/types/expiration.go
```diff
@@ -0,0 +1,100 @@
...
+// and is registered and whitelisted with the InterfaceRegistry
+func (e *Expiration) validateMessage() error {
+	// this may occur in message decoding
+	if e.Message.Value == nil {
+		return fmt.Errorf("expecting non nil value to validate an Any message")
+	}
+	if e.Message.TypeUrl == "" {
+		return fmt.Errorf("expecting message type URL to unpack Any")
+	}
+	return nil
```
nullpointer0x00, line 63: Expiration looks to contain a `sdk.Msg` that will get dispatched later.  We should probably call the `ValidateBasic()` on the msg as well.

iramiller: I agree.... the nested msgs should have their ValidateBasic called by the parent msg ValidateBasic flow... beyond this check no deeper validation on the nested msg would be appropriate.

x/expiration/types/expiration.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// validateMessage validates `expiration.Message` conforms to `sdk.Msg`
+// and is registered and whitelisted with the InterfaceRegistry
+func (e *Expiration) validateMessage() error {
+	// this may occur in message decoding
+	if e.Message.Value == nil {
+		return fmt.Errorf("expecting non nil value to validate an Any message")
+	}
+	if e.Message.TypeUrl == "" {
```
nullpointer0x00, line 60: Do we need to validate that this msg is of a certain TypeUrl?

proto/provenance/metadata/v1/tx.proto
egaxhaj: ExpireScope has been removed. Defaulted back to using DeleteScope msg.

x/expiration/types/expiration.go
egaxhaj: Yes this one is a bit tricky to solve.

I initially was registering the `delete` Msg types with the codec to be able to Unmashal and call `ValidateBasic()` on the `msg` but we don't want the expiration module to have references to other modules as it may lead to circular dependencies (ref: https://github.com/provenance-io/provenance/pull/1115#discussion_r987141369). I have an open ticket to address this one based on what @SpicyLemon and I discussed.


```go
func whiteListedInterfaceRegistry() types.InterfaceRegistry {
 	ir := types.NewInterfaceRegistry()
 	ir.RegisterImplementations((*sdk.Msg)(nil),
 		&metadatatypes.MsgDeleteScopeRequest{},
 		&metadatatypes.MsgDeleteRecordRequest{},
...
```

https://github.com/provenance-io/provenance/blob/9f743b422abde49db364c95211ecd9524d9a1ff6/x/expiration/client/cli/tx.go#L287-L293

egaxhaj: Actually it's simpler than I thought. We can call `msg.ValidateBasic()` in `SetExpiration`. Same as we do in `InvokeExpiration`.

x/expiration/types/expiration.go
egaxhaj: Yes, but there's no mechanism in place at the moment that allows us to do that. (ref: https://github.com/provenance-io/provenance/pull/1115#discussion_r987141369)

proto/provenance/attribute/v1/tx.proto
egaxhaj: updated comments.

>// expiration is an optional expiration timespan string, e.g. "1y"
  // If provided, it sets an expiration with the given time period for the attribute.
  // accepted values are n{h,d,w,y}
  // where 1h = 60m, 1d = 24h, 1w = 7d (or 168h), 1y = 365d (or 8760h)

app/app.go
egaxhaj: fixed

x/expiration/types/expiration.go
egaxhaj: added `ValidateBasic()` in the keeper. 

https://github.com/provenance-io/provenance/blob/19afc634a6204f27598350ce59ecd578419ee2bd/x/expiration/keeper/keeper.go#L136-L149

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,565 @@
...
+			types.ErrMsgHandler.Wrapf("got nil sdk.Result for message %q", msg).Error())
+	}
+
+	// refund deposit from expiration module account to depositor
+	if err := k.refundDeposit(ctx, expiration.Deposit, refundTo); err != nil {
+		return types.ErrInvokeExpiration.Wrapf("failed to refund deposit to [%s]: %v", refundTo, err)
+	}
+
+	// remove record from store
+	if err := k.removeExpiration(ctx, moduleAssetID); err != nil {
```
nullpointer0x00, line 279: This might not be of concern for this PR. I was wondering if there is a case where this method fails (and will always fail...due to some issue) and never reaches the `removeExpiration` function.  The expiration will remain in the keeper and someone could keep calling `InvokeExpiration` on that expiration since it will always show up as a valid one to call when queried.

proto/provenance/expiration/v1/events.proto
```diff
@@ -0,0 +1,55 @@
+// Since: provenance 1.12
```
SpicyLemon, line 1: Not accurate. Also, I'm not a fan of the "since ..." comments that the sdk has started using everywhere; I just don't see them as useful.

proto/provenance/expiration/v1/expiration.proto
```diff
@@ -0,0 +1,43 @@
...
+  // the module asset identifier
+  string module_asset_id = 1;
+  // The bech32 address the expiration is bound to
+  string owner = 2;
+  // The time the module asset expires
+  google.protobuf.Timestamp time = 3 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+  // The deposit amount held while module asset is in use
+  repeated cosmos.base.v1beta1.Coin deposit = 4
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+  // Message relating to the expiring module asset
```
SpicyLemon, line 41: I'd like this to have more info. "relating to" doesn't really answer the question, "What is this field for?"

proto/provenance/expiration/v1/query.proto
SpicyLemon, line 8: This can be deleted.

proto/provenance/expiration/v1/tx.proto
```diff
@@ -0,0 +1,72 @@
...
+message MsgAddExpirationRequest {
+  option (gogoproto.equal)            = false;
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.stringer)         = true;
+  option (gogoproto.goproto_getters)  = false;
+
+  // expiration is the Expiration you want to add
+  Expiration expiration = 1 [(gogoproto.nullable) = false];
+  // signers is the list of addresses of those signing this request.
+  repeated string signers = 2;
```
SpicyLemon, line 32: An expiration can only have one owner. So, there's no need to have multiple signers, is there?

SpicyLemon (commented):
A few more proto notes.

Also, once the proto stuff is finalized, be sure to `make proto-swagger-gen`.

x/attribute/client/cli/cli_test.go
```diff
@@ -3,6 +3,8 @@ package cli_test
 import (
 	"encoding/base64"
 	"fmt"
+	expcli "github.com/provenance-io/provenance/x/expiration/client/cli"
+	exptypes "github.com/provenance-io/provenance/x/expiration/types"
```
SpicyLemon, line 7: Nit: Incorrect import order.

x/attribute/client/cli/cli_test.go
```diff
@@ -591,6 +593,83 @@ func (s *IntegrationTestSuite) TestAttributeTxCommands() {
...
+		args            []string
+		expectErr       bool
+		respType        proto.Message
+		expectedCode    uint32
+		expectExpRecord bool
+	}{
+		{
+			"bind a new attribute name for testing",
+			namecli.GetBindNameCmd(),
+			[]string{
```
SpicyLemon, line 610: Please use named parameters for these definitions. It makes them much easier to read and maintain.

x/attribute/client/cli/cli_test.go
```diff
@@ -591,6 +593,83 @@ func (s *IntegrationTestSuite) TestAttributeTxCommands() {
 	}
 }
 
+func (s *IntegrationTestSuite) TestAttributeTxCommands_WithExpiration() {
+
+	testCases := []struct {
+		name            string
+		cmd             *cobra.Command
+		args            []string
+		expectErr       bool
```
SpicyLemon, line 602: An `expectErr bool` is almost never worthwhile. If you're expecting an error, you should testing the error message for specific content. Otherwise, things can be going drastically wrong in unexpected ways, but the test will pass because it doesn't realize the error received has nothing to do with the test.

x/attribute/client/cli/cli_test.go
```diff
@@ -591,6 +593,83 @@ func (s *IntegrationTestSuite) TestAttributeTxCommands() {
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		s.Run(tc.name, func() {
+			clientCtx := s.testnet.Validators[0].ClientCtx
+			out, err := clitestutil.ExecTestCLICmd(clientCtx, tc.cmd, tc.args)
+
+			if tc.expectErr {
+				s.Require().Error(err)
```
SpicyLemon, line 654: Please always include messages in test assertions. When asserting stuff on errors, the message should answer, "What action was taken to produce this error." In this case, just "ExecTestCLICmd" would be acceptable.

x/attribute/client/cli/cli_test.go
```diff
@@ -591,6 +593,83 @@ func (s *IntegrationTestSuite) TestAttributeTxCommands() {
...
+			} else {
+				s.Require().NoError(err)
+				s.Require().NoError(clientCtx.Codec.UnmarshalJSON(out.Bytes(), tc.respType), out.String())
+				switch t := tc.respType.(type) {
+				case *sdk.TxResponse:
+					s.Require().Equal(tc.expectedCode, t.Code)
```
SpicyLemon, line 660:
```suggestion
					s.Require().Equal(int(tc.expectedCode), int(t.Code), "response code from:\n%s", t)
```

By converting them to ints, they're easier to understand when they fail. Without that conversion, they are printed as hexadecimal, e.g. "Expected: 0x00, Actual: 0x10". That wouldn't be error code 10, it'd be 16. It's easy to confuse and waste time.

Also, by including the response in the message, it saves time trying to figure out what exactly the outcome was. This is crucial when tests fail intermittently, but still nice to have when it's a consistent failure.

x/attribute/client/cli/tx.go
```diff
@@ -222,3 +229,14 @@ func NewDeleteAccountAttributeCmd() *cobra.Command {
...
+func addExpiresFlagCmd(cmd *cobra.Command) {
+	cmd.Flags().String(FlagExpires, "", "expiration period (e.g. '1y')")
+}
+
+// parseExpires checks expires flag, else uses the default expiration period
+func parseExpires(cmd *cobra.Command) string {
+	flagSet := cmd.Flags()
+	expires, _ := flagSet.GetString(FlagExpires)
+	return expires
+}
```
SpicyLemon, line 242: These should be public in the expiration module. Also, the description should have more details. Specifically, I'd like to see all the options present.

SpicyLemon (commented):
Some CLI and test notes.

The comments on the tests apply to all tests, not just the ones I specifically commented on.

x/attribute/keeper/msg_server.go
```diff
@@ -43,6 +45,13 @@ func (k msgServer) AddAttribute(goCtx context.Context, msg *types.MsgAddAttribut
 		return nil, err
 	}
 
+	// create an expiration record if an expiration value is specified
+	if len(msg.Expiration) > 0 {
+		if err := k.setExpirationForAttribute(ctx, attrib, ownerAddr, msg.Expiration); err != nil {
+			return nil, err
+		}
+	}
+
```
SpicyLemon, line 54: In order to use the default expiration stuff, `setExpirationForAttribute` needs to always be called. In that function, if there really isn't an expiration value skip adding the expiration.
```suggestion
	if err := k.setExpirationForAttribute(ctx, attrib, ownerAddr, msg.Expiration); err != nil {
		return nil, err
	}

```

proto/provenance/expiration/v1/expiration.proto
```diff
@@ -0,0 +1,43 @@
...
+  option (gogoproto.stringer)         = true;
+  option (gogoproto.goproto_stringer) = false;
+
+  // the deposit amount for holding module assets on chain
+  cosmos.base.v1beta1.Coin deposit = 1 [(gogoproto.nullable) = false];
+  // duration is the period upon which an expiration will stay on chain
+  // when the period elapses the module asset is up for extension or deletion
+  //
+  // the accepted values for a duration are n{h,d,w,y}
+  // where 1h = 60m, 1d = 24h, 1w = 7d (or 168h), 1y = 365d (or 8760h)
```
SpicyLemon, line 26: Please use sentences in all these proto comments. Line spacing isn't maintained everywhere.

proto/provenance/expiration/v1/expiration.proto
```diff
@@ -0,0 +1,43 @@
...
+  string duration = 2;
```
SpicyLemon, line 27: I can't imagine this being useful unless it's a default. It'd be good to name it that way.

x/attribute/keeper/keeper.go
```diff
@@ -297,6 +302,12 @@ func (k Keeper) DeleteAttribute(ctx sdk.Context, addr string, name string, value
 	return nil
 }
 
+func (k Keeper) GetDefaultAttributeExpiration(ctx sdk.Context, attr types.Attribute) string {
+	// use default duration param value from expiration module now,
+	// but may be calculated in the future based on attribute info.
+	return exptypes.DefaultDuration
+}
+
```
SpicyLemon, line 310: Since there's nothing looked up here, this function doesn't need to take in any arguments.

That being said, the expiration params have a `duration` entry, but it's not being used here. that `exptypes.DefaultDuration` is code defined and never updated, so it's not going to match what the params have.

x/attribute/keeper/msg_server.go
```diff
@@ -203,3 +222,42 @@ func (k msgServer) DeleteDistinctAttribute(goCtx context.Context, msg *types.Msg
...
+	// parse expiration into a duration
+	duration, durErr := exptypes.ParseDuration(expShort)
+	if durErr != nil {
+		return durErr
+	}
+
+	// create expiration metadata that will expire scope when executed.
+	expTime := ctx.BlockTime().Add(*duration)
+	expDeposit := sdk.NewCoins(k.expKeeper.GetDeposit(ctx))
+	expiration := exptypes.NewExpiration(owner.String(), owner.String(), expTime, expDeposit, *wrapper)
```
SpicyLemon, line 256: The owner address is not a good thing to use as the `ModuleAssetID`. Doing so means they can only ever have one expiration. If someone owns multiple attributes, this falls apart.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,565 @@
...
+
+	return nil
+}
+
+// InvokeExpiration invokes an expiration message through the MsgServiceRouter.
+// The expiration message is removed when the message invocation is successful.
+//
+// Caution: Do NOT call InvokeExpiration from within a module to delete a self-owned asset.
+// This will lead to an infinite loop as expiration messages in this method are routed through
+// the message router. Use RemoveExpiration instead.
```
SpicyLemon, line 239: It'd probably be worthwhile to create an `exported/exported.go` with an interface that this keeper fulfills, but only has the functions we expect people to want in other modules, e.g. `SetExpiration` and `RemoveExpiration` (and maybe `GetDeposit` and `GetDuration` depending on other discussions).

Modules that want to implement expirations could then import that for the field in their keeper. That would dissuade use of this function and also cut down on circular dependency possibilities.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,565 @@
...
+
+	return expiration, nil
+}
+
+// ResolveDepositor resolves the account address where the expiration message deposit will be refunded.
+func (k Keeper) ResolveDepositor(
+	ctx sdk.Context,
+	expiration types.Expiration,
+	msg *types.MsgInvokeExpirationRequest,
+) (sdk.AccAddress, error) {
```
SpicyLemon, line 416: I feel like this is significantly more complicated than it needs to be. The account that gets the deposit should be the one signing the `MsgInvokeExpirationRequest`.

No matter what, it should be the signer of the `MsgInvokeExpirationRequest`.

If it's expired, there's nothing left to check. If it's not expired, that signer needs to be the owner.

This function would also be better named something like `GetRefundTo` since it's not actually trying to look up who made the deposit.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,565 @@
...
+		return false, fmt.Errorf("invalid owner: %w", err)
+	}
+	grantee, err := sdk.AccAddressFromBech32(signer)
+	if err != nil {
+		return false, fmt.Errorf("invalid signers: %w", err)
+	}
+
+	authorization, exp := k.authzKeeper.GetAuthorization(ctx, grantee, granter, msgTypeURL)
+	if authorization != nil {
+		resp, err := authorization.Accept(ctx, nil)
```
SpicyLemon, line 494: The message being processed needs to be provided to this `Accept` call. We don't have any expectations about either the authorizations being used or the messages they're being applied to, so standard authz processing needs to be used.

proto/provenance/expiration/v1/tx.proto
egaxhaj: Yes, the expiration module supports `grants`. One reason for adding `authz` support is the metadata module. The other is that it would be nice to support deleting of assets with granted permissions.

x/expiration/keeper/keeper.go
```diff
@@ -0,0 +1,565 @@
...
+			switch {
+			case resp.Delete:
+				err = k.authzKeeper.DeleteGrant(ctx, grantee, granter, msgTypeURL)
+				if err != nil {
+					return false, err
+				}
+			case resp.Updated != nil:
+				if err = k.authzKeeper.SaveGrant(ctx, grantee, granter, resp.Updated, exp); err != nil {
+					return false, err
+				}
```
SpicyLemon, line 508: This is going to require some discussion.

1. Do the signers on an expiration's message need to be validated at time of expiration creation?
2. When it gets invoked?
3. When should any `authz` authorization resources be consumed? At the very least, I know they need to only be consumed once.
4. Should they be validated at all?

If we remove the `AddExpiration` endpoint, and only allow creation of expirations via code in other modules, I feel like we shouldn't care about the signers at all. Simply trust that the module isn't trying to do anything nefarious.

If we don't check the signers when the expiration is invoked, it can lead to a situation where a thing is created by one person (along with its expiration), then sent (sold?) to another (i.e. the things owner changes), then the original owner triggers the expiration of the thing. If expirations can only be created by another module, it'd be up to that module to properly update the expiration entry to reflect the new owner.

If we do check the signers when the expiration is invoked, what should happen when that validation fails? Again, this could happen when a things owners change, but the expiration is updated. We could easily end up stuck with an expiration entry that can't be invoked by anyone.

If `authz` gets involved in that signer checking, things get really crazy. It'd be easy for an authorization to expire between when an expiration is created and when it ends. Basically, there's no guarantee that an authorization that is valid now, will still be valid (or applicable) in the future. Conversely, just because an authorization doesn't exist now, doesn't mean it won't in the future. As such, I feel like there's no valid answer to "When should an authorization's resourced be used?"

SpicyLemon (commented):
I've still got a lot of files to review here, but I feel like there's some conceptual stuff to address here. I also feel like there's a disconnect between the module's functionality and its purpose.

egaxhaj:
Moving this to draft per discussion on v1.15.0 release (02/13/2023). Besides attribute expiration, there hasn't been a real need for something as complex as the expiration module. Therefore, this work has been moved to the backlog (for a quarter or two) to gage if there is a real need for it. In the meantime we'll add lightweight implementation that can handle attribute expirations.

Issue #1136, opened 2022-10-06
Allow restricted marker transfer to happen over IBC


arnabmitra:
## Summary
Allow restricted markers to be moved by ibc

## Problem Definition
This issue is related to #1127 and #1093
Currently restricted markers are not allowed to be moved by ibc.
This issue and associated code changes should allow the restricted markers to be moved over IBC so that Figure Equity Solution can use it.

## Proposal
Step 1. Create a marker path to allow for transfer of restricted markers.
This should follow the same path as transfer i.e check for authz grants to allow it etc.

Step 2: Change the fork of ibc-go to allow the caller to specify if they have already checked the send enabled bank restrictions and hence a regular ibc transfer can proceed.

Ibc markers will still need to be created on the destination chain for now, to bring them into the marker paradigm.
We will try to address the automatic creation of markers in a later ticket.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1158, opened 2022-10-14
1136 restricted marker transfer over ibc

nullpointer0x00:
## Description

Allow restricted markers to be moved by ibc.  This entails adding a new grpc endpoint that is similar to marker transfer.  This includes a cli command as well.

closes: #1136

---

write a little note why.


x/marker/types/events.go
```diff
@@ -33,6 +33,8 @@ const (
 	EventTelemetryKeyMint string = "mint"
 	// EventTelemetryKeyTransfer transfer telemetry metrics key
 	EventTelemetryKeyTransfer string = "transfer"
+	// EventTelemetryKeyIbcTransfer ibctransfer telemetry metrics key
```
arnabmitra, line 36: do we need this telemetry key? just asking

x/marker/types/msg.go
```diff
@@ -515,6 +523,54 @@ func (msg MsgTransferRequest) GetSigners() []sdk.AccAddress {
...
+func (msg MsgIbcTransferRequest) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(&msg)
+	return sdk.MustSortJSON(bz)
+}
+
+// GetSigners indicates that the message must have been signed by the address provided.
+func (msg MsgIbcTransferRequest) GetSigners() []sdk.AccAddress {
+	adminAddr, err := sdk.AccAddressFromBech32(msg.Administrator)
+	if err != nil {
+		panic(err)
```
arnabmitra, line 568: nit: if we use `MustAccAddressFromBech32`  we can skip line 567 and 568 i think

arnabmitra:
overall lgtm

x/marker/spec/03_messages.md
```diff
@@ -243,6 +244,17 @@ This service message is expected to fail if:
   - The given administrator address does not currently have the "transfer" access granted on the marker
   - The marker types is not `RESTRICTED_COIN`
 
+## Msg/IbcTransferRequest
+
+Ibc transfer Request defines the Msg/IbcTransfer request type.  A ibc transfer request is used to transfer coin between two
```
Taztingo, line 249: A few typos. Feel free to use the following or take pieces from it...

The IbcTransferRequest is used to transfer `RESTRICTED_COIN` type markers to another chain via ibc. These coins have their `send_enabled` flag disabled by the `bank` module and thus cannot be sent using a normal `send_coin` operation.  

### Note
A transfer request also requires a signature from an account with the transfer permission as well as approval from the account the funds will be withdrawn from.

x/marker/keeper/marker.go
```diff
@@ -697,6 +700,62 @@ func (k Keeper) TransferCoin(ctx sdk.Context, from, to, admin sdk.AccAddress, am
...
+	sender, admin sdk.AccAddress,
+	receiver string,
+	timeoutHeight clienttypes.Height,
+	timeoutTimestamp uint64,
+	checkRestrictionsHandler ibckeeper.CheckRestrictionsHandler) error {
+	m, err := k.GetMarkerByDenom(ctx, token.Denom)
+	if err != nil {
+		return fmt.Errorf("marker not found for %s: %w", token.Denom, err)
+	}
+	if m.GetMarkerType() != types.MarkerType_RestrictedCoin {
```
Taztingo, line 718: Does it matter if the marker is restricted? Can they get around this by using a normal transfer on unrestricted coins? If that is the case, do you think it would convenient if we just encourage users to send all types of markers with this implementation?

Taztingo (changes requested):
Just two minor recommendations/comments, but looks great for the most part.

Taztingo (dismissed):

x/marker/keeper/marker.go
iramiller: I don't believe that users should use this api as the endpoint requires transfer auth to be granted and that grant can only be made for restricted coin markers. Adding extra logic to bypass this requirement seems unwarranted when there is an existing and more common api on the bank module already.

x/marker/keeper/marker.go
```diff
@@ -697,6 +700,62 @@ func (k Keeper) TransferCoin(ctx sdk.Context, from, to, admin sdk.AccAddress, am
...
+		return fmt.Errorf("marker type is not restricted_coin, brokered transfer not supported")
+	}
+	if !m.AddressHasAccess(admin, types.Access_Transfer) {
+		return fmt.Errorf("%s is not allowed to broker transfers", admin.String())
+	}
+	if !admin.Equals(sender) {
```
iramiller, line 724: The admin right is not the correct grant to use for a transfer operation.  That is what the transfer grant is for.

x/marker/keeper/msg_server.go
```diff
@@ -442,6 +443,51 @@ func (k msgServer) Transfer(goCtx context.Context, msg *types.MsgTransferRequest
...
+	if err != nil {
+		return nil, err
+	}
+	admin, err := sdk.AccAddressFromBech32(msg.Administrator)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.IbcTransferCoin(ctx, msg.Transfer.SourcePort, msg.Transfer.SourceChannel, msg.Transfer.Token, from, admin, msg.Transfer.Receiver, msg.Transfer.TimeoutHeight, msg.Transfer.TimeoutTimestamp, func(ctx sdk.Context, k ibckeeper.Keeper, sender sdk.AccAddress, token sdk.Coin) (canTransfer bool, err error) {
+		if !k.GetSendEnabled(ctx) {
```
iramiller, line 467: This check is not appropriate for a restricted marker. The transfer operation checks the transfer grant has been assigned to a user and thus bypasses this check

iramiller (changes requested):
A few points related to the required checks for transfer authority.

x/marker/types/msg.go
nullpointer0x00: True, I will change the others in this file to match.

x/marker/keeper/msg_server.go
arnabmitra: > This check is not appropriate for a restricted marker. The transfer operation checks the transfer grant has been assigned to a user and thus bypasses this check

that is the ibc module send enabled, not the bank send enabled flag, imo ibc module send enabled should still be honored since it maybe useful to turn off all transfers.

x/marker/types/events.go
nullpointer0x00: I meant to add telemetry to match the regular send.  Added.

x/marker/spec/03_messages.md
nullpointer0x00: Updated.

iramiller (approved):
seems like the send enabled and transfer checks are all good.

Taztingo (approved):

PR #1215, opened 2022-11-07
ConversionFeeDenom param missing from GetParams in MsgFees module

nullpointer0x00:
## Description

The ConversionFeeDenom is missing in the GetParams() method of the msgfees module.

closes: #1214

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #1221, opened 2022-11-09
Fix GetParams in `msgfees` modules to return ConversionFeeDenom (#1215)

nullpointer0x00:
## Description

Adds `ConversionFeeDenom` to get params and adds required setting in the upgrade handler for `ochre`

closes: #XXXX

---

write a little note why.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -40,6 +40,8 @@ var handlers = map[string]appUpgrade{
 	"ochre-rc1": { // upgrade for 1.13.0-rc1
 		Added: []string{group.ModuleName, rewardtypes.ModuleName, icacontrollertypes.StoreKey, icahosttypes.StoreKey},
 		Handler: func(ctx sdk.Context, app *App, plan upgradetypes.Plan) (module.VersionMap, error) {
+			params := app.MsgFeesKeeper.GetParams(ctx)
+			app.MsgFeesKeeper.SetParams(ctx, params)
 			versionMap := app.UpgradeKeeper.GetModuleVersionMap(ctx)
 			UpgradeICA(ctx, app, &versionMap)
 			ctx.Logger().Info("Starting migrations. This may take a significant amount of time to complete. Do not restart node.")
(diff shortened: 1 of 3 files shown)


arnabmitra (approved):

iramiller (dismissed):

app/upgrades.go
```diff
@@ -40,6 +40,8 @@ var handlers = map[string]appUpgrade{
 	"ochre-rc1": { // upgrade for 1.13.0-rc1
 		Added: []string{group.ModuleName, rewardtypes.ModuleName, icacontrollertypes.StoreKey, icahosttypes.StoreKey},
 		Handler: func(ctx sdk.Context, app *App, plan upgradetypes.Plan) (module.VersionMap, error) {
+			params := app.MsgFeesKeeper.GetParams(ctx)
+			app.MsgFeesKeeper.SetParams(ctx, params)
```
SpicyLemon, line 44: `ochre-rc1` has already run on testnet, so this ends up doing nothing.

We can't just change `ochre-rc1` to `ochre` for the mainnet upgrade either. Doing so would make `provenanced` not work on testnet anymore because it'd be missing the upgrade handler.

So... This either needs to be added to a new `ochre-rc2` upgrade handler (if we plan on doing another rc upgrade in testnet) or else we need to copy the existing `ochre-rc1` handler to a new `ochre` one, and add these lines there.

If we're going to skip an `-rc2` upgrade, we should probably have someone put in a gov proposal on testnet to fix the params on there.

arnabmitra: i agree with @SpicyLemon we need to take this out of `ochre-rc1` and add it to `ochre` (which will be added when we are ready to release to mainnet imo. We should take out those two lines ^^

nullpointer0x00: Agree, I was thinking we would change `ochre-rc1` to `ochre`...but that is wrong.

@SpicyLemon we already did a gov proposal for the fix before this PR.  So, we are good on that front.

SpicyLemon (approved):

Issue #1257, opened 2022-12-09
DBMigrator tool cannot convert cleveldb -> goleveldb


Taztingo:
## Summary of Bug

When attempting to convert a `cleveldb` to a `goleveldb`, the tool does not do any conversion. It detects an incorrect source database type and thinks the source is of type `goleveldb`. It then attempts to convert `goleveldb` -> `goleveldb`. Since they are the same type, no work is performed.

## Version

1.13.0

## Steps to Reproduce

1. Have your home directory use cleveldb for data.
2. Have your home directory's config.toml configured for cleveldb.
3. Run the dbmigrator for goleveldb. `dbmigrate goleveldb --home xyz`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1258, opened 2022-12-09
Updated dbmigrate to obtain source db type from config.toml

Taztingo:
## Description

It was hard for dbmigrate to differentiate between goleveldb and cleveldb. I updated the code to take the db source type from the config.toml instead of attempting to detect it.

closes: #1257 

---

write a little note why.


nullpointer0x00 (changes requested):
Need to add change log.

cmd/dbmigrate/cmd/dbmigrate.go
```diff
@@ -98,10 +98,17 @@ Migration process:
 			if err != nil {
 				return fmt.Errorf("could not parse --%s option: %w", FlagBatchSize, err)
 			}
+
+			sourceDB, err := ReadDBBackendConfigValue(command)
```
nullpointer0x00, line 102: We should probably explain this is what is happening in the command Long section?  That it using the configured database and converting it to new source.   Not sure if we should also mention if the configured database is incorrect the process will not convert it correctly...thoughts?

Taztingo: Agreed, that should be added.

cmd/dbmigrate/cmd/dbmigrate.go
```diff
@@ -73,6 +73,10 @@ Migration process:
    and by default will be in the {home} directoyr.
 3. Move the staging data directory into place as the current data directory.
 4. Update the config file to reflect the new database backend type.
+
+Source Database Type:
+The source database type is read from the config.toml within the home directory. If this does not match the database types, then the conversion
+process will not work properly.
```
SpicyLemon, line 79: To address this, I'd rather add a `--source-db-backend` flag to the command that could be provided to the migrator. If it's provided, use that for all dbs, otherwise, auto-detect for each.

I didn't add this originally because there's no guarantee that all dbs in the directory have the same backend. Also because the `application` and `snapshots` db backend types were originally defined using a set-at-compile-time `DBBackend` variable; the`dbmigrator` had no way of knowing what that value was.

nullpointer0x00 (approved):

arnabmitra (approved):

arnabmitra (dismissed):

cmd/dbmigrate/utils/migrator.go
```diff
@@ -455,16 +460,19 @@ func (m *migrationManager) MigrateDBDir(dbDir string) (summary string, err error
 		}
 		// always wrap any error with some extra context.
 		if err != nil {
-			err = fmt.Errorf("could not convert %q from %q to %q: %w", dbDir, sourceDBType, m.TargetDBType, err)
+			err = fmt.Errorf("could not convert %q to %q: %w", dbDir, m.TargetDBType, err)
```
SpicyLemon, line 463: Please keep the source db type in this error message. It's a crucial piece of info when troubleshooting. I know it might be in the log output somewhere, but it might not. And having it here means we don't have to dig through the logs for it.

Taztingo: Added it back.

iramiller (approved):

cmd/dbmigrate/utils/migrator.go
```diff
@@ -460,11 +466,14 @@ func (m *migrationManager) MigrateDBDir(dbDir string) (summary string, err error
 	}()
 
 	m.Status = "detecting db type"
-	var ok bool
-	sourceDBType, ok = DetectDBType(dbName, sourceDir)
+	sourceDBType, ok := tmdb.BackendType(m.Migrator.SourceDBType), true
```
SpicyLemon, line 469:
```suggestion
	var ok bool
	sourceDBType, ok = tmdb.BackendType(m.Migrator.SourceDBType), true
```
Using `:=` here creates a new var which shadows the one above that would get used in the error.

SpicyLemon: Nevermind. All good: https://go.dev/play/p/W_hgqMD6M9f?v=goprev
https://go.dev/ref/spec#Short_variable_declarations
It's not creating a new/shadow var because `sourceDBType` is defined earlier in the same block, and go handles it.

SpicyLemon (approved):
lgtm

Issue #1004, opened 2022-08-12
Name Module Modifications via Governance Proposal


hyperschwartz:
## Summary

This request is to create a new governance proposal type to change a name's owner account and restricted value.

## Problem Definition

Each feature discussed below would be a configurable field/field-set in the prospective governance proposal.

* __Ownership change__:  This feature would enable a name to be owned by a new account without the inherent security issues that would be presented if it were to be available outside of a governance proposal.  This is very useful in the case of a name being erroneously created under an incorrect account, or if an account's keys are lost.    Due to sub-names being owned by separate accounts unrelated to the parent name, changing a name's owner should not require a ripple effect to sub-names.  The obvious downside to this feature is that a name can effectively be "stolen" if the entire community agrees on the choice to do so.

* __Restriction change__: This feature would enable a name's restricted status to be toggled between its states.  This would allow for an erroneously-declared name to properly allow sub-names, or for sub-names to be denied after the block in which this change is written.  This might present a strange set of states, allowing restricted parent names to have sub-names.  As stated in the previous change, this likely shouldn't be an issue, but it could lead to some unexpected states based on assumptions of the current name module's code about when sub-names can exist.

## Proposal

* This set of changes would require a new governance proposal message to be created that allows the specification of a new target owner and a new restricted status for a name.  Validation would ensure that one or both fields are set, and reject a request that effectively changes nothing.
* Name module modifications to change the target values after a successful proposal would need to be added that can modify restricted and owner values for a name.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1266, opened 2022-12-14
Add modify name proposal

Taztingo:
## Description

Add a proposal to modify the name and restricted flag of a name.

closes: #1004

---

write a little note why.


proto/provenance/name/v1/name.proto
```diff
@@ -47,6 +47,20 @@ message CreateRootNameProposal {
...
+message ModifyNameProposal {
+  option (gogoproto.equal)            = false;
+  option (gogoproto.goproto_getters)  = false;
+  option (gogoproto.goproto_stringer) = false;
+
+  string title       = 1;
+  string description = 2;
+  string name        = 3;
+  string owner       = 4;
+  bool   restricted  = 5;
```
SpicyLemon, line 61: Could use some field comments here to indicate what's the supposed to be old/existing, and which fields are new to be applied.

x/name/client/cli/gov.go
```diff
@@ -106,7 +110,93 @@ $ %s tx gov submit-proposal param-change tx gov submit-proposal \
 		},
 	}
 
-	cmd.Flags().String(FlagOwner, "", "The owner of the new name, optional (defaults to from address")
+	cmd.Flags().String(FlagOwner, "", "The owner of the new name, optional (defaults to from address)")
+	cmd.Flags().BoolP(flagRestricted, "r", true, "Restrict creation of child names to owner only, optional (default false)")
```
SpicyLemon, line 114: This `flagRestricted` needs some love both here and in the new `GetModifyNameProposalCmd ` function.

If you run `provenanced tx gov submit-legacy-proposal root-name-proposal --help`, you'll see this in the flags listing:
```
  -r, --restrict                 Restrict creation of child names to owner only, optional (default false) (default true)
```

1. The "(default false)" in the description is just flat out incorrect. And boolean flags should always be optional, so the whole `", optional (default false)"` part should be removed.
2. Cobra automatically adds a "(default <val>)" when it's something of interest (e.g not `""`).
3. Boolean flags should default to `false`. The absence of the flag as an arg should be the default behavior. The presence of a flag as an arg should change that behavior. When you default a boolean flag to `true`, that gets messed up; absence of the flag is now the same as providing it. In order to override the default, it would have to be provided as `--restricted=false` (you can't even do `--restricted false`).

TL;DR (too late!):
This should either default to `false`, or, if the same current behavior is desired, should be renamed to `--unrestricted` so that it can have a `false` default value. Either way the description should be cleaned up too.

See also: The `--iavl-disable-fastnode` debacle.

x/name/client/cli/gov.go
```diff
@@ -106,7 +110,93 @@ $ %s tx gov submit-proposal param-change tx gov submit-proposal \
...
+func GetModifyNameProposalCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "modify-name-proposal [name] (--owner [address]) (--restrict) [flags]",
+		Short: "Submit a modify name creation governance proposal",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Submit a modify name governance proposal along with an initial deposit.
+The proposal title and description must be provided through their respective flags.
+
+IMPORTANT: The creation of sub-names will be enabled by default unless the restrict flag is included.
+When included the owner must approve all child name creation unless an alterate owner is provided.
```
SpicyLemon, line 131: The `--restricted` flag is added to the command with a default of `true` (see other comment), so this blurb is inaccurate.

x/name/spec/04_events.md
```diff
@@ -18,3 +19,13 @@ The name module emits the following events:
 | --------------------- | --------------------- | ------------------------- |
 | name_unbound          | name                  | {NameRecord|Name}         |
 | name_unbound          | address               | {NameRecord|Address}      |
+| name_unbound          | restricted            | {NameRecord|Restricted}   |
+
+
+### MsgBindNameRequest
```
SpicyLemon, line 25:
```suggestion
### ModifyNameProposal
```

Also, does there need to be a section in here about events from a root name proposal?

SpicyLemon (commented):
Couple small things, but mostly looks good.

x/name/client/cli/gov.go
```diff
@@ -27,6 +27,10 @@ const (
 	FlagDeposit string = "deposit"
 )
 
+func GetProposalCmd() *cobra.Command {
+	return nil
+}
+
```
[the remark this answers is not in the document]
Taztingo: Whoops, forgot to remove this.

proto/provenance/name/v1/name.proto
Taztingo: Agreed

x/name/client/cli/gov.go
Taztingo: I cleaned up the help text for this. Also, renamed the flag to `--unrestrict` in the BindName request, RootName proposal, and in this new proposal. I also made sure to mention this in the README as well.

x/name/spec/04_events.md
Taztingo: Yes, it does emit an event. So I added it.

x/name/keeper/keeper.go
```diff
@@ -291,3 +294,34 @@ func isValidUUID(s string) bool {
 	}
 	return true
 }
+
+func (keeper Keeper) addRecord(ctx sdk.Context, name string, addr sdk.AccAddress, restrict, isModifiable bool) error {
+	key, err := types.GetNameKeyPrefix(name)
```
iramiller, line 299: I noticed that we still have amino in this module (albeit as a deprecated legacy compatibility path) ... this would be a good time to clean that up with a module upgrade method.  I am not sure how much longer amino will be present at all.

Taztingo: I created a separate issue for this since the removal of legacy amino code will affect the attribute module as well.

x/name/client/cli/gov.go
```diff
@@ -106,8 +106,94 @@ $ %s tx gov submit-proposal param-change tx gov submit-proposal \
...
+IMPORTANT: The restricted creation of sub-names will be enabled by default unless the unrestricted flag is included.
+The owner must approve all child name creation unless an alterate owner is provided.
+
+Example:
+$ %s tx gov submit-legacy-proposal \
+	modify-name-proposal \
+	<name> \
+	--unrestrict  \ 
+	--owner <key_or_address> \
```
SpicyLemon, line 134: Nit: The keyring isn't referenced for this field, so this should just be `<address>`:
```suggestion
	--owner <address> \
```
It's also incorrectly stated as `<key_or_address>` in `GetRootNameProposalCmd`.

The `--from` field DOES use the keyring though, so it can stay as `<key_or_address>`.

proto/provenance/name/v1/name.proto
```diff
@@ -40,11 +40,35 @@ message CreateRootNameProposal {
...
+  string name = 3;
+  // the address the name will resolve to
+  string owner = 4;
+  // a flag that indicates if an owner signature is required to add sub-names
+  bool restricted = 5;
+}
+
+// ModifyNameProposal details a proposal to update the address
+// and/or the restricted flag of a name.
+message ModifyNameProposal {
```
SpicyLemon, line 57: I feel like we should do this as a `gov.v1`-style msg proposal instead of as `gov.v1beta1` content.

I know @llama-del-rey is working on updating our governance proposals to the `gov.v1` style, e.g. #1288. If we just do that right off the bat with this one, we don't end up with as many deprecated proto messages.

Does anyone else have thoughts on this?

Taztingo: Sure, I can definitely convert it to that.

x/name/keeper/keeper.go
```diff
@@ -291,3 +294,34 @@ func isValidUUID(s string) bool {
...
+		return err
+	}
+
+	store := ctx.KVStore(keeper.storeKey)
+	if store.Has(key) && !isModifiable {
+		return types.ErrNameAlreadyBound
+	}
+
+	record := types.NewNameRecord(name, addr, restrict)
+	if err = record.ValidateBasic(); err != nil {
```
nullpointer0x00, line 310: I'm wondering if we should start renaming these to just `Validate`.   Since, `ValidateBasic` is supposed to be for sdk.Msg types?

SpicyLemon: While it'd be nice to do that, it's probably better left for a PR doing only that for any applicable `ValidateBasic` functions.

x/name/keeper/msg_server.go
```diff
@@ -148,3 +149,28 @@ func (s msgServer) DeleteName(goCtx context.Context, msg *types.MsgDeleteNameReq
 
 	return &types.MsgDeleteNameResponse{}, nil
 }
+
+// ModifyName updates an existing name
+func (s msgServer) ModifyName(goCtx context.Context, msg *types.MsgModifyNameRequest) (*types.MsgModifyNameResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if msg.GetAuthority() != s.Keeper.GetAuthority() {
+		return nil, sdkerrors.ErrInvalidRequest.Wrap(govtypes.ErrInvalidSigner.Error())
```
SpicyLemon, line 158:
```suggestion
		return nil, sdkerrors.ErrInvalidRequest.Wrap(govtypes.ErrInvalidSigner.Wrapf("expected %s got %s", s.Keeper.GetAuthority(), msg.GetAuthority()))
```

x/name/spec/03_messages.md
```diff
@@ -52,6 +52,30 @@ This message is expected to fail if:
...
+  // The address signing the message
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // The record being updated
+  NameRecord record = 2 [(gogoproto.nullable) = false];
+}
+```
+
+This message is expected to fail if:
+- Any components of the request do not pass basic integrity and format checks
+- The record to remove does not exist
```
SpicyLemon, line 73:
```suggestion
- The record to update does not exist
```

x/name/spec/03_messages.md
```diff
@@ -52,6 +52,30 @@ This message is expected to fail if:
...
+- The requestor does not match the owner listed on the record.
```
SpicyLemon, line 74:
```suggestion
```

This is inaccurate.

x/name/types/msgs.go
```diff
@@ -101,3 +103,34 @@ func (msg MsgDeleteNameRequest) GetSigners() []sdk.AccAddress {
...
+}
+
+// ValidateBasic runs stateless validation checks on the message.
+func (msg MsgModifyNameRequest) ValidateBasic() error {
+	if strings.TrimSpace(msg.Record.Name) == "" {
+		return fmt.Errorf("name cannot be empty")
+	}
+	if strings.TrimSpace(msg.Record.Address) == "" {
+		return fmt.Errorf("address cannot be empty")
+	}
```
SpicyLemon, line 122: Should check if the `msg.Record.Address` is a valid Bech32 instead of just checking for empty/whitespace.
```suggestion
	if _, err := sdk. AccAddressFromBech32(msg.Record.Address); err != nil {
		return fmt.Errorf("invalid record address: %w", err)
	}
```
For the authority, though, it's okay to just do the empty string check since it's only ever used as a string.

x/name/types/proposal.go
```diff
@@ -11,13 +11,15 @@ import (
...
 	ProposalTypeCreateRootName = "CreateRootName"
+	ProposalTypeModifyName     = "ModifyName"
 )
 
 // Assert CreateRootNameProposal implements govtypesv1beta1.Content at compile-time
 var _ govtypesv1beta1.Content = &CreateRootNameProposal{}
 
 func init() {
 	govtypesv1beta1.RegisterProposalType(ProposalTypeCreateRootName)
+	govtypesv1beta1.RegisterProposalType(ProposalTypeModifyName)
```
SpicyLemon, line 22: Don't need this since it doesn't use the v1beta1 stuff.

derekadams (approved):
LGTM

SpicyLemon (approved):

derekadams (approved):

Issue #1268, opened 2022-12-15
Add bips field to MsgAssessCustomMsgFeeRequest and use it in provenance


arnabmitra:
## Summary
Based on this https://github.com/provenance-io/provenance/issues/1263  it may still be good to provide a way to distribute fees between recipient and module on a more dynamic basis

## Problem Definition
Based on this https://github.com/provenance-io/provenance/issues/1263 
we should probably  let the executor of the fee accessor have the ability to distribute any additional fee to the ecosystem
i.e add a `string bips      = 5; // bips the recipient gets` field to `MsgAssessCustomMsgFeeRequest`
```
// MsgAssessCustomMsgFeeRequest defines an sdk.Msg type
message MsgAssessCustomMsgFeeRequest {
  option (gogoproto.equal)            = false;
  option (gogoproto.goproto_getters)  = false;
  option (gogoproto.goproto_stringer) = true;

  string name = 1; // optional short name for custom msg fee, this will be emitted as a property of the event
  cosmos.base.v1beta1.Coin amount = 2 [(gogoproto.nullable) = false]; // amount of additional fee that must be paid
  string recipient = 3; // optional recipient address, the total amount is given to the recipient, if present.
  string from      = 4; // the signer of the msg
  string bips      = 5; // bips the recipient gets

}
```
that maybe a better option

## Proposal

Change proto as defined ^^.
Change code to read that field.
Change provwasm and other clients that consume it.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #1263, opened 2022-12-14
MsgFees Module 50/50 Fee Split on MsgAssessCustomMsgFeeRequest Removal


hyperschwartz:
## Summary

The current process for assessing a custom fee via the `MsgAssessCustomMsgFeeRequest` msg takes 50% of the amount charged and sends it to Provenance.  This fee should be removed.

## Problem Definition

This is a planned change, and this issue is to ensure that this change is tracked.

## Proposal

The code that currently subtracts 50% of the fees sent should be removed, and the comments on the `MsgAssessCustomMsgFeeRequest` in the proto definition stating that this behavior in place should be removed.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

nullpointer0x00:
Was finished with PR https://github.com/provenance-io/provenance/pull/1267

PR #72, opened 2021-02-16
Update the ScopeSpecification proto

dwedul-figure:
## Description

Change the ScopeSpecification proto to better represent what's needed.

closes: #71

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting))
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] ~~Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)~~ - n/a: only updating data models.
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Review `Codecov Report` in the comment section below once CI passes

PR #1285, opened 2023-01-06
Add bips field to MsgAssessCustomMsgFeeRequest 

nullpointer0x00:
## Description

Adding recipient basis point field to MsgAssessCustomMsgFeeRequest.  

When recipient is set but basis points in not set it will default to 10,000 (100% of fee).  

Currently, the split amount is 5,000(50:50) in versions <= 1.13.x.  The default was changed in issue #1263  

closes: #1268 

---

write a little note why.


client/docs/swagger-ui/swagger.yaml
```diff
@@ -67080,6 +67080,11 @@ definitions:
   provenance.marker.v1.MsgAddAccessResponse:
     type: object
     title: MsgAddAccessResponse defines the Msg/AddAccess response type
+  provenance.marker.v1.MsgAddFinalizeActivateMarkerResponse:
```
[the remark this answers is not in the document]
nullpointer0x00: Yes, the docs must have not been updated when that branch was merged to main.

docs/proto-docs.md
```diff
@@ -5636,8 +5636,9 @@ MsgAssessCustomMsgFeeRequest defines an sdk.Msg type
 | ----- | ---- | ----- | ----------- |
 | `name` | [string](#string) |  | optional short name for custom msg fee, this will be emitted as a property of the event |
 | `amount` | [cosmos.base.v1beta1.Coin](#cosmos.base.v1beta1.Coin) |  | amount of additional fee that must be paid |
-| `recipient` | [string](#string) |  | optional recipient address, the total amount is given to the recipient, if present. |
+| `recipient` | [string](#string) |  | optional recipient address, the basis points amount is sent to the recipient |
 | `from` | [string](#string) |  | the signer of the msg |
+| `recipient_basis_points` | [string](#string) |  | optional basis points 0 - 10,000 for recipient defaults to 10,000 |
```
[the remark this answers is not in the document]
nullpointer0x00: Yes, this allows for use with our default value.  It it isn't set we use the internal default value for assess msg bips.  This also allows for backwards compatibility with contracts that already use the msg.  They will not have to change anything to their contracts to keep the same behavior.  If it was set to a uint or int it would have defaulted to 0 when the proto is used.  I might be missing a better impl here?

x/msgfees/types/msgs.go
```diff
@@ -50,9 +55,29 @@ func (msg MsgAssessCustomMsgFeeRequest) ValidateBasic() error {
 	if !msg.Amount.IsPositive() {
 		return errors.New("amount must be greater than zero")
 	}
+	_, err = msg.GetBips()
+	if err != nil {
```
egaxhaj, line 59: nit: since the first argument is never used you can combine this into a one liner. Same goes for the other occurrences in the method where the first arg is never used.

```suggestion
	if _, err := msg.GetBips(); err != nil {
```

x/msgfees/types/msgs.go
```diff
@@ -50,9 +55,29 @@ func (msg MsgAssessCustomMsgFeeRequest) ValidateBasic() error {
...
 }
 
+// GetBips converts the msg RecipientBasisPoints to a uint32 basis point value 0 - 10,000
+func (msg MsgAssessCustomMsgFeeRequest) GetBips() (uint32, error) {
+	if msg.RecipientBasisPoints == "" {
+		return AssessCustomMsgFeeBips, nil
+	}
+	bips, err := strconv.ParseUint(msg.RecipientBasisPoints, 10, 32)
+	if err == nil {
+		if bips > 10_000 {
```
egaxhaj, line 72: `10_000` should probably be a constant `RECIPIENT_MAX_BIPS` or something like that.

x/msgfees/types/msgs.go
```diff
@@ -50,9 +55,29 @@ func (msg MsgAssessCustomMsgFeeRequest) ValidateBasic() error {
...
 		return errors.New("amount must be greater than zero")
 	}
+	_, err = msg.GetBips()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
+// GetBips converts the msg RecipientBasisPoints to a uint32 basis point value 0 - 10,000
```
egaxhaj, line 65: This goes in combination with the comment I made on line `#72`. Not certain if there's a need to mention the value ranger in the comments. If so, it should be something like  `0 - RECIPIENT_MAX_BIPS` or something like that.

x/msgfees/types/msgs.go
```diff
@@ -50,9 +55,29 @@ func (msg MsgAssessCustomMsgFeeRequest) ValidateBasic() error {
...
+func (msg MsgAssessCustomMsgFeeRequest) GetBips() (uint32, error) {
+	if msg.RecipientBasisPoints == "" {
+		return AssessCustomMsgFeeBips, nil
+	}
+	bips, err := strconv.ParseUint(msg.RecipientBasisPoints, 10, 32)
+	if err == nil {
+		if bips > 10_000 {
+			return 0, fmt.Errorf("recipient basis points can only be between 0 and 10,000 : %v", msg.RecipientBasisPoints)
```
egaxhaj, line 73: hardcoded max value in comments.

x/msgfees/types/msgs.go
egaxhaj: Come to think of it. Could this be better improved as a default parameter and then we can set min max through params?

x/msgfees/wasm/encode.go
```diff
@@ -28,8 +28,10 @@ type AssessCustomFeeParams struct {
 	From string `json:"from"`
 	// An optional short name
 	Name string `json:"name,omitempty"`
-	// An optional address to receive the fees. if present, the total amount is sent to the address.
+	// An optional address to receive the fees. if present, the split amount from basis points is sent to address.
 	Recipient string `json:"recipient,omitempty"`
+	// An optional recipient basis points (0 - 10,000). if not present, defaults to 10,000
```
egaxhaj, line 33: same thing here as mentioned in my previous comments about using a constant/default param.

x/msgfees/types/msgs.go
```diff
@@ -50,9 +55,29 @@ func (msg MsgAssessCustomMsgFeeRequest) ValidateBasic() error {
...
+		}
+	}
+
+	return uint32(bips), err
+}
+
+// GetSignBytes encodes the message for signing
```
egaxhaj, line 80: This comment looks to be reversed with the comment on `GetSignerBytes` method in `ln:89`.
```suggestion
// GetSigners indicates that the message must have been signed by the address provided
```

x/msgfees/types/msgs.go
nullpointer0x00: I'm not sure if we need to have a variable for this.  `RECIPIENT_MAX_BIPS` would imply that we are capping the amount of bips a recipient can have.  Here we are allowing the user of the assess msg to freely set how much they want.  The only possible range of basis points is 0 - 10,000 (0% - 100%).

x/msgfees/types/msgs.go
nullpointer0x00: Along the lines as above.  This also makes it more readable without the user having to refer to what a constant is set at.

x/msgfees/types/msgs.go
nullpointer0x00: Basis points should never be anything other than 0 - 10,000.  Having it hard coded allows me to format it as easily format it as `10,000`.

x/msgfees/wasm/encode.go
nullpointer0x00: See comment above.

x/msgfees/types/msgs.go
egaxhaj: 👍

x/msgfees/types/msgs.go
egaxhaj: 👍

x/msgfees/wasm/encode.go
egaxhaj: 👍

x/msgfees/types/msgs.go
egaxhaj: 👍

x/msgfees/types/msgs.go
nullpointer0x00: I copied and pasted this from how we have it for 5 the other modules (25 total times). I would rather keep it consistent with what we already have in our other modules.  So, I think we change all of them now or do it in some other PR?

x/msgfees/types/msgs.go
nullpointer0x00: True.  I changed the others in the validate basic as well to fit the same pattern.

x/msgfees/types/msgs.go
egaxhaj: If it exists in other modules as well then we'll need to do it in a separate PR.

egaxhaj: I created an issue for this.

iramiller:
One small comment related to the comments that note that BIPS is between 0-10_000 .... technically this is an inclusive range but the "between" wording implies an exclusive range ...

egaxhaj (approved):
LGTM

derekadams (approved):
LGTM

PR #1286, opened 2023-01-06
Llama 1283 create root name

ghost:
## Description
Not fully sure how to go about making it compatible with the legacy proposal handlers. Opening a draft to get some feedback

closes: #XXXX

---

write a little note why.

### DIFF
--- a/x/name/wasm/encode.go
+++ b/x/name/wasm/encode.go
@@ -23,7 +23,7 @@ type NameMsgParams struct {
 	// Encode a MsgUnBindName
 	Delete *DeleteNameParams `json:"delete_name,omitempty"`
 	// Encode a MsgCreateRootName
-	CreateRoot *CreateRootNameParams `json:create_root_name,omitempty`
+	CreateRoot *CreateRootNameParams `json:"create_root_name,omitempty"`
 }
 
 // BindNameParams are params for encoding a MsgBindName.


proto/provenance/name/v1/tx.proto
```diff
@@ -46,3 +49,18 @@ message MsgDeleteNameRequest {
 
 // MsgDeleteNameResponse defines the Msg/DeleteName response type.
 message MsgDeleteNameResponse {}
+
+// MsgCreateRootNameProposal defines an sdk.Msg type to create a new root name
+// that is controlled by a given owner and optionally restricted to the owner
+// for the sole creation of sub names.
+message MsgCreateRootNameRequest {
```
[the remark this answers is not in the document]
SpicyLemon: Gotta have a new message primarily so that the authority can be provided and checked. Relatedly, the "fromAddress" here should be "authority" to match other gov prop messages as well as what it's checked against in the keeper.

x/name/handler.go
```diff
@@ -23,6 +23,9 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
 		case *types.MsgDeleteNameRequest:
 			res, err := msgServer.DeleteName(sdk.WrapSDKContext(ctx), msg)
 			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgCreateRootNameRequest:
+			res, err := msgServer.CreateRootName(sdk.WrapSDKContext(ctx), msg)
+			return sdk.WrapServiceResult(ctx, res, err)
```
SpicyLemon, line 28: This isn't needed since we're using the gov `v1` stuff (instead of `v1beta`).

x/name/keeper/msg_server.go
```diff
@@ -148,3 +150,26 @@ func (s msgServer) DeleteName(goCtx context.Context, msg *types.MsgDeleteNameReq
...
+	}
+
+	// Validate
+	if err := msg.ValidateBasic(); err != nil {
+		ctx.Logger().Error("unable to validate message", "err", err)
+		return nil, sdkerrors.ErrInvalidRequest.Wrap(err.Error())
+	}
+
+	//TODO: route to existing handler?
+	err := HandleCreateRootNameProposal(ctx, s.Keeper, types.NewCreateRootNameProposal(msg.Title, msg.Description, msg.Name, sdk.AccAddress(msg.Owner), msg.Restricted))
```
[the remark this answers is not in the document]
SpicyLemon: Looks like that `HandleCreateRootNameProposal` function has all the logic/processing needed, and doesn't do anything else. So calling it like this is okay. It doesn't use the `Title` or `Description` fields either, so you can provide those as`""` and be okay.

x/name/types/msgs.go
```diff
@@ -101,3 +103,65 @@ func (msg MsgDeleteNameRequest) GetSigners() []sdk.AccAddress {
...
+		Description: description,
+		Name:        name,
+		Owner:       owner,
+		Restricted:  restricted,
+		FromAddress: fromAddress,
+	}
+}
+
+// Route implements Msg
+func (msg MsgCreateRootNameRequest) Route() string { return ModuleName }
```
SpicyLemon, line 120: The only funcs you need to implement for the `Msg` interface are:
```golang
ValidateBasic() error
GetSigners() []AccAddress
```
(Also `String() string` if that's not created automatically from the proto).

But you no longer need `Route`, `Type`, or `GetSignBytes`.

Since we don't need to implement v1beta1 `Content` either, you also don't need:
```golang
GetTitle() string
GetDescription() string
ProposalRoute() string
ProposalType() string
```

x/name/wasm/encode.go
```diff
@@ -22,6 +22,8 @@ type NameMsgParams struct {
 	Bind *BindNameParams `json:"bind_name,omitempty"`
 	// Encode a MsgUnBindName
 	Delete *DeleteNameParams `json:"delete_name,omitempty"`
+	// Encode a MsgCreateRootName
+	CreateRoot *CreateRootNameParams `json:create_root_name,omitempty`
```
SpicyLemon, line 26: Missing the `"` around the stuff.
```suggestion
	CreateRoot *CreateRootNameParams `json:"create_root_name,omitempty"`
```

PR #1288, opened 2023-01-09
1283 Migrate CreateRootNameProposal to v1

ghost:
## Description

closes: #1283 

---

write a little note why.

### DIFF
--- a/x/name/types/name_test.go
+++ b/x/name/types/name_test.go
@@ -52,7 +52,7 @@ func (s *NameRecordTestSuite) TestNameRecordValidateBasic() {
 		tc := tc
 
 		s.Run(n, func() {
-			err := tc.name.ValidateBasic()
+			err := tc.name.Validate()
 			if tc.expectErr {
 				s.Error(err)
 				if s != nil {


iramiller:
This seems pretty good ... the Authority is in the signers list and also hard coded to require a match against the gov module account.

SpicyLemon (commented):
There's a couple small things to address, but conceptually, it looks good.

I'm not sure how these should be handled when it comes to wasm though.

x/name/types/msgs_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func TestMsgCreateRootNameRequestGetSignBytes(t *testing.T) {
```
SpicyLemon, line 11: I'm not saying these need to be refactored, but for future new tests, consider using [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests). Here's a `ValidateBasic` example from our metadata module: https://github.com/provenance-io/provenance/blob/bcc58b01f5ca00dacb2fb908fee9f374bef5eaf0/x/metadata/types/msg_test.go#L99-L148

Taztingo (changes requested):

x/name/keeper/msg_server.go
```diff
@@ -148,3 +152,27 @@ func (s msgServer) DeleteName(goCtx context.Context, msg *types.MsgDeleteNameReq
...
+// CreateRootName binds a name to an address
+func (s msgServer) CreateRootName(goCtx context.Context, msg *types.MsgCreateRootNameRequest) (*types.MsgCreateRootNameResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if s.Keeper.GetAuthority() != msg.Authority {
+		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "expected %s got %s", s.Keeper.GetAuthority(), msg.Authority)
+	}
+
+	// Validate
+	if err := msg.ValidateBasic(); err != nil {
```
Taztingo, line 165: This is redundant because ValidateBasic will be ran before this is called.

Taztingo (changes requested):

proto/provenance/name/v1/tx.proto
```diff
@@ -46,3 +51,18 @@ message MsgDeleteNameRequest {
...
 message MsgDeleteNameResponse {}
+
+// MsgCreateRootNameProposal defines an sdk.Msg type to create a new root name
+// that is controlled by a given owner and optionally restricted to the owner
+// for the sole creation of sub names.
+message MsgCreateRootNameRequest {
+  option (cosmos.msg.v1.signer)    = "authority";
+
+  string authority   = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string name        = 2;
```
Taztingo, line 62: I know we have a record object that includes these 3 fields. Maybe you want to use that instead.

Taztingo (dismissed):

x/name/keeper/keeper_test.go
```diff
@@ -320,3 +323,31 @@ func (s *KeeperTestSuite) TestSecp256r1KeyAlgo() {
...
+	s.Run("create valid root name", func() {
+		_, err := s.msgSrvr.CreateRootName(s.ctx, &msg)
+		s.Require().NoError(err)
+	})
+
+	s.Run("invalid authority", func() {
+		msg.Authority = "..."
+		_, err := s.msgSrvr.CreateRootName(s.ctx, &msg)
+		s.Require().Error(err)
+		s.Require().Equal("expected cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn got ...: expected gov account as only signer for proposal message", err.Error())
```
SpicyLemon, line 351: For future reference, these two can be combined using `EqualError`. There's also `ErrorContains` which is good to use when multiple things might be involved in an error's final message.

Feel free to resolve this comment.

SpicyLemon (dismissed):

x/name/wasm/encode.go
```diff
@@ -88,3 +98,9 @@ func (params *DeleteNameParams) Encode(contract sdk.AccAddress) ([]sdk.Msg, erro
 	msg := types.NewMsgDeleteNameRequest(record)
 	return []sdk.Msg{msg}, nil
 }
+
+// Encode creates a MsgCreateRootNameRequest.
+func (params *CreateRootNameParams) Encode(contract sdk.AccAddress) ([]sdk.Msg, error) {
+	msg := types.NewMsgCreateRootNameRequest(params.Authority, params.Name, params.Owner, params.Restricted)
+	return []sdk.Msg{msg}, nil
+}
```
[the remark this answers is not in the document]
SpicyLemon: I suggest we don't include this in the wasm stuff.

I feel like, if we did want an encoder for this message, it would create the `MsgCreateRootNameRequest`, put it in a `MsgSubmitProposal` and return that.

I just don't trust that it's not possible for wasm to execute one of these messages arbitrarily.

While it might end up that we want to have this as an option in the long run, I'd like to err on the side of caution and plan out what these should look like first.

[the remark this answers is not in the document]
iramiller: I agree... I don't see a use case for a smart contract to generate a governance proposal.

proto/provenance/name/v1/tx.proto
```diff
@@ -21,6 +21,9 @@ service Msg {
 
   // ModifyName defines a method to modify the attributes of an existing name.
   rpc ModifyName(MsgModifyNameRequest) returns (MsgModifyNameResponse);
+
+  // CreateRootName defines a governance method  for creating a root name.
```
nullpointer0x00, line 25: NIT: remove extra space between method and for.

x/name/types/msgs.go
```diff
@@ -84,6 +96,32 @@ func NewMsgModifyNameRequest(authority string, name string, owner sdk.AccAddress
...
+func (msg MsgCreateRootNameRequest) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
+		return ErrInvalidAddress
+	}
+
+	if _, err := sdk.AccAddressFromBech32(msg.Record.Address); err != nil {
+		return ErrInvalidAddress
+	}
+
+	if strings.TrimSpace(msg.Record.Name) == "" {
```
nullpointer0x00, line 108: Name record has a `ValidateBasic` method that does these checks. You could probably call that instead. It doesn't check the Address though.  Maybe it should?

While you are at it.  You could probably name the `ValidateBasic` method of NameRecord to `Validate`.  Since we just want to reserve `ValidateBasic` to types that implement the sdk.Msg interface.

nullpointer0x00 (commented):
We probably should update the spec.

SpicyLemon (approved):

nullpointer0x00 (approved):

Taztingo (dismissed):

SpicyLemon (approved):

nullpointer0x00 (approved):

PR #1293, opened 2023-01-10
Migrate SupplyIncreaseProposal

ghost:
## Description

closes: #1290

---

write a little note why.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -65,7 +65,7 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Publish to Docker Hub
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v4
         with:
           context: .
           target: run
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
 	golang.org/x/text v0.6.0
 	google.golang.org/genproto v0.0.0-20221227171554-f9683d7f8bef
-	google.golang.org/grpc v1.52.1
+	google.golang.org/grpc v1.52.3
 	google.golang.org/protobuf v1.28.2-0.20220831092852-f930b1dc76e8
 	gopkg.in/yaml.v2 v2.4.0
 
--- a/go.sum
+++ b/go.sum
@@ -1418,8 +1418,8 @@ google.golang.org/grpc v1.33.1/go.mod h1:fr5YgcSWrqhRRxogOsw7RzIpsmvOZ6IcH4kBYTp
 google.golang.org/grpc v1.33.2/go.mod h1:JMHMWHQWaTccqQQlmk3MJZS+GWXOdAesneDmEnv2fbc=
 google.golang.org/grpc v1.34.0/go.mod h1:WotjhfgOW/POjDeRt8vscBtXq+2VjORFy659qA51WJ8=
 google.golang.org/grpc v1.35.0/go.mod h1:qjiiYl8FncCW8feJPdyg3v6XW24KsRHe+dy9BAGRRjU=
-google.golang.org/grpc v1.52.1 h1:2NpOPk5g5Xtb0qebIEs7hNIa++PdtZLo2AQUpc1YnSU=
-google.golang.org/grpc v1.52.1/go.mod h1:pu6fVzoFb+NBYNAvQL08ic+lvB2IojljRYuun5vorUY=
+google.golang.org/grpc v1.52.3 h1:pf7sOysg4LdgBqduXveGKrcEwbStiK2rtfghdzlUYDQ=
+google.golang.org/grpc v1.52.3/go.mod h1:pu6fVzoFb+NBYNAvQL08ic+lvB2IojljRYuun5vorUY=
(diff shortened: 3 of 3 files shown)


arnabmitra:
In current code, cli would create this proposal, i.e user would not need to worry about the json and just the parameters, any thoughts on creating this for the new path? I might be missing where that is created?

x/marker/types/msg_test.go
```diff
@@ -245,3 +245,18 @@ func TestMsgAddFinalizeActivateMarkerRequestValidateBasic(t *testing.T) {
...
+	amount :=
+		sdk.Coin{
+			Amount: math.NewInt(100),
+			Denom:  "",
+		}
+
+	msg := NewMsgSupplyIncreaseProposalRequest(amount, targetAddress.String(), authority.String())
+	res := msg.GetSigners()
+	require.Equal(t, 1, len(res))
+	require.True(t, authority.Equals(res[0]))
```
SpicyLemon, line 261: ### Thoughts on `.Equal(... len(...))` vs `.Len(...)`

When `.Len(t, res, 1)` fails, the failure message indicates that the differing values are a slice length. It also contains the contents of the slice. This extra info can often help identify problems (e.g. what was unexpectedly added and what was unexpectedly removed). Still, it's best to also include a msg to indicate what slice is being evaluated.

When using `.Equal(t, 1, len(res))`, all the failure message contains is that an expected value was 1, and what the actual value was. There's no context. Including a msg can help with that, but it's clunkier than using `.Len(...)`.

Really, the only time to use `Equal(... len(...))` is when including the values in a failure message has very, very little chance of being helpful.

If you're expecting zero entries, the `.Empty(...)` assertion is the way to go.

### Thoughts on `.True(...)` or `.False(...)`

The ONLY time to ever use this is when you're checking the value of a boolean returned by something being tested. E.g. use it if you're making sure a value is in a map: `val, ok := myMap[foo]; assert.True(t, ok, "is foo in myMap")`. Or e.g. `val, found := k.GetEntry(ctx, 3); assert.True(t, found, "GetEntry(3) found")`.

When a `.True(...)` test fails, the only info in the failure message is that a value was false when it was expected to be true. Such a failure messages is almost never helpful unless there's significant amounts of extra context provided as a msg.

### For this specific case

It's best to define an expected slice, and just pass that slice to `.Equal` with a message indicating what the slice is.
```suggestion
	exp := []sdk.AccAddress{authority}
	require.Equal(t, exp, res, "GetSigners")
```
* Any failure here will indicate that the value came from GetSigners. Even though the name of the test indicates that GetSigners is what's being tested, it's good to have it included in the failure message so that a reader knows for sure that that's the values they're looking at.
* All expected and actual values are included in the failure message.
* Any missing entries are noted in the failure message.
* Any extra entries are noted in the failure message.
* If they're in the wrong order, that's identifiable from the contents of the failure message too.

x/marker/types/msg_test.go
```diff
@@ -245,3 +245,18 @@ func TestMsgAddFinalizeActivateMarkerRequestValidateBasic(t *testing.T) {
...
 	}
 }
+
+func TestMsgSupplyIncreaseProposalRequestGetSigners(t *testing.T) {
+	authority := sdk.AccAddress("input111111111111111")
+	targetAddress := sdk.AccAddress("input111111111111111")
+	amount :=
+		sdk.Coin{
+			Amount: math.NewInt(100),
+			Denom:  "",
```
SpicyLemon, line 255: Safest to give a denom here in the off chance that not having one causes an unrelated/unexpected problem.

SpicyLemon (commented):
A couple nits and a few things to address.

SpicyLemon:
> In current code, cli would create this proposal, i.e user would not need to worry about the json and just the parameters, any thoughts on creating this for the new path? I might be missing where that is created?

In our next SDK bump, we'll get the `AddGovPropFlagsToCmd` and `ReadGovPropFlags` helpers, which will greatly help facilitate this. I expect a few follow-ups at some point to add new CLI `tx` commands that generate and send proposals for different purposes (like this).

SpicyLemon (approved):

x/marker/keeper/keeper.go
```diff
@@ -176,3 +184,8 @@ func (k Keeper) IterateMarkers(ctx sdk.Context, cb func(marker types.MarkerAccou
 func (k Keeper) GetEscrow(ctx sdk.Context, marker types.MarkerAccountI) sdk.Coins {
 	return k.bankKeeper.GetAllBalances(ctx, marker.GetAddress())
 }
+
+// GetAuthority is signer of the proposal
+func (k Keeper) GetAuthority() string {
```
Taztingo, line 189: This is a pretty simple function, but it would be nice to have a test for it. Maybe something like what we have in the name module? The following code is what we have in the name module, and may need to be adjusted slightly:

```
func (s *KeeperTestSuite) TestGetAuthority() {
	s.Run("has correct authority", func() {
		authority := s.app.NameKeeper.GetAuthority()
		s.Require().Equal("cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn", authority)
	})
}
```

Taztingo (changes requested):

x/marker/keeper/keeper.go
Taztingo: Besides that, everything else looks really good :).

SpicyLemon (approved):

Taztingo (approved):

iramiller (approved):

Issue #1287, opened 2023-01-09
Migrate AddMarkerProposal to message based gov proposal


llama-del-rey:

SpicyLemon:
This was done with #1358

PR #1321, opened 2023-01-26
Migrate add marker gov proposal to v1 message based proposal

ghost:
## Description

closes: #1287

---

write a little note why.

### DIFF
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -168,7 +168,7 @@ func TestSimple(t *testing.T) {
 // /usr/local/go/bin/go test -benchmem -run=^$ github.com/provenance-io/provenance -bench ^BenchmarkFullAppSimulation$ -Commit=true -cpuprofile cpu.out
 func TestAppImportExport(t *testing.T) {
 	// uncomment to run in ide without flags.
-	//sdksim.FlagEnabledValue = true
+	sdksim.FlagEnabledValue = true
 
 	config, db, dir, logger, skip, err := sdksim.SetupSimulation("leveldb-app-sim", "Simulation")
 	if skip {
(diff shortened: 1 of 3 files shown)


x/marker/handler.go
```diff
@@ -73,8 +78,6 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
 func NewProposalHandler(k keeper.Keeper) govtypesv1beta1.Handler {
 	return func(ctx sdk.Context, content govtypesv1beta1.Content) error {
 		switch c := content.(type) {
-		case *types.AddMarkerProposal:
-			return keeper.HandleAddMarkerProposal(ctx, k, c)
```
SpicyLemon, line 77: This should probably be kept so that the old stuff continues to work. Then, in some later version, we can finally remove support for it.

nullpointer0x00: I initially agreed with your statement, but then I went back and looked at the old gov proposals on testnet and mainnet.  There hasn't been an AddMarkerProposal for a long time.  No one is actively using them.  So, I would say it is safe to remove.  When we start using it more there will just be the new way of doing the proposal.

x/marker/client/cli/tx.go
```diff
@@ -154,8 +154,6 @@ Valid Proposal Types (and associated parameters):
 			var proposal govtypesv1beta1.Content
 
 			switch args[0] {
-			case types.ProposalTypeAddMarker:
-				proposal = &types.AddMarkerProposal{}
 			case types.ProposalTypeIncreaseSupply:
```
nullpointer0x00, line 157: We should probably add a cli command to execute this proposal.  There is an example in the name module for  `ModifyNameProposal`.  This way we can set the authority as the gov module address easily.

PR #1367, opened 2023-02-14
Remove mac binary release artifact.

SpicyLemon:
## Description

This PR updates the release github action. The mac binary will no longer be signed, and it will no longer be part of the release artifacts or upgrade plan.

Reasons for this change:
1. The signing was being done by a account that should no longer be doing so.
2. The Provenance Blockchain Foundation doesn't yet have an Apple Developer ID, and the only reason it would get one is for this signing (there are no other Mac/IOS apps).
3. The mac binary is primarily for use by developers who can hopefully build from source without issues, which doesn't require signing.
4. While I can't say for certain that there aren't any validator nodes running the mac binary, I feel like that's a very good probability. MacOS servers aren't very common, even in the cloud world.
5. Fewer macs are running AMD chipsets these days. Github does not have standard ARM runners though, only AMD. So the compiled mac binary is for older hardware and might not even run on current hardware.

---

write a little note why.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,33 +65,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install osx tools
-        run: |
-          brew tap mitchellh/gon
-          brew install mitchellh/gon/gon
       - name: Setup go
         uses: actions/setup-go@v3
         with:
           go-version: ${{ needs.build_init.outputs.go_version }}
       - name: Build osx binary
         run: |
-          make WITH_CLEVELDB=false WITH_ROCKSDB=false VERSION=${{ needs.build_init.outputs.version }} build-release-bin build-release-libwasm
+          export VERSION=${{ needs.build_init.outputs.version }}
+          export WITH_CLEVELDB=false
+          export WITH_ROCKSDB=false
+          make build-release-zip
       - name: Provenanced version
         run: build/provenanced version --long
-      - name: Import Code-Signing Certificates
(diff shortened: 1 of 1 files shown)


iramiller (dismissed):

SpicyLemon:
Since the `update_release` ("Attach Release Artifacts") job only runs for releases, we won't be able to see it in action until we tag a release. One of the goals of this PR is to no longer include the mac binary as a release artifact, and that is done during this step.

Here's some notes about what's changing and what it means:
* The direct change to this job is to no longer download the mac binary (built in the `build_osx` job), and subsequently, not to upload it as a release artifact. The `build_osx` job will still compile the binary, zip it (with the libwasm library), and upload it as an artifact on the _action_, but it won't be listed as an artifact on the _release_.
* The [`build-release-checksum` build target](https://github.com/provenance-io/provenance/blob/dwedul/no-mac-artifact/Makefile#L271-L277) acts on all `.zip` files in `build/`. By not downloading the mac artifact before running it, it won't be included in the checksum file. That step will run just fine without the mac zip.
* The [`build-release-plan` target/script](https://github.com/provenance-io/provenance/blob/dwedul/no-mac-artifact/scripts/release-plan) is hard-coded to look for the `build/provenance-linux-amd64.zip` and `build/provenance-darwin-amd64.zip` files (downloaded earlier in the job). Since we're not downloading the darwin zip anymore, it won't exist during this step. When the script looks for it, and can't find it, it will just skip it. It therefore won't be included in the release plan.

So, long story short (too late!), the other parts of the release process should still behave as expected.

egaxhaj (approved):

Issue #1200, opened 2022-10-26
Fix use of deprecated ::set-output in github actions.


SpicyLemon:
## Summary

Find and fix all uses of the `::set-output` directive in Github actions.

## Problem Definition

Creating step output using by printing a line starting with `::set-output` has been deprecated as of Oct 11, 2022: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
It will stop working on May 31, 2023.

## Proposal

Find all uses of `::set-output` and replace them with output redirection to `$GITHUB_OUTPUT`.

As of writing this ticket, there are several uses of it in the `.github/workflow/*` files. Fixing those should be easy.

There is also a use of it in `scripts/protoball.sh`. There's a chance that use can be changed the same way as the others, but maybe not. Unfortunately, that script is only run by github actions during a release. So you might need to tweak that workflow a little bit at least to test that it's setting the variable appropriately in the github action. It's also possible that the variable it's creating isn't actually used (and those lines can be deleted from the script).

The `::save-state` directive has also been deprecated. As of writing this, we don't use that anywhere. But a check should be made again as part of resolving this issue to make sure nothing was added that uses it. Similarly, a check should be made for added uses of `::set-output` outside of what is listed here already.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1369, opened 2023-02-15
Replace set-output uses in github actions.

SpicyLemon:
## Description

closes: #1200 

Replace uses of the deprecated `set-output` github directive.

---

write a little note why.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -51,9 +51,13 @@ jobs:
           if [[ $VERSION =~ ^v[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
             TAGS="$TAGS,${DOCKER_IMAGE}:${VERSION}"
           fi
-          echo ::set-output name=version::${VERSION}
-          echo ::set-output name=tags::${TAGS}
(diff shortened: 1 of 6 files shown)


nullpointer0x00 (approved):

egaxhaj (approved):
LGTM

In a future update we should consider passing to `GITHUB_ENV` and reference variables directly in subsequent steps vs `needs.{ID}.outputs`. 

`echo "{environment_variable_name}={value}" >> $GITHUB_ENV`

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

SpicyLemon:
> LGTM
> 
> In a future update we should consider passing to `GITHUB_ENV` and reference variables directly in subsequent steps vs `needs.{ID}.outputs`.
> 
> `echo "{environment_variable_name}={value}" >> $GITHUB_ENV`
> 
> https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

Looking up variables from the github environment doesn't work in all situations. For example, you can't do it in the `if` clause for a step/job. They're also not shared between jobs, only steps. Writing to `$GITHUB_ENV` is only helpful for setting things up or passing info from one step to another in the same job; it can't be used to replace the `needs.{ID}.outputs` stuff.

Issue #1339, opened 2023-02-02
Protobuf flow changes and updates


SpicyLemon:
## Summary

We need to push the third_party protobuf stuff up more often so that we can properly update the `buf.lock` file allowing us to push our own protos up when we need to.

We also need to bring our protobuf related scripts and stuff up-to-date with more recent commands.

## Problem Definition

When we created `v1.14.0-rc1`, the protobuf push job failed. It failed because the `proto/buf.lock` file points to an old commit that doesn't have a file recently added. In order to update that file, though, we need to first push up our third_party stuff.

This highlights a major flow problem. In order to keep both the third_party and provenance stuff updated in `buf.build`, the third party stuff needs to be pushed up first, then we need to update the lock file before we can push up our stuff. Workflows can't be allowed to update files though, so it's something we need to do.

Additionally, `protoc` isn't a thing anymore and we need to not stop using it.

## Proposal

Update the github actions workflows related to the protobuf pushing. Currently, this is done in `releases.yaml` in its own job that's triggered when we tag a release.
1. Create a new job that pushes the third_party stuff any time a PR is merged into main that has changes in the third_party directory. This might be better placed in a new workflow just for that. 
2. Update the releases.yaml to only push the provenance protos. They should still only be pushed upon release.
3. See if there's a way to have a github action create an issue. If so, whenever this third_party stuff is pushed, have it create an issue to update the lock file. If not, update the release process documentation (in `CONTRIBUTING.md`) to include a step about update that lock file.

Also, take a look at what the SDK does for proto generation stuff and update our own scripts to be similar. They don't do the same thing with docs that we do though, so some research will be needed on how we can continue to have our documentation auto-generated from the protos.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1374, opened 2023-02-16
fix protobuf release steps

egaxhaj:
## Description

- Fixes third_party buf.build push steps to ensure that proto breaking checks don't break during releases.

closes: #1339

---

write a little note why.


egaxhaj:
Because Provenance Protobuf files have a dependency on the third_party Protobuf library, we first need to push the third_party to BSR (https://buf.build/provenance-io) repo and update third_party/proto/buf.lock manifest with the last commit hash. This will allow for Provenance Protobuf releases to be checked agains the latest third_party release.

This workflow requires a `Personal Access Token` and a `GPG` key be set up on a code owners' GitHub account. They'll be used to sign commits used in the `create pull request` workflow step. _It is recommend that these keys be set up `in addition` to any existing keys._

1. Create a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) (PAT) and paste it as a [secret](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) named `CPR_PAT` (CRP stands for Create Pull Request).
3. [Generate a GPG key](https://docs.github.com/en/github/authenticating-to-github/generating-a-new-gpg-key) `RSA 4096` and export the GPG private key as an ASCII armored version to your clipboard:

```
# macOS
gpg --armor --export-secret-key joe@foo.bar | pbcopy

# Ubuntu (assuming GNU base64)
gpg --armor --export-secret-key joe@foo.bar -w0 | xclip

# Arch
gpg --armor --export-secret-key joe@foo.bar | xclip -selection clipboard -I

# FreeBSD (assuming BSD base64)
gpg --armor --export-secret-key joe@foo.bar | xclip
```
Paste your clipboard as a [secret](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) named `GPG_PRIVATE_KEY` for example. Create another secret for the `GPG_PASSPHRASE`.

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+      - "v[0-9]+.[0-9]+.[0-9]+"     # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v1.0-rc1, v20.15.10-rc5
+
+# Concurrency is not defined in here because this job usually doesn't run
+# long enough to need cancelling, and it's okay for it to run multiple times.
+
+# Provenance publishes Protobuf content (including third_party dependencies) to buf.build.
+# Because of third_party dependencies declared in buf.yml, a buf.lock will be generated
+# containing the dependency manifest. Among other things, the buf.lock contains the
+# latest commit hash of the third_party library in the buf.build Provenance repository.
```
SpicyLemon, line 21: Probably worth noting that it's a `buf.build` commit hash (not one from this repo).

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+# When new files are added to the third_party library we need:
+#   1. publish the third_party library
+#   2. update buf.lock to point to the new commit hash
+#   3. sign and create a PR to commit buf.lock changes back to main
+# This helps avoid breaking changes when we cut new releases.
```
SpicyLemon, line 26: It's not helping avoid breaking changes. It's helping avoid a buf push failure due to a new third party proto reference.

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Buf setup action
+        uses: bufbuild/buf-setup-action@v1.13.1
+      - name: Buf push 'proto/'
+        uses: bufbuild/buf-push-action@v1
+        with:
+          input: 'proto'
+          buf_token: ${{ secrets.BUF_TOKEN }}
```
SpicyLemon, line 92: Since we should only push our protos during a release, this should really be in the `release.yaml`. That way A) we don't have to duplicate logic, and B) its in the chart when looking at the release Action in Github.

Flow:
1. The third-party protos get pushed whenever something is merged to `main`.
2. That creates a PR to update the lock.
3. That PR gets merged.
4. Things get back-ported to a release branch.
5. We tag the release.
6. Our protos get pushed, which reference the third party ones that were previously pushed and so are up-to-date.

We should take this job out of this workflow. Then, in the releases workflow, restore the `buf_push` job, but remove the step that pushes the third party protos.

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
+name: Buf Push
```
SpicyLemon, line 1: The name and filename don't need to be exact matches, but it's good to have them in sync.
```suggestion
name: Third Party Proto Registry
```
See other comments on why "Third Party" is in there.

Basically, you want it to be easy to start at Github, looking at the results of a workflow, and know which file to edit locally. Slightly less important is knowing the filename and wanting to find a run of the workflow, but it does happen.

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+# Protobuf runs buf (https://buf.build/) push updated proto files to https://buf.build/provenance-io
+# This workflow is only run when a .proto file has been changed or when this workflow has been modifed.
```
SpicyLemon, line 3:
```suggestion
# This workflow pushes the third party protos to buf.build whenever any of them are updated in `main`.
```

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "**/**.proto"
+      - ".github/workflows/proto-registry.yml"
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"     # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v1.0-rc1, v20.15.10-rc5
```
SpicyLemon, line 13:
```suggestion
on:
  push:
    branches:
      - main
    paths:
      - "third_party/**.proto"
      - ".github/workflows/proto-registry.yml"
```

See other comment for reasoning.

SpicyLemon (changes requested):
Some tweaks are needed.

.github/workflows/proto-registry.yml
egaxhaj: I'll put it back. 

I wasn't quite sure if it be better keeping the BSR push together even though they happened on different triggers (main, release).

.github/workflows/proto-registry.yml
egaxhaj: updated to reference `https://buf.build/provenance-io`

.github/workflows/proto-registry.yml
egaxhaj: updated.

.github/workflows/proto-registry.yml
egaxhaj: `proto_push` restored in `release.yml`

.github/workflows/proto-registry.yml
egaxhaj: fixed.

.github/workflows/proto-registry.yml
egaxhaj: done.

SpicyLemon (approved):

.github/workflows/proto-registry.yml
```diff
@@ -0,0 +1,74 @@
+name: Third Party Proto Registry
+# Protobuf runs buf (https://buf.build/) push updated proto files to https://buf.build/provenance-io
+# This workflow pushes the third party protos to buf.build whenever any of them are updated in `main`.
+on:
+  push:
```
SpicyLemon, line 5: Oh! Do we want to add `workflow_dispatch` in here too: [events-that-trigger-workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)? The we have the ability to manually kick off the action if we ever need to.

I've never tried it though.

egaxhaj: Neither have I. The docs state that it needs to be merged to `main` (default branch) before it can be manually triggered but am not sure if it's restricted to only run on `main` after that. See step 5 in [manually-running-a-workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). As long as we can only run it on main then we're fine.

I've enabled it but let me run a test on a test repo first before we merge it.

nullpointer0x00 (approved):

SpicyLemon (approved):

Issue #1256, opened 2022-12-08
Support for Tokens Restricted with Attributes


iramiller:
## Summary

Support for restricting tokens created by a marker which require accounts to possess certain attributes before allowing the tokens to move.

## Problem Definition

Currently the restricted marker system has a common use case of requiring attributes such as KYC records to exist on accounts before tokens can be sent.  This is implemented with extra steps using smart contracts to broker transfers.  While this is a workable process it prevents more complex workflows such as exchange functions from using these tokens without adding in auth grant steps.  These auth grant steps allow the checks to be bypassed which while workable is less than ideal.

## Proposal

With the bank transfer authorization hooks being added to the bank module it will be possible to support new authorization flows during any transfer of tokens between accounts.  With these hooks we should be able to look up a token marker registration and determine if there are restrictions in place.  If we added support for defining a set of attributes that must be present on the destination account then these hooks could allow the constraints to be easily enforced in a universal way.  This would mean that exchange functions and other smart contracts could start accepting these tokens while the marker admin can ensure that their attribute assertions are uniformly and consistently applied.

1. Update the marker module with a method for defining bank transfer restriction rules
2. Update the bank module send authorization hooks with a method that checks the marker module to ensure that any require checks are performed before approving a transfer of tokens between accounts.

Requires: https://github.com/cosmos/cosmos-sdk/pull/14224 and associated development of the hooks that are registered.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
The ideas here overlap a bit with this Ethereum proposal: https://eips.ethereum.org/EIPS/eip-3643

iramiller:
More detail:

- Restrictions can be added to the existing Restricted Coin type.
- A repeated list of name/value elements can be added to the Marker Account proto to specify the restrictions
- Values for the name should support wildcards (e.g. - `*.kyc.passport.pb`)
- The bank hooks for restrictions on transfer can be used so that normal bank transfers can occur and deprecate the Transfer method on the marker.
- Transfer access-grant can be checked as an override to the additional restrictions
- The Transfer access-grant can be used for controlling the transfer process ... so by extension this role should be required for requests that add/remove/modify restrictions on a marker

iramiller:
[...]

PR #1377, opened 2023-02-17
support for tokens restricted with attributes

nullpointer0x00:
## Description

Adds support for allowing sends when defined attributes are present in receiver of funds.

closes: #1256

---

write a little note why.


CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
+### Features
+* Add support for tokens restricted marker sends with required attributes [#1256](https://github.com/provenance-io/provenance/issues/1256))
```
SpicyLemon, line 41:
```suggestion
### Features

* Add support for tokens restricted marker sends with required attributes [#1256](https://github.com/provenance-io/provenance/issues/1256))
```

x/marker/client/cli/tx.go
```diff
@@ -235,14 +241,19 @@ with the given supply amount and denomination provided in the coin argument
...
+			}
+			msg := types.NewMsgAddMarkerRequest(coin.Denom, coin.Amount, callerAddr, callerAddr, typeValue, supplyFixed, allowGovernanceControl, requiredAttributes)
 
 			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
 		},
 	}
 	cmd.Flags().String(FlagType, "COIN", "a marker type to assign (default is COIN)")
 	cmd.Flags().Bool(FlagSupplyFixed, false, "a true or false value to denote if a supply is fixed (default is false)")
 	cmd.Flags().Bool(FlagAllowGovernanceControl, false, "a true or false value to denote if marker is allowed governance control (default is false)")
+	cmd.Flags().String(FlagRequiredAttributes, "", "comma delimited list of required attributes needed for a restricted marker to have send authority")
```
SpicyLemon, line 256: You might check out `Flags().StringSlice(...)` for the comma-sepparated stuff:

```
// StringSlice defines a string flag with specified name, default value, and usage string.
// The return value is the address of a []string variable that stores the value of the flag.
// Compared to StringArray flags, StringSlice flags take comma-separated value as arguments and split them accordingly.
// For example:
//   --ss="v1,v2" --ss="v3"
// will result in
//   []string{"v1", "v2", "v3"}
```

You'd then use `GetStringSlice` to get the vals out.

cmd/provenanced/cmd/genaccounts.go
```diff
@@ -503,6 +510,7 @@ enforced immediately.  An optional type flag can be provided or the default of C
 	cmd.Flags().String(flagEscrow, "", "A list of coins held by the marker account instance.  Note: Any supply not allocated to other accounts should be assigned here.")
 	cmd.Flags().BoolP(flagFinalize, "f", false, "Set the marker status to finalized.  Requires manager to be specified. (recommended)")
 	cmd.Flags().BoolP(flagActivate, "a", false, "Set the marker status to active.  Total supply constraint will be enforced as invariant.")
+	cmd.Flags().String(markercli.FlagRequiredAttributes, "", "comma delimited list of required attributes needed for a restricted marker to have send authority")
```
SpicyLemon, line 513: When you've got a common parser function for a flag, it's usually a good idea to also have an `Add<whatever>FlagsToCmd` function.

x/marker/keeper/proposal_handler_test.go
```diff
@@ -34,7 +34,7 @@ type IntegrationTestSuite struct {
 func (s *IntegrationTestSuite) SetupSuite() {
 	s.app = provenance.Setup(s.T())
 	s.ctx = s.app.BaseApp.NewContext(false, tmproto.Header{})
-	s.k = markerkeeper.NewKeeper(s.app.AppCodec(), s.app.GetKey(markertypes.ModuleName), s.app.GetSubspace(markertypes.ModuleName), s.app.AccountKeeper, s.app.BankKeeper, s.app.AuthzKeeper, s.app.FeeGrantKeeper, s.app.TransferKeeper, s.app.GetKey(banktypes.StoreKey))
+	s.k = markerkeeper.NewKeeper(s.app.AppCodec(), s.app.GetKey(markertypes.ModuleName), s.app.GetSubspace(markertypes.ModuleName), s.app.AccountKeeper, s.app.BankKeeper, s.app.AuthzKeeper, s.app.FeeGrantKeeper, s.app.TransferKeeper, s.app.AttributeKeeper, s.app.NameKeeper, s.app.GetKey(banktypes.StoreKey))
```
SpicyLemon, line 37: I'm not sure why we've got the bank key in our keeper. It's no longer used for anything though, so we should probably remove it.

x/marker/keeper/send_restrictions.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	accessAllowed, success := hasAccess.(bool)
+	if !success {
+		return false, fmt.Errorf("incorrect type for context %s value", AddressHasAccessKey)
+	}
+	return accessAllowed, nil
+}
+
+// WithAddressHasAccess returns the context with the address has access set used for allowing address to send restricted markers
+func WithAddressHasAccess(ctx sdk.Context, hasAccess bool) sdk.Context {
```
SpicyLemon, line 121: I kind of expected this to be named something more like `WithMarkerSendRestrictionBypass`. It's a little confusing to me to call it "address has access" when there's multiple addresses involved an this doesn't identify which one.

I'd also change `GetAddressHasAccess` to `HasMarkerSendRestrictionBypass` and have it only return a bool and return false if there's any error, basically `return success && accessAllowed`.

This can also be called first thing in `AllowMarkerSend` to skip over some lookups and processing.

x/marker/keeper/send_restrictions.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil
+	}
+
+	if len(marker.GetRequiredAttributes()) == 0 {
+		return fmt.Errorf("%s does not have transfer permissions", caller.String())
+	}
+	contains, err := k.ContainsRequiredAttributes(ctx, marker.GetRequiredAttributes(), to)
+	if err != nil {
+		return err
+	}
```
SpicyLemon, line 50: The `hasAccess` check should happen right off the bat (so that all the lookup and processing is skipped).
The `caller.Equals(k.markerModuleAddr)` check should happen next since it can be done and doesn't require any lookups.

Then, the `return fmt.Errorf("%s does not have transfer permissions", caller.String())` should be inside an if block checking for transfer permissions.

cmd/provenanced/cmd/genaccounts.go
SpicyLemon: I just realized I should have used the new `AddNewMarkerFlags` here (and `ParseNewMarkerFlags` above) (added in #1426).

nullpointer0x00: Ok, I will refactor to use this on when I merge PR #1426.

x/marker/client/cli/tx.go
nullpointer0x00: Thanks for pointing this out.

x/marker/types/msg.go
```diff
@@ -470,11 +478,16 @@ func (msg MsgGrantAllowanceRequest) GetSigners() []sdk.AccAddress {
...
+	totalSupply sdkmath.Int,
+	fromAddress sdk.AccAddress,
+	manager sdk.AccAddress,
 	markerType MarkerType,
-	supplyFixed, allowGovernanceControl, allowForcedTransfer bool,
+	supplyFixed bool,
+	allowGovernanceControl bool,
 	accessGrants []AccessGrant,
+	allowForcedTransfer bool,
+	requiredAttributes []string,
```
SpicyLemon, line 490: Please put the `accessGrants` argument last. That way, it's the same signature as `NewMsgAddMarkerRequest` with the addition of the `accessGrants` argument.

```suggestion
	supplyFixed bool,
	allowGovernanceControl bool,
	allowForcedTransfer bool,
	requiredAttributes []string,
	accessGrants []AccessGrant,
```

There's some calls to this func that'll need updating too.

x/marker/types/marker_test.go
```diff
@@ -108,95 +108,95 @@ func TestNewMarkerValidate(t *testing.T) {
 		},
 		{
 			"insufficient supply",
-			NewMarkerAccount(baseAcc, sdk.NewCoin("test", sdk.ZeroInt()), manager, nil, StatusFinalized, MarkerType_Coin, true, true, false),
+			NewMarkerAccount(baseAcc, sdk.NewCoin("test", sdk.ZeroInt()), manager, nil, StatusFinalized, MarkerType_Coin, true, false, false, []string{}),
```
SpicyLemon, line 111: Is there a reason to change the `allowGovernanceControl` value in these tests?

x/marker/types/marker.go
```diff
@@ -228,6 +233,20 @@ func ValidateGrantsForMarkerType(markerType MarkerType, grants ...AccessGrant) e
 	return ValidateGrants(grants...)
 }
 
+// ValidateRequiredAttributes checks that required attributes are of the correct format
+func ValidateRequiredAttributes(requiredAttributes []string) error {
```
SpicyLemon, line 237: Should `MarkerAccount.Validate()` call this?

x/marker/types/msg.go
```diff
@@ -52,14 +57,12 @@ func NewMsgAddMarkerRequest(
...
 		AllowForcedTransfer:    allowForcedTransfer,
+		RequiredAttributes:     requiredAttributes,
 	}
 }
 
 // ValidateBasic runs stateless validation checks on the message.
 func (msg MsgAddMarkerRequest) ValidateBasic() error {
-	if msg.Status == StatusUndefined {
-		return ErrInvalidMarkerStatus
-	}
```
SpicyLemon, line 62: Why is this being removed? It seems like a good validation check to have here.

x/marker/types/msg_test.go
```diff
@@ -165,6 +165,105 @@ func TestMsgIbcTransferRequestValidateBasic(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.msg.ValidateBasic()
+			if len(tc.errorMsg) > 0 {
+				assert.Error(t, err)
+				assert.Equal(t, tc.errorMsg, err.Error())
```
SpicyLemon, line 259: These can be combined using `EqualError`:
```suggestion
				assert.EqualError(t, err, tc.errorMsg)
```

x/marker/keeper/send_restrictions_test.go
```diff
@@ -0,0 +1,409 @@
...
+			reqAttr:        "",
+			attr:           "test.provenance.com",
+			expectedResult: false,
+		},
+		{
+			name:           "should fail - empty required attr and attr ",
+			reqAttr:        "",
+			attr:           "",
+			expectedResult: false,
+		},
```
SpicyLemon, line 403: Some other test cases that would be good:

```golang
		{
			name:           "should fail - extra ending",
			reqAttr:        "test.provenance.io",
			attr:           "test.provenance.iox",
			expectedResult: false,
		},
		{
			name:           "should fail - wildcard extra ending",
			reqAttr:        "*.provenance.io",
			attr:           "test.provenance.iox",
			expectedResult: false,
		},
		{
			name:           "should fail - wildcard extra beginning",
			reqAttr:        "*.provenance.io",
			attr:           "test.xprovenance.io",
			expectedResult: false,
		},
```

internal/handlers/bank_send_restriction_test.go
```diff
@@ -0,0 +1,117 @@
...
+	assert.Equal(tt, "100nonrestrictedmarker,100restrictedmarker,1000000000stake", addr2afterBalance, "addr2beforeBalance")
+
+	sendRMarker = banktypes.NewMsgSend(addr2, addr1, sdk.NewCoins(sdk.NewInt64Coin(nrMarkerDenom, 50)))
+	ConstructAndSendTx(tt, *app, ctx, acct2, priv2, sendRMarker, abci.CodeTypeOK, "")
+	addr1afterBalance = app.BankKeeper.GetAllBalances(ctx, addr1).String()
+	assert.Equal(tt, "950nonrestrictedmarker,900restrictedmarker,999400000stake", addr1afterBalance, "addr1afterBalance")
+	addr2afterBalance = app.BankKeeper.GetAllBalances(ctx, addr2).String()
+	assert.Equal(tt, "50nonrestrictedmarker,100restrictedmarker,999850000stake", addr2afterBalance, "addr2beforeBalance")
+
+	stopIfFailed(tt)
```
SpicyLemon, line 103: First, you don't need this `stopIfFailed(tt)` call at the end since it doesn't do anything that the unit test won't already do at the end of a test.

Second, It'd be good to have a couple other test cases that demonstrate some failed `MsgSend`s.
1. On a restricted coin without required attributes where the sender doesn't have `TRANSFER` permission.
2. On a restricted coin with required attributes from a sender that has `TRANSFER` permission, but the receiver doesn't have the required attributes.
3. On a restricted coin with required attributes from a sender that does not have `TRANSFER` permission, but the receiver DOES have the required attributes.

Also some failed `MsgTransfer`s:
1. On a restricted coin with required attributes using an admin that has `TRANSFER` permission, but the receiver doesn't have the required attributes.
2. On a restricted coin with required attributes using an admin that does not have `TRANSFER` permission, but the receiver DOES have the required attributes.

internal/handlers/bank_send_restriction_test.go
```diff
@@ -0,0 +1,117 @@
...
+	encCfg := sdksim.MakeTestEncodingConfig()
+	fees := sdk.NewCoins(sdk.NewInt64Coin(sdk.DefaultBondDenom, int64(NewTestGasLimit())))
+	acct = app.AccountKeeper.GetAccount(ctx, acct.GetAddress()).(*authtypes.BaseAccount)
+	txBytes, err := SignTxAndGetBytes(NewTestGasLimit(), fees, encCfg, priv.PubKey(), priv, *acct, ctx.ChainID(), msg)
+	require.NoError(tt, err, "SignTxAndGetBytes")
+	res := app.DeliverTx(abci.RequestDeliverTx{Tx: txBytes})
+	require.Equal(tt, expectedCode, res.Code, "res=%+v", res)
+	if len(expectedError) > 0 {
+		require.True(tt, strings.Contains(res.Log, expectedError), fmt.Sprintf("error msg does not contain %s", expectedError))
+	}
```
SpicyLemon, line 116: This is not what `require.True` is to be used for.

```suggestion
	if len(expectedError) > 0 {
		require.Contains(res.Log, expectedError, "DeliverTx result.Log")
	}	
```

That way, when it fails, the resulting test failure output includes the `res.Log` content which greatly helps identify why the test failed.

The only time to use `.True(...)` or `.False(...)` is when the thing you're testing outputs a boolean. Otherwise, there's always better ways to make your assertions.

Furthermore, you don't need the `fmt.Sprintf` for the test failure message. In fact, it's better without it because a) that delays the string generation until it's needed, and b) the line is shorter.

Basically, this: `require.Equal(t, a, b, "result from thing %v", someThing)` results in the same test output as `require.Equal(t, a, b, fmt.Sprintf("result from thing %v", someThing))`. The former only does a `Sprintf` if the test fails and that message is needed, though.

x/marker/types/marker_test.go
nullpointer0x00: Nope, did it by accident while fixing merge.  Changing it back.

x/marker/types/msg.go
nullpointer0x00: Just seemed redundant since we have this a few lines down: 
```
	if msg.Status != StatusFinalized && msg.Status != StatusProposed {
		return fmt.Errorf("marker can only be created with a Proposed or Finalized status")
	}
```

x/marker/types/marker.go
nullpointer0x00: I was actually meaning to remove this.  The attribute validation is a little more complex than I first thought.    The validation will happen when the attributes are being normalized in the msg_server.  
```
	normalizedReqAttrs, err := k.NormalizeRequiredAttributes(ctx, msg.RequiredAttributes)
	if err != nil {
		return nil, err
	}
```

x/marker/keeper/send_restrictions_test.go
nullpointer0x00: Thanks! Your `"should fail - wildcard extra beginning",` found an issue.  Fixed now.

x/marker/types/msg_test.go
```diff
@@ -165,6 +165,105 @@ func TestMsgIbcTransferRequestValidateBasic(t *testing.T) {
...
+func TestMsgAddMarkerRequestValidateBasic(t *testing.T) {
+	validAddress := sdk.MustAccAddressFromBech32("cosmos1sh49f6ze3vn7cdl2amh2gnc70z5mten3y08xck")
+
+	cases := []struct {
+		name     string
+		msg      MsgAddMarkerRequest
+		errorMsg string
+	}{
+		{
+			"should fail on attributes for non restricted coin",
```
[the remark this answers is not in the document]
nullpointer0x00: Fixed older tests in that file too.

x/marker/keeper/send_restrictions.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+const AddressHasAccessKey = "address_has_access"
+
+func (k Keeper) AllowMarkerSend(ctx sdk.Context, from, to, denom string) error {
+	if HasMarkerSendRestrictionBypass(ctx) {
+		return nil
+	}
+
+	caller, err := sdk.AccAddressFromBech32(from)
```
iramiller, line 20: is this the most efficient way to handle this or are we doing the address/bech32 conversion multiple times...

iramiller (approved):

x/marker/keeper/send_restrictions.go
nullpointer0x00: From this point on there are no more conversions happening.  I was actually thinking about this while writing this area of code.  

For example, the SendCoin call from the sdk uses AccAddr while the multi send uses the bech32 string.  So, I opted to convert the SendCoin `to` and `from` to strings in my function signature.  I needed to convert the from address back to AccAddress for checking module account and transfer permissions.  Also, for the `from` address uses the bech32 string to look up the attributes. 
I was originally going to use the function signature `AllowMarkerSend(ctx sdk.Context, from sdkAccAddress, to sdkAccAddress, denom string)` instead of `AllowMarkerSend(ctx sdk.Context, from, to, denom string)`.  However, after seeing how the multi-send worked I kept it as strings.

Looking in the future with this PR https://github.com/cosmos/cosmos-sdk/pull/14224 the signature will have to change to something `(ctx sdk.Context, from sdk.AccAddress, to sdk.AccAddress, coins sdk.Coins)`.  In their PR I'm not seeing how they are handling the multi-send and their respective string addresses.  Also, the sdk.Coins part will change the impl a bit too.

If you think it is worth the effort at the moment, I could change the signature to the future sdk change.  This will need some re-working of the sdk and the sdk.Coins will require changes to the `AllowMarkerSend` to loop through the coins and ensure all of them pass.

iramiller: I am happy to hear that the area is being noted. 👍🏻

Issue #677, opened 2022-02-24
Add gRPC query to get all contract specs and record specs for a scope spec


minxylynx:
## Summary

There a several gRPC queries used by Explorer to fetch spec data. However, there is not a one and done query to get all contract and record specs under a scope spec, as there is for sessions and records under a scope. This would add a query that gives all requested information under a scope spec.

## Proposal

* TBD @iramiller 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
I just got some counts of specs on mainnet.

Scope Specs: 22
Contract Specs: 7609
Record Specs: 24983

Ave record specs per contract spec = 24983/7609 = 3.28334866605336
Ave contract specs per scope spec = 7609/22 = 345.863636363636
Ave record specs per contract spec = 24983/22 = 1135.59090909091

So the average result would have 1 scope spec + 3 contract specs + 1136 record specs = 1140 things in the result.

That's a pretty big result.

Furthermore, since there's only 22 scopes, I grabbed them all to get contract spec counts for each. Here's the distribution:
| scope specs | number of contract specs |
|-------------|--------------------------|
| 15          | 1                        |
|  1          | 112                      |
|  1          | 129                      |
|  1          | 2401                     |
|  1          | 30                       |
|  1          | 5                        |
|  1          | 747                      |
|  1          | 9                        |

That scope spec with 2401 contract specs probably has 2401 * 3.28334866605336 = 7883.32014719412 records. That's 1 scope + 2401 contract specs + 7883 record specs = 10285 things in the result. That's more than likely a problem.

Pagination doesn't make sense for this query since we're returning three different types of things. I guess only two might need pagination, but that's still complicated.

Testnet has more of each spec type, but the averages and that max are pretty similar.

I guess this is something to consider, and keep an eye on, but it's possible the query just doesn't work for some scope specs (or takes multiple tries).

SpicyLemon:
Also, rather than adding a new query for this, it's probably best to just enhance the `ScopeSpecification` query to behave similar to the `Scope` query.

PR #1399, opened 2023-03-02
support for grpc query of contract and record specs for scope spec

egaxhaj:
## Description

Adds support so a user can `opt-in` and query all contract and record specs for a scope spec.

When a user performs a gRPC query to retrieve a scope spec they can now opt-in to also retrieve contract and record specs belonging to the scope spec by setting `IncludeContractSpecs = true` and/or `IncludeRecordSpecs = true` in `ScopeSpecificationRequest` msg.

closes: #677

---

write a little note why.


nullpointer0x00 (commented):
Don't we need to update the cli command?

egaxhaj:
> Don't we need to update the cli command for the optional includes?

 I could not find any references to other queries using similar flags in the CLI so I left it out. For example: `Scope()` has  `IncludeSessions, IncludeRecords` flags without any references in CLI, only gRPC.

https://github.com/provenance-io/provenance/blob/0adff7fa0eac6277f30ebd1fa81689c71528a5a2/x/metadata/keeper/query_server.go#L104-L122

@SpicyLemon do we need CLI support for the `IncludeContractSpecs, IncludeRecordSpecs` flags.

x/metadata/keeper/query_server.go
```diff
@@ -649,6 +650,38 @@ func (k Keeper) ScopeSpecification(c context.Context, req *types.ScopeSpecificat
 		retval.ScopeSpecification = types.WrapScopeSpecNotFound(specAddr)
 	}
 
+	if found && req.IncludeContractSpecs {
+		err := k.IterateContractSpecs(ctx, func(cs types.ContractSpecification) (stop bool) {
```
SpicyLemon, line 654: This is way too much work. Just loop over `spec.ContractSpecIds` and get each one specifically: `k.GetContractSpecification(ctx, contractSpectID)`.

x/metadata/keeper/query_server.go
```diff
@@ -649,6 +650,38 @@ func (k Keeper) ScopeSpecification(c context.Context, req *types.ScopeSpecificat
...
+		if err != nil {
+			return &retval, fmt.Errorf("error iterating scope spec [%s] contract specs: %w", spec.SpecificationId, err)
+		}
+	}
+
+	if found && req.IncludeRecordSpecs {
+		var err error
+		for _, id := range spec.ContractSpecIds {
+			err = k.IterateRecordSpecsForContractSpec(ctx, id, func(recordSpecID types.MetadataAddress) (stop bool) {
+				err = k.IterateRecordSpecs(ctx, func(recordSpec types.RecordSpecification) (stop bool) {
```
SpicyLemon, line 671: Whoa. Way too much work here. You've got the record id, just get it directly: `k.GetRecordSpecification(ctx, recordSpecID)`.

SpicyLemon (changes requested):
A couple performance-related changes are needed.

SpicyLemon:
> @SpicyLemon do we need CLI support for the `IncludeContractSpecs, IncludeRecordSpecs` flags.

Yes. There needs to be CLI support for the options you're adding to the queries.

The scope, session and record commands all have related flags, e.g. The get scope command has a `--include-sessions` flag.

See:
https://github.com/provenance-io/provenance/blob/main/x/metadata/client/cli/query.go#L1136-L1140

and
https://github.com/provenance-io/provenance/blob/main/x/metadata/client/cli/query.go#L604-L617

A flag already exists for record specs `--include-record-specs` to variable `includeRecordSpecs`. That can be re-used here, but one will need to be added for `--include-contract-specs`.

x/metadata/keeper/query_server.go
egaxhaj: I avoided `k.GetContractSpecification(ctx, contractSpectID)` because it uses `MustUnmarshal`, which panics. It's an edge case but I wanted to avoid a panic on a standard query.

SpicyLemon: You can't iterate over all contract specs every time you need a couple. The query just won't work in the long run.

For example on either mainnet or testnet, if you try to get a count of contract specs using `provenanced q metadata all contractspecs --limit 1 --count-total`, the first attempt will time out because there's just too many too loop through all at once. A second attempt can succeed because it's loaded the first chunk of them into a cache. But it might even take three or more. Try getting a count of scopes; I can't remember how many times I had to try before it returned something for me.

We have all the information we need, there's no need to look at everything to find it.

egaxhaj: changed to use `k.GetContractSpecification` for better performance.

x/metadata/keeper/query_server.go
egaxhaj: changed to use `k.GetRecordSpecification` for better performance.

x/metadata/keeper/query_server.go
```diff
@@ -649,6 +649,35 @@ func (k Keeper) ScopeSpecification(c context.Context, req *types.ScopeSpecificat
 		retval.ScopeSpecification = types.WrapScopeSpecNotFound(specAddr)
 	}
 
+	if found && req.IncludeContractSpecs {
+		for _, id := range spec.ContractSpecIds {
+			cs, ok := k.GetContractSpecification(ctx, id)
+			if !ok {
+				return &retval, fmt.Errorf("error iterating scope spec [%s] contract specs: %w", spec.SpecificationId, err)
```
SpicyLemon, line 656:
```suggestion
				return &retval, fmt.Errorf("error getting contract spec [%s]: %w", spec.SpecificationId, err)
```

egaxhaj: ah, this actually needs to be `id` instead of `spec.SecificcationId`

x/metadata/keeper/query_server.go
```diff
@@ -649,6 +649,35 @@ func (k Keeper) ScopeSpecification(c context.Context, req *types.ScopeSpecificat
...
+		}
+	}
+
+	if found && req.IncludeRecordSpecs {
+		var err error
+		for _, id := range spec.ContractSpecIds {
+			err = k.IterateRecordSpecsForContractSpec(ctx, id, func(recordSpecID types.MetadataAddress) (stop bool) {
+				rs, ok := k.GetRecordSpecification(ctx, recordSpecID)
+				if !ok {
+					err = fmt.Errorf("error retrieving record spec [%s] for contract spec [%s]", recordSpecID, id)
```
SpicyLemon, line 668: I know this is being changed because of the other comment, but I wanted to point something out here.

This `err` value will never be seen. When `IterateRecordSpecsForContractSpec` ends, the `err` value will be set to something returned directly by that function. In this case, it will be `nil` because the iterator didn't encounter any errors.

x/metadata/keeper/query_server.go
```diff
@@ -649,6 +649,35 @@ func (k Keeper) ScopeSpecification(c context.Context, req *types.ScopeSpecificat
 		retval.ScopeSpecification = types.WrapScopeSpecNotFound(specAddr)
 	}
 
+	if found && req.IncludeContractSpecs {
+		for _, id := range spec.ContractSpecIds {
+			cs, ok := k.GetContractSpecification(ctx, id)
+			if !ok {
+				return &retval, fmt.Errorf("error getting scope spec [%s] contract spec [%s]: %w", spec.SpecificationId, id, err)
+			}
```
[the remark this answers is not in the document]
egaxhaj: fixed. Both contract and record specs use NotFound wrappers instead of returning an error.

SpicyLemon (approved):

SpicyLemon (approved):

Issue #1402, opened 2023-03-06
Remove the P8e endpoints


SpicyLemon:
## Summary

Remove the P8e endpoints and P8e-specific messages and code.

## Problem Definition

The P8e endpoints haven't been used since June 2022. They only existed to help facilitate the transition to the split-out metadata endpoints. They can now be removed.

## Proposal

Remove the P8e-specific endpoints, messages, and code.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Changed issue theme from "deprecate" to "remove" because the Figure folks said we could :killitwithfire:.

PR #1406, opened 2023-03-08
Delete the P8E endpoints

SpicyLemon:
## Description

closes: #1402 

Delete the P8E endpoints (`WriteP8eContractSpec` and `P8eMemorializeContract`) and all the stuff just for them.

This PR also has a couple tertiary updates that resulted from running `make proto-format` or `make proto-gen`. It updates the `proto-all` make target to use `update-swagger-docs` instead of `proto-swagger-gen`; the former uses the latter, so it's still doing everything it used to, but now it also updates the `statik.go` file (which is always needed).

---

This PR is a proto-breaking change. Any clients using old protos that attempt to call one of those endpoints, will get an error. The `github.com/provenance-io/provenance/x/metadata/types/p8e` package was also removed. The only reason to use anything in there was when calling one of those endpoints though.

In order to not have a breaking change, I could have either A) just deprecated all the stuff, or B) created a `v2` without those endpoints. I chose not to do A because the stuff was already thematically deprecated, hasn't been used since June 2022, and even having it in an officially deprecated/unusable state can cause confusion. I chose not to do B because switching to a v2 would cause headaches for everyone currently using the `x/metadata` stuff, whereas leaving it at v1 only causes headaches for anyone trying to use those endpoints.

---

write a little note why.


iramiller (approved):

SpicyLemon:
Note that the "Protobuf Checks / Protobuf Checks (pull_request) action is expected to fail here.

```
Run bufbuild/buf-breaking-action@v1.1.2
  with:
    input: proto
    against: https://github.com/provenance-io/provenance.git#branch=main,subdir=proto
    buf_input_https_username: SpicyLemon
    buf_input_https_password: ***
Error: Previously present message "MsgP8eMemorializeContractRequest" was deleted from file.
Error: Previously present message "MsgP8eMemorializeContractResponse" was deleted from file.
Error: Previously present message "MsgWriteP8eContractSpecRequest" was deleted from file.
Error: Previously present message "MsgWriteP8eContractSpecResponse" was deleted from file.
Error: Previously present RPC "P8eMemorializeContract" on service "Msg" was deleted.
Error: Previously present RPC "WriteP8eContractSpec" on service "Msg" was deleted.
Error: buf found [6](https://github.com/provenance-io/provenance/actions/runs/4368422087/jobs/7640983374#step:7:6) breaking changes.
```

From what I can tell, the action that does it's check against github (as opposed to `buf.build`) referencing the PRs target branch (`main` in this case), so it shouldn't cause problems for other branches.

That being said, there's a good chance that the `buf push` action will fail when we include this change in a release. I'm not yet sure how to address that though, but we've got time.

egaxhaj (approved):

egaxhaj (approved):

iramiller (approved):

Issue #1368, opened 2023-02-15
Clawback on Restricted Tokens


iramiller:

iramiller:
See also: #1256

SpicyLemon:
After some discussions, we decided to go a slightly different route with this.

We discussed "forced transfers" vs "clawback" and which one we want here. "Forced transfers" allow the coins to be transferred to any destination, whereas "clawback" would only allow the funds to go back the marker.

We decided to add the ability to do forced transfers on restricted coins that are set up to allow forced transfers.

The title of this issue will be updated to reflect that.

PR #1421, opened 2023-03-15
Allow marker clawback

SpicyLemon:
## Description

closes: #1368

Add the ability to allow clawback of restricted marker coins.

---

write a little note why.


app/app_test.go
```diff
@@ -113,6 +113,7 @@ func TestExportAppStateAndValidators(t *testing.T) {
 			MarkerType:             markertypes.MarkerType_RestrictedCoin,
 			SupplyFixed:            true,
 			AllowGovernanceControl: true,
+			ClawbackEnabled:        false,
```
iramiller, line 116: We should consider using the term "ForcedTransfer" as a clawback is a specific type of forced transfer but this implementation supports more than just pulling the tokens back into the marker account.

x/marker/types/marker_test.go
```diff
@@ -107,75 +108,95 @@ func TestNewMarkerValidate(t *testing.T) {
...
 			nil,
 		},
 		{
 			"valid marker account",
-			NewMarkerAccount(baseAcc, sdk.NewCoin("test", sdk.OneInt()), manager, nil, StatusActive, MarkerType_Coin, true),
+			NewMarkerAccount(baseAcc, sdk.NewCoin("test", sdk.OneInt()), manager, nil, StatusActive, MarkerType_Coin, true, false),
+			nil,
+		},
+		{
+			"coin type with clawback is ok",
```
iramiller, line 183: Because you can not add a transfer grant to an unrestricted marker it probably doesn't make sense to allow forced transfer on regular coins.

x/marker/keeper/marker.go
```diff
@@ -673,7 +673,7 @@ func (k Keeper) TransferCoin(ctx sdk.Context, from, to, admin sdk.AccAddress, am
 	if !m.AddressHasAccess(admin, types.Access_Transfer) {
```
iramiller, line 673: This specific check is why the forced transfer/clawback only applies to restricted coins ... because regular coins can not have this grant created.

iramiller (commented):
Some high level comments about naming and implementation.

iramiller:
Along the same lines of clawback ... if we keep this naming then we should narrow the transfer capability such that the token can literally only be "clawed back" into the Marker.  This would allow a nice interaction with the Withdraw grant because it would be a separate permission to subsequently redeploy the tokens that were clawed back.

SpicyLemon:
After a bit of discussion, we're going to take this in a slightly different direction. So I'm going to close this PR and will make a new one later.

SpicyLemon:
Replaced by #1426

Issue #1427, opened 2023-03-16
Add upgrade handler migration to enable send for restricted markers


nullpointer0x00:
## Summary

Currently, all restricted markers have been set to SendEnabled false.  With the implementation of https://github.com/provenance-io/provenance/issues/1256 we can delete all items from the table.  This will allow the send restrictions to be applied with the new handler.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
For this migration ... are the "send disabled" records within the bank module simply deleted? Or is there still a record for each token created ... this will have some impact on the management of that configuration within the marker module....

nullpointer0x00:
As of right now, we are setting the send enabled to true(coin) or false(restricted) within from the marker module.  With #1256 I set true for both with our current `ensureSendEnabledStatus` method.  However, we could probably just delete them all and not have our marker module set them.  The cosmos sdk has a parameter that indicates what it should default to if the coin is not found.  On both testnet and mainnet it is set to true.  Therefore, if they all deleted from the table on upgrade IsSendEnabled will be true for the coin.  

It would be nice to delete the `ensureSendEnabledStatus` method from the marker module and have the upgrade remove all entries. @SpicyLemon Thoughts?

nullpointer0x00:
I agree.  It is now removed and I updated the description of the ticket.

SpicyLemon:
Just catching up on this.

I agree with removing `ensureSendEnabledStatus`.

And the only reason to keep all of those `SendEnabled` records around would be in the crazy off-chance that we change the `DefaultSendEnabled` param. I feel like that's pretty far outside the realm of possibilities, though, so it's okay to remove all those records. It'll save a little state, but each is just a key/value pair where the value is a single byte.

But since we need to either update them all to `true` or delete them, might as well just delete them.

PR #1429, opened 2023-03-16
add new upgrade handler to remove is send enabled items

nullpointer0x00:
## Description

Adds upgrade handler to remove all IsSendEnabled entries.

closes: #1427 

---

write a little note why.


app/upgrades.go
```diff
@@ -181,73 +95,10 @@ func GetUpgradeStoreLoader(app *App, info upgradetypes.Plan) baseapp.StoreLoader
...
-	params := &sanction.Params{
-		ImmediateSanctionMinDeposit:   sdk.NewCoins(sdk.NewInt64Coin("nhash", 1_000_000_000_000_000)),
-		ImmediateUnsanctionMinDeposit: sdk.NewCoins(sdk.NewInt64Coin("nhash", 1_000_000_000_000_000)),
-	}
-	err := app.SanctionKeeper.SetParams(ctx, params)
-	if err != nil {
-		return fmt.Errorf("could not set sanction params: %w", err)
+// RemoveIsSendEnabledEntries removes all entries in the bankkeepers send enabled table
+func RemoveIsSendEnabledEntries(ctx sdk.Context, app *App) {
+	sendEnabledItems := app.BankKeeper.GetAllSendEnabledEntries(ctx)
```
iramiller, line 100: This is probably fine ... but should we verify that there is a marker associated with this denom before removing the entry just to be extra careful?

app/upgrades.go
```diff
@@ -38,74 +30,15 @@ type appUpgrade struct {
...
-		},
-	},
-	"paua-rc2": { // upgrade for v1.14.0-rc3
-		Handler: func(ctx sdk.Context, app *App, plan upgradetypes.Plan) (module.VersionMap, error) {
-			// Reapply the max commissions thing again so testnet gets the max change rate bump too.
-			IncreaseMaxCommissions(ctx, app)
-			UndoMaxGasIncrease(ctx, app)
-			return app.UpgradeKeeper.GetModuleVersionMap(ctx), nil
-		},
-	},
```
SpicyLemon, line 108: This was the last one that ran on testnet, so it's still needed.

app/upgrades.go
```diff
@@ -38,74 +30,15 @@ type appUpgrade struct {
...
-			}
-
-			ctx.Logger().Info("Starting migrations. This may take a significant amount of time to complete. Do not restart node.")
-			return app.mm.RunMigrations(ctx, app.configurator, versionMap)
-		},
-	},
-	"ochre": { // upgrade for v1.13.0
-		Added: []string{group.ModuleName, rewardtypes.ModuleName, icacontrollertypes.StoreKey, icahosttypes.StoreKey},
+	"quicksilver-rc1": { // upgrade for v1.15.0-rc1
+		Added: []string{},
```
SpicyLemon, line 34: This can be omitted.

app/upgrades_test.go
```diff
@@ -16,16 +20,47 @@ type IntegrationTestSuite struct {
 	ctx sdk.Context
 }
 
+func TestIntegrationTestSuite(t *testing.T) {
+	suite.Run(t, new(IntegrationTestSuite))
+}
+
 func (s *IntegrationTestSuite) SetupSuite() {
 	s.app = Setup(s.T())
```
SpicyLemon, line 28: I haven't checked the result of the github actions, but last I knew, you couldn't call `Setup` in this directory. I didn't nail down the cause but I think it had to do with some `init()` functions somewhere.

If the tests are passing, though, then I'm wrong, and it's al good (and you can resolve this comment). But if the these tests are failing, that's probably what's up with them. I'm not yet sure how best to address that problem.

nullpointer0x00: I see what you mean.  Adding:
 ```func TestIntegrationTestSuite(t *testing.T) {
	suite.Run(t, new(IntegrationTestSuite))
}```
Actually, started executing the tests in the upgrades_test.go file.  The `TestUpgradeICA` we had for a previous upgrade was not being executed. Running the test after my changes works, but you need to run it by itself.  Running all the tests in the package causes a failure.  Something to do with the InitGenesis and creating the fee_module account when setup is called.

app/upgrades.go
```diff
@@ -38,50 +30,15 @@ type appUpgrade struct {
...
-		Added: []string{group.ModuleName, rewardtypes.ModuleName, icacontrollertypes.StoreKey, icahosttypes.StoreKey},
-		Handler: func(ctx sdk.Context, app *App, plan upgradetypes.Plan) (module.VersionMap, error) {
-			versionMap := app.UpgradeKeeper.GetModuleVersionMap(ctx)
-			UpgradeICA(ctx, app, &versionMap)
-			ctx.Logger().Info("Starting migrations. This may take a significant amount of time to complete. Do not restart node.")
-			return app.mm.RunMigrations(ctx, app.configurator, versionMap)
-		},
-	},
-	"ochre-rc2": { // upgrade for 1.13.0-rc5
+	"quicksilver-rc1": { // upgrade for v1.15.0-rc1
```
SpicyLemon, line 33: Super nitty on this...

1. There's a comment at the end of the `handlers` hash indicating that's where new handlers should go, and this one is being added to the top of the hash instead.
2. It'd be best to keep the `paua` and `paua-rc2` handlers in the same order as they currently are.

This will make this diff easier to process. `git` will correctly identify that the `ochre-rc1`, `ochre-rc2`, `ochre`, and `paua-rc1` upgrades are being deleted, nothing is changing with `paua-rc2` or `paua`, and `quicksilver-rc1` is being added with all new code.

app/upgrades_test.go
SpicyLemon: I mentioned this in slack, but thought I'd put it here too in case we're wondering about it in the future.

The reason the tests were failing in here when we tried to use that `Setup(s.T())` function has to do with concurrent interactions from the `app/prefix_test.go` tests. Those tests call the `SetConfig` function which (among other things) sets the bech32 HRP to either `tp` or `pb` as global variables.

What was happening is that, when running all the tests in `./app/...`:
1. Part of the `Setup` process was happening with the bech32 HRP set to `cosmos` (the default). It gets through the part where it generates a `genesis.json` file where all the addresses in it start with `cosmos1`.
2. The tests in `app/prefix_test.go` run and complete and change the HRP to `tp` in the process.
3. The `Setup` process gets to the point where it calls `InitGenesis`. Now, though, it can't decode those bech32 addresses because it's expecting them to have the HRP `tp` but they have `cosmos`. That error from `AccAddressFromBech32` is getting ignored, though, in the `BaseAccount.GetAddress()`, which instead just returns an empty address. In turn, that caused the actual error in question because that address is being provided to a `store.Set(key, value)` call as the `value`, which cannot be empty.

The fix is to put the `app/prefix_test.go` tests in a different package (e.g. `app_test`), or move them to a different directory. That way, nothing else is affected by the calls to `SetConfig`.

x/marker/keeper/keeper_test.go
```diff
@@ -943,3 +946,40 @@ func TestGetAuthority(t *testing.T) {
...
+		Supply:     sdk.NewInt(1000),
+		MarkerType: markertypes.MarkerType_Coin,
+		Status:     markertypes.StatusActive,
+	})
+	require.NoError(t, err, "should have added marker")
+	app.BankKeeper.SetSendEnabled(ctx, "testcoin", false)
+	app.BankKeeper.SetSendEnabled(ctx, "nonmarkercoin", false)
+
+	sendEnabledItems := app.BankKeeper.GetAllSendEnabledEntries(ctx)
+	assert.Equal(t, len(sendEnabledItems), 2, "should have 2 items before removal")
```
SpicyLemon, line 976: When doing slice length assertions, the `.Len(...)` assertion is the way to go.
```suggestion
	assert.Len(t, sendEnabledItems, 2, "before removal")
```

The nice thing there is that when the assertion fails, the contents of `sendEnabledItems` will be in the failure message.

Even better would be asserting things exactly using `.Equals`. If you don't want to worry about ordering, you can use `ElementsMatch`. If the `actual` might have extra things, and you just want to make sure your things exist, you can use `.Contains` once for each expected entry.

But at the very least, please use `.Len` instead of `.Equals(t, len(...), ...)`.

x/marker/keeper/keeper_test.go
```diff
@@ -943,3 +946,40 @@ func TestGetAuthority(t *testing.T) {
...
+
+	app.MarkerKeeper.RemoveIsSendEnabledEntries(ctx)
+	sendEnabledItems = app.BankKeeper.GetAllSendEnabledEntries(ctx)
+	assert.Equal(t, len(sendEnabledItems), 1, "denom without a marker should only exist")
+	assert.Equal(t, sendEnabledItems[0].Denom, "nonmarkercoin", "denom without a marker should only exist")
+	assert.False(t, app.BankKeeper.IsSendEnabledDenom(ctx, "nonmarkercoin"), "should be in table as false since there is not a marker associated with it")
+	assert.True(t, app.BankKeeper.IsSendEnabledDenom(ctx, "testcoin"), "should not exist in table therefore default to true")
```
SpicyLemon, line 983: Good use of `.False` and `.True` here.

The thing being tested is the result of `IsSendEnabledDenom` which returns a boolean. So it makes sense to use those tests here.

Just wanted to point that out. Feel free to resolve this comment whenever.

x/marker/keeper/keeper_test.go
nullpointer0x00: Thanks, didn't know this existed or think to look.

iramiller (approved):

app/upgrades.go
```diff
@@ -38,50 +30,7 @@ type appUpgrade struct {
...
-			params := app.MsgFeesKeeper.GetParams(ctx)
-			app.MsgFeesKeeper.SetParams(ctx, params)
-
-			UpgradeICA(ctx, app, &versionMap)
-			ctx.Logger().Info("Starting migrations. This may take a significant amount of time to complete. Do not restart node.")
-			return app.mm.RunMigrations(ctx, app.configurator, versionMap)
-		},
-	},
-	"paua-rc1": { // upgrade for v1.14.0-rc2
+	"paua": { // upgrade for v1.14.0
```
SpicyLemon, line 33: The `"paua"` and `"paua-rc2"` entries are still backwards from what they originally were. That's why it looks like your new `"quicksilver-rc1"` entry is modifying existing stuff.

SpicyLemon (approved):

app/upgrades.go
SpicyLemon: Approving anyway.

Issue #1447, opened 2023-03-28
Support the lookup of address by attribute name


nullpointer0x00:
## Summary

Add the ability to look up addresses that have a specific attribute name.  This will include a new table to query the attribute and receive the address of the accounts associated to it and a new query proto with cli command.

NOTE: An upgrade handler will be required to populate the new table in testnet and mainnet releases.

## Problem Definition

Currently, we can only look up attributes by address and unable to do the reverse of looking up address by attribute name.

## Proposal

1. Create new table that has a key of [prefix][attribute_name][address]
2. Add entry to table in current flow for adding/updating/removing attributes
3. Add proto to query for address that have attribute
4. Add cli command to query for addresses
5. Add upgrade handler to populate table on upgrade

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1448, opened 2023-03-28
Support lookup of address by attribute name

nullpointer0x00:
## Description

closes: #1447

---

write a little note why.


egaxhaj (approved):
LGTM

x/attribute/keeper/keeper.go
```diff
@@ -217,13 +249,15 @@ func (k Keeper) UpdateAttribute(ctx sdk.Context, originalAttribute types.Attribu
 		if attr.Name == updateAttribute.Name && bytes.Equal(attr.Value, originalAttribute.Value) && attr.AttributeType == originalAttribute.AttributeType {
 			found = true
 			store.Delete(it.Key())
+			k.DecAddNameAddressLookup(ctx, attr)
```
iramiller, line 252: This is an odd flow ... an update with a decrement then an increment... While it might seem like there could be a break due to error handling in the marshal of the update ... this same error would prevent the transaction from committing at all so the delete would not happen.  Given this I don't see why we need to do a `-1/+1` in the same flow.

x/attribute/keeper/keeper_test.go
```diff
@@ -364,57 +410,66 @@ func (s *KeeperTestSuite) TestDeleteAttribute() {
...
-	s.NoError(s.app.AttributeKeeper.SetAttribute(s.ctx, attr, s.user1Addr), "should save successfully")
+	s.Assert().NoError(s.app.AttributeKeeper.SetAttribute(s.ctx, attr, s.user1Addr), "should save successfully")
 
+	deletedAttr := types.NewAttribute("deleted", s.user1, types.AttributeType_String, []byte("test"))
 	// Create a name, make an attribute under it, then remove the name leaving an orphan attribute.
-	s.NoError(s.app.NameKeeper.SetNameRecord(s.ctx, "deleted", s.user1Addr, false), "name record should save successfully")
-	s.NoError(s.app.AttributeKeeper.SetAttribute(s.ctx, types.NewAttribute("deleted", s.user1, types.AttributeType_String, []byte("test")), s.user1Addr), "should save successfully")
-	s.NoError(s.app.NameKeeper.DeleteRecord(s.ctx, "deleted"), "name record should be removed successfully")
+	s.Assert().NoError(s.app.NameKeeper.SetNameRecord(s.ctx, "deleted", s.user1Addr, false), "name record should save successfully")
+	s.Assert().NoError(s.app.AttributeKeeper.SetAttribute(s.ctx, deletedAttr, s.user1Addr), "should save successfully")
```
iramiller, line 418: To prove out the increment/decrement flow there should be multiple updates here so that the usage counter is verified as matching the number of instances that exist.

x/attribute/types/keys.go
```diff
@@ -58,6 +60,29 @@ func AddrStrAttributesNameKeyPrefix(addr string, attributeName string) []byte {
...
+}
+
+// AttributeNameAddrKeyPrefix returns a prefix key for attribute and address
+func AttributeNameAddrKeyPrefix(attributeName string, addr []byte) []byte {
+	key := AttributeKeyPrefixAddrLookup
+	key = append(key, GetNameKeyBytes(attributeName)...)
+	return append(key, address.MustLengthPrefix(addr)...)
+}
+
+// GetAddressFromKey returns the AccAddress from full attribute address key ([prefix][name hash][length + AccAddress bytes][attribute hash])
```
iramiller, line 76: attribute hash is no longer present... comment is out of date

iramiller (commented):
Some minor questions

x/attribute/keeper/keeper.go
```diff
@@ -160,6 +162,36 @@ func (k Keeper) SetAttribute(
...
+	key := types.AttributeNameAddrKeyPrefix(attr.Name, attr.GetAddressBytes())
+	bz := store.Get(key)
+	id := uint64(0)
+	if bz != nil {
+		value := binary.BigEndian.Uint64(bz)
+		if value <= uint64(1) {
+			store.Delete(key)
+		} else {
+			id = binary.BigEndian.Uint64(bz) - uint64(1)
+			store.Set(key, sdk.Uint64ToBigEndian(id-1))
```
Taztingo, line 190: Are you purposely double decrementing? Also you may want to use value instead of `binary.BigEndian.Uint64(bz)`. If you have a value of 2 then id will be 1, and you will be setting the value to 0.

Taztingo (changes requested):

Taztingo (dismissed):
LGTM

x/attribute/keeper/keeper.go
nullpointer0x00: The update allows for an address change which would mean a different key from the original.  So, I thought it would be appropriate just to decrement the address count when the old attribute is removed and increment it when the updated one is added.  I could do a check to see if the update would change what the lookup key would be and then do a dec of the old key and inc for the new key.

x/attribute/keeper/keeper_test.go
nullpointer0x00: I added a test to add multiple attributes with same name to be deleted.  The test checks the before and after count.

x/attribute/keeper/keeper.go
iramiller: Makes sense!

iramiller (approved):
LGTM

Taztingo (approved):

Issue #1280, opened 2022-12-27
Enable ability for smart contracts to be owners of metadata entries.


SpicyLemon:
## Summary

Allow metadata scopes, sessions, and/or records to have owners that are smart contracts.

## Problem Definition

As a smart-contract creator, I want to be able to have my smart-contract own metadata entries so I can more easily manage updating information.

## Proposal

1. In the `x/metadata` wasm encoders, add the smart contract address as the first signer of the message being sent.
2. In the `x/metadata` msg validation, make sure that the first signer is one of the required signers.
3. In our `wasmd` fork, update `handleSdkMessage` to only check the first address from `GetSigners()`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
It looks like, for 3, we could instead use the `WithMessageHandler` option and define the handler in the provenance repo instead. Since you've already got it in our wasmd fork, though, we can probably use that just fine. It'd be nice to get off our fork at some point, and go back to using the standard repo. If we ever do that, we'll need to go with the `WithMessageHandler` way, but until then.... eh.

PR #1473, opened 2023-04-14
Require smart contract signers to either be a party or be authorized by other signers.

SpicyLemon:
## Description

closes: #1280

This makes it so that, for metadata messages, if a smart contract is a signer, one of the following must be satisfied:
* They are a party/owner of the thing being written.
* All signers after it have granted authorizations (via authz) to it for the message type. The last signer cannot be a smart contract address unless they are a party/owner.

This also:
* Updates the non-rollup validation to enforce the provenance/smart-contract role.
* Fixes rollup provenance/smart-contract role validation to correctly look at the proposed parties.

---

write a little note why.

### DIFF
--- a/x/metadata/keeper/scope.go
+++ b/x/metadata/keeper/scope.go
@@ -323,7 +323,7 @@ func (k Keeper) ValidateWriteScope(
 			}
 			// Note: This means that a scope can be initially written without consideration for signers and roles.
 			if existing != nil {
-				if validatedParties, err = k.ValidateSignersWithParties(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
+				if validatedParties, err = k.validateAllRequiredPartiesSigned(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
 					return err
 				}
 			} else {
@@ -381,7 +381,7 @@ func (k Keeper) ValidateDeleteScope(ctx sdk.Context, msg *types.MsgDeleteScopeRe
 				return err
 			}
 		} else {
-			if validatedParties, err = k.ValidateSignersWithParties(ctx, scope.Owners, scope.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
+			if validatedParties, err = k.validateAllRequiredPartiesSigned(ctx, scope.Owners, scope.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
 				return err
 			}
 		}
@@ -445,7 +445,7 @@ func (k Keeper) ValidateAddScopeDataAccess(
 		if !found {
 			return fmt.Errorf("scope specification %s not found", existing.SpecificationId)
 		}
-		if _, err := k.ValidateSignersWithParties(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
+		if err := k.ValidateSignersWithParties(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
 			return err
 		}
 	}
@@ -497,7 +497,7 @@ dataAccessLoop:
 		if !found {
 			return fmt.Errorf("scope specification %s not found", existing.SpecificationId)
 		}
-		if _, err := k.ValidateSignersWithParties(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
+		if err := k.ValidateSignersWithParties(ctx, existing.Owners, existing.Owners, scopeSpec.PartiesInvolved, msg); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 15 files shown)


x/metadata/spec/01_concepts.md
```diff
@@ -95,6 +95,16 @@ If the value owner address is changing as well as one or more other fields, thes
 * When a value owner address is a non-marker address, and is being changed, that existing address must be one of the signers.
 * When a value owner address is empty, and is being changed, standard scope signer requirements are also applied even if that's the only change to the scope.
 
+### Smart Contract Requirements
+
+The following are requirements related to smart contract usage of the `x/metadata` module:
+
+* A party with a smart contract address MUST have the `PROVENANCE` role.
+* A party with the `PROVENANCE` role MUST have the address of a smart contract.
+* When a smart contract signs a message, it MUST not have any non-smart-contract signers before it and SHOULD include the invoker address(es) after.
```
Taztingo, line 104: Might be easier to understand if reworded...

When a smart contract signs a message, **it MUST have only smart-contract signers** before it and SHOULD include the invoker address(es) after.

iramiller: the only reason to not include the invoker is if the smart contract is the sole owner/manager of the scope/records.

iramiller (dismissed):
minor comments... seems that the "wasm" definition here is a bit too narrow because technically any system or module account could satisfy the no public key/zero sequence check.

x/metadata/spec/01_concepts.md
SpicyLemon: Much better. Updated.

x/metadata/keeper/signers.go
```diff
@@ -338,29 +352,81 @@ func (k Keeper) validateProvenanceRole(ctx sdk.Context, parties []*PartyDetails)
 	return nil
 }
 
+// isWasmAccount returns true if the provided addr is the address of a smart contract account.
+// A smart contract account is a BaseAccount that exists, has a sequence of 0 and does not have a public key.
+func (k Keeper) isWasmAccount(ctx sdk.Context, addr sdk.AccAddress) bool {
```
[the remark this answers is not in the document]
SpicyLemon: A marker address wouldn't return `true` because it's not a `BaseAccount`. However, something like the gov module account would return true. I'm not sure of another way to test if it's a smart contract though. In the wasm module, this same check is done. I can't remember if they limit it to only `BaseAccount`s though.

I guess, if a false positive starts causing problems, we can look deeper into it. Last I knew, the wasm module didn't have an "is this a smart contract address" function. We could try getting it, but that's a whole lot of reading that isn't needed.

iramiller (approved):
nice verification on the account types.

x/metadata/keeper/scope.go
```diff
@@ -528,7 +554,14 @@ func (k Keeper) ValidateUpdateScopeOwners(
 		if err := validateRolesPresent(proposed.Owners, scopeSpec.PartiesInvolved); err != nil {
```
Taztingo, line 554: I just noticed that lines 537-542 are equivalent to 554-559. Are we able to pull them out of the branch?

x/metadata/keeper/scope_test.go
```diff
@@ -1370,6 +1401,41 @@ func (s *ScopeKeeperTestSuite) TestValidateScopeUpdateOwners() {
...
 		},
+		{
+			name:     "smart contract without provenance role added",
+			existing: scopeWithOwners(ownerPartyList(s.user1)),
+			proposed: scopeWithOwners(ownerPartyList(s.user1, s.scUser)),
+			signers:  []string{s.user1},
+			errorMsg: `account "` + s.scUser + `" is a smart contract but does not have the PROVENANCE role`,
+		},
+		{
+			name:     "smart contract without provenance role removed",
```
Taztingo, line 1412: Can't wrap my head around if this test should have an error or not. I was initially thinking it should because every smart contract is supposed to have a Provenance Role, but it might be safer to allow removals.

x/metadata/keeper/scope_test.go
```diff
@@ -1370,6 +1401,41 @@ func (s *ScopeKeeperTestSuite) TestValidateScopeUpdateOwners() {
...
+		},
+		{
+			name:     "with rollup smart contract without provenance role added",
+			existing: rollupScopeWithOwners(pt(s.user1, owner, false)),
+			proposed: rollupScopeWithOwners(pt(s.user1, owner, false), pt(s.scUser, owner, true)),
+			signers:  []string{s.user1},
+			errorMsg: `account "` + s.scUser + `" is a smart contract but does not have the PROVENANCE role`,
+		},
+		{
+			name:     "with rollup smart contract without provenance role removed",
```
Taztingo, line 1433: Same as above.

Taztingo (changes requested):

x/metadata/keeper/scope_test.go
SpicyLemon: This one is basically checking that the validation is NOT run on the existing entry. E.g. if an existing scope has a smart contract in a party that isn't provenance, it can be updated to either remove it or switch it to provenance.

x/metadata/keeper/scope_test.go
SpicyLemon: Same here. Only the proposed should matter for the provenance/smart-contract check.

This one threw for a second too, though. I thought it might need to be signed by the smart contract too, but the smart contract is optional, so it's not required as a signer.

x/metadata/keeper/scope.go
SpicyLemon: Good call. I pulled them out here and also up in ValidateWriteScope.

Taztingo (approved):

iramiller (approved):

Issue #1329, opened 2023-01-30
New endpoint for updating a scope's value owner address.


SpicyLemon:
## Summary

Create a new `UpdateValueOwner` endpoint for updating a scope's value owner address field.

## Problem Definition

As a scope owner, I want to be able to update its `value_owner_address` field, without needing to make a call to `WriteScope`, in order to make it easier to do and also not have to pay the `WriteScope` msg-based fee.

As the owner of many scopes, I want to be able to update their `value_owner_address` fields all to the same value, without needing to make a call to `WriteScope`, in order to make it easier and cheaper to do so.

## Proposal

Create an `UpdateValueOwner` endpoint in the metadata module. It should take in a list of scope ids, new value owner, and list of signers. It doesn't need to output anything. Signer validation needs to be the same as when using `WriteScope` to only update the `value_owner_address` field. Validation should be done only once for each unique existing value owner, and once for the new value owner (as opposed to once per scope).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
If you are merging loan pools it probably makes sense to be able to move all scopes with one value_owner into another value_owner... This probably makes sense as another similar endpoint. (merge value owners)

PR #1523, opened 2023-04-29
Create UpdateValueOwners and MigrateValueOwner endpoints

SpicyLemon:
## Description

Create an `UpdateValueOwners` endpoint.

closes: #1329 

---

write a little note why.


nullpointer0x00 (approved):

iramiller (dismissed):
The only item that caught my eye was switching the index from string to accaddress ...

SpicyLemon:
> The only item that caught my eye was switching the index from string to accaddress ...

The indexes didn't actually change. All that changed with the indexing is _when_ the strings are converted to acc addresses. That was originally done in `scopeIndexValues.IndexKeys`, but I moved it to the `scopeIndexValues` constructor. And now that I look back on it, none of those changes ended up being needed. Originally, I was going to try to save a bit of processing by trying to skip some index updates if possible. But it ended up being a tricky thing to do and would have been likely to cause bugs later. So I went back to having it just do the standard indexing. But the external behavior of the `scopeIndexValues`s and `indexScope` didn't change at all.

x/metadata/keeper/msg_server.go
```diff
@@ -178,6 +178,66 @@ func (k msgServer) DeleteScopeOwner(
 	return types.NewMsgDeleteScopeOwnerResponse(), nil
 }
 
+func (k msgServer) UpdateValueOwners(
```
nullpointer0x00, line 181: Should we be adding the comment for the method name in msg_server.  I know they are missing all over our code base.  I have been trying to be a better about adding them myself (I sometimes forget though).

SpicyLemon: Heh. I haven't been adding them since non of the other ones ever had them, but I prefer having function comments. So I'll add some.

SpicyLemon: Done. I copied the comments from the tx.proto onto their `MsgServer` funcs in here.

I'm guessing that we usually didn't have them since the `msgServer` struct is private and implements the metadata `MsgServer` interface defined automatically from the protos. And that interface has the comments on the functions.

But I like having them in `msg_server.go` too.

iramiller (approved):
This signing flow stuff is getting exceedingly complex...

SpicyLemon:
> This signing flow stuff is getting exceedingly complex...

Yeah. Lots of logic that shouldn't be duplicated, but needs to be slightly different in a couple circumstances. There's also added code complexity due to attempts to minimize processing. So the requirements are pretty complex, and then there's added complexity from optimizations. :/

nullpointer0x00 (approved):

Issue #1546, opened 2023-05-15
Allow forced_transfer to be set on existing markers using governance


iramiller:
## Summary

Add the ability for a governance proposal to be submitted which can change the `forced_transfer` setting on an existing `RESTRICTED` marker.

## Problem Definition

The forced transfer capability is a new feature which allows markers to be created with a transfer agent that has complete control of the token movement regardless of where it is stored.  As this setting breaks the rules of an account requiring its signature before moving funds this option can only be set when a marker is initially created so users acquiring the token can be aware of the terms of its possession.

This request would add the ability for a marker that allows governance control and is set to `RESTRICTED` to have the `forced_transfer` flag changed when a governance proposal request is passed.

## Proposal

Create a new `msg` that can be dispatched by the governance module (signed by the governance module as the authority) which allows the `forced_update` setting on an existing marker to be changed.  This request should only be allowed for `RESTRICTED` markers that have `allow_governance` control enabled.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1547, opened 2023-05-16
Create marker UpdateForceTransfer endpoint

SpicyLemon:
## Description

closes: #1546 

Create a new `UpdateForceTransfer` endpoint so that the `allow_force_transfer` field can be changed via governance proposal.

---

write a little note why.


x/marker/keeper/msg_server.go
```diff
@@ -694,3 +694,39 @@ func (k msgServer) UpdateRequiredAttributes(goCtx context.Context, msg *types.Ms
...
+
+	if marker.GetMarkerType() != types.MarkerType_RestrictedCoin {
+		return nil, fmt.Errorf("cannot update forced transfer on unrestricted marker %s", msg.Denom)
+	}
+
+	if !marker.HasGovernanceEnabled() {
+		return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
+	}
+
+	if marker.AllowsForcedTransfer() == msg.AllowForcedTransfer {
```
nullpointer0x00, line 717: I'm on the  fence about this of the force transfer flag.  Do we really care at this point?  I'm just wondering about the case that the state changes between when the gov proposal is submitted and when it runs.  The delayed setting of this state make this a bit problematic.  If this was just a normal tx with immediate feedback this would be fine imo.

iramiller: Since the flag can only change via governance proposal ... this failure would indicate an invalid proposal (setting something already set, user probably made a mistake) ... or perhaps as you mention, two different overlapping governance proposals (one of these shouldn't have happened) ... also feels like a user mistake/error...

SpicyLemon: I waffled on this too.

One one hand, it's similar behavior as most of our stuff.

One the other hand, the reason for a gov prop failure can be a bit of a pain to track down. If a prop fails for this reason, it's kind of a false flag. I.e. it shows as a failure even though state is as the proposal requested.

In the end, I went with our standard behavior. I was really on the fence about it though, and it'd be easy to take out if we decide we don't want that check in this endpoint.

x/marker/types/codec.go
```diff
@@ -12,21 +14,11 @@ import (
...
-		&MsgWithdrawRequest{},
-		&MsgTransferRequest{},
-		&MsgIbcTransferRequest{},
-		&MsgSetDenomMetadataRequest{},
-	)
+	messages := make([]proto.Message, len(allRequestMsgs))
+	for i, msg := range allRequestMsgs {
+		messages[i] = msg
+	}
+	registry.RegisterImplementations((*sdk.Msg)(nil), messages...)
```
nullpointer0x00, line 21: I like this a lot.  Since, adding a change to both places sometimes gets missed.

nullpointer0x00 (approved):

iramiller (approved):

Issue #1539, opened 2023-05-11
Bump go to 1.20


SpicyLemon:
## Summary

Bump all our stuff to use go 1.20 to build provenance.

## Problem Definition

All our stuff using go 1.18 currently, which is no longer supported.

However, the `provenanced` executable generated using go 1.18 is not state-compatible with the executables built by go 1.19 or 1.20. This is demonstrated in [a unit test in our sdk fork](https://github.com/provenance-io/cosmos-sdk/blob/52688d71541d182225b056578f0d5b7388e6dd7f/store/rootmulti/snapshot_test.go#L113) that fails if using go 1.19 or 1.20, but passes when using 1.18.

## Proposal

1. Update our stuff to use go 1.20.
2. Make sure that an upgrade from 1.15 (built using go 1.18) to the new version (built using 1.20) is successful.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1550, opened 2023-05-17
Bump go to 1.20 and linter to v1.52.2

SpicyLemon:
## Description

closes: #1539

1. Bump Go version to 1.20.
2. Bump golangci-lint to v1.52.2 and fix all the new linter errors.
3. Created some make targets for installing and updating `golangci-lint`.
    * `make golangci-lint` will install `golangci-lint` at the proper version, but only if `golangci-lint` isn't yet installed.
    * `make golangci-lint-upgrade` will install `golangci-lint` at the proper version even if `golangci-lint` is already installed.

New linter errors were all one of two things (both from the `revive` linter):
1. It wants unused function arguments to be `_` instead of having names.
2. It doesn't like when you `if err != nil { return err }; return nil`. It just wants you to `return err`.

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * Add usage of `AddGovPropFlagsToCmd` and `ReadGovPropFlags` cli for `GetModifyNameCmd` [#1542](https://github.com/provenance-io/provenance/issues/1542).
 * Bump Cosmos-SDK to `v0.46.10-pio-4` (from `v0.46.10-pio-3`) for the `SendRestrictionFn` changes [PR 1506](https://github.com/provenance-io/provenance/pull/1506).
 * Switch to using a `SendRestrictionFn` for restricting sends of marker funds [PR 1506](https://github.com/provenance-io/provenance/pull/1506).
+* Create `rust` upgrade handlers [PR 1549](https://github.com/provenance-io/provenance/pull/1549).
 
 ### API Breaking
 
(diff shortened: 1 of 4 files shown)


SpicyLemon:
Version `v0.46.10` of the SDK has a unit test that passes if using go 1.18, but fails if using go 1.20. The failed test relates to a hash that changes. So I thought it would be prudent to test an upgrade from a v1.15 provenance chain to one using go 1.20.

Here's what I did (all locally):
1. `make clean`
2. Made sure I was using go `1.18`.
3. I checked out `v1.15.0` and did `make build`, then moved it to `build/provenanced-1.18`.
4. I switched to this branch and switched to go `1.20`.
5. I added an empty upgrade handler.
6. `make build`, then moved it to `build/provenanced-1.20`.
7. I created a `build/provenanced` symlink pointing to `build/provenanced-1.18`.
8. For good measure, I switched back to go `1.18`. This shouldn't matter since the thing was already compiled, but whatever.
9. I did `make build run-config` then `./scripts/upgrade-config.sh`.
10. `make run`.
11. In another window, I used the `./scripts/upgrade-test.sh` script to schedule an the empty upgrade I had added.
12. I waited for the upgrade height to hit and the chain to halt.
13. I switched back to go `1.20` (again, probably didn't matter).
14. I switched the `build/provenanced` to `build/provenanced-1.20`.
15. `make run` again.

The no-op upgrade "ran" and the chain started back up again just fine.

If I find the time, I'll redo the test and add some more chain data between steps 10 and 11. But I'm confident that it'll be fine.

The changing hash had to do with the behavior of a `Close()` function somewhere deep in everything. In one version it was getting called an extra time which was adding some extra bytes somewhere.

Taztingo (approved):
LGTM

Issue #1557, opened 2023-05-23
Review and refactor attribute iterators on `update` and `delete` keeper methods


nullpointer0x00:
## Summary

In the `attribute` module we are using an iterator for the `UpdateAttribute` call.  This could be removed and just use the key from the address, name, and value to get the attribute from the store.

We are also modifying data from within an iterator in the `DeleteAttribute` method.  We should remove the the modifying behavior to outside the iterator.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1560, opened 2023-05-24
Refactor attribute update and delete iterators

nullpointer0x00:
## Description

Update iterators in attribute keeper to not do modifications to store within the iterator. 

closes: #1557 

---

write a little note why.


x/attribute/keeper/keeper.go
```diff
@@ -383,38 +383,42 @@ func (k Keeper) DeleteAttribute(ctx sdk.Context, addr string, name string, value
...
 
 		if attr.Name == name && (!deleteDistinct || bytes.Equal(*value, attr.Value)) {
-			count++
-			store.Delete(it.Key())
-			k.DecAttrNameAddressLookup(ctx, attr.Name, attr.GetAddressBytes())
-			k.deleteAttributeExpireLookup(store, attr)
+			attrToDelete = append(attrToDelete, attr)
+		}
+	}
+	it.Close()
```
iramiller, line 397: Because there is a return on err statement above this close might not get called.  This should be a defer.

nullpointer0x00: Good catch. Fixed.

x/attribute/keeper/keeper.go
```diff
@@ -382,39 +382,50 @@ func (k Keeper) DeleteAttribute(ctx sdk.Context, addr string, name string, value
...
+				return err
 			}
 		}
 	}
+
 	errm := "no keys deleted"
-	if count == 0 && deleteDistinct {
+	if len(attrToDelete) == 0 && deleteDistinct {
 		ctx.Logger().Error(errm, "name", name, "value")
 		return fmt.Errorf("%s with name %s value %s", errm, name, string(*value))
```
SpicyLemon, line 427: We probably want `%q` for the `string(*value)`.

```suggestion
		return fmt.Errorf("%s with name %s value %q", errm, name, string(*value))
```

x/attribute/keeper/keeper.go
```diff
@@ -439,10 +450,15 @@ func (k Keeper) PurgeAttribute(ctx sdk.Context, name string, owner sdk.AccAddres
 		return err
 	}
 	for _, acct := range accts {
+		attrToDelete := [][]byte{}
 		store := ctx.KVStore(k.storeKey)
```
SpicyLemon, line 454: This should be moved outside the loop. There's no reason to create it again for each account.

x/attribute/keeper/keeper.go
```diff
@@ -439,10 +450,15 @@ func (k Keeper) PurgeAttribute(ctx sdk.Context, name string, owner sdk.AccAddres
...
 		it := sdk.KVStorePrefixIterator(store, types.AddrAttributesNameKeyPrefix(acct, name))
 		for ; it.Valid(); it.Next() {
-			store.Delete(it.Key())
+			attrToDelete = append(attrToDelete, it.Key())
+		}
+		it.Close()
```
SpicyLemon, line 459: This needs to be extracted to its own function so that you can ensure the iterator is always properly closed after each use. The reason you always want a `defer` that closes the iterator, is so that it gets closed even if there's a panic.

But because this is in a for loop, those defers wouldn't be called until much later. It'd probably also be safe to do the `if it != nil { it.Close() }` way on it and leave it in the loop, but it's probably cleaner to just extract it to its own function. I'd have the function take in a `KVStore` instead of the `ctx` too, so you only need to create that store once.

x/attribute/keeper/keeper.go
nullpointer0x00: I thought it made more sense here because I needed to do the `DecAttrNameAddressLookup` call while in the context of having the acct.  Anyway, the point is moot now since I have refactored this area.

x/attribute/keeper/keeper.go
```diff
@@ -438,18 +449,32 @@ func (k Keeper) PurgeAttribute(ctx sdk.Context, name string, owner sdk.AccAddres
...
+	defer func() {
+		if it != nil {
+			it.Close()
+		}
+	}()
+	for ; it.Valid(); it.Next() {
+		attrToDelete = append(attrToDelete, it.Key())
+	}
+	it.Close()
+	it = nil
```
SpicyLemon, line 474: No need to deal with `it != nil` here since there's nothing being done in this function after we're done using it.

```suggestion
	defer it.Close()
	for ; it.Valid(); it.Next() {
		attrToDelete = append(attrToDelete, it.Key())
	}
```

x/attribute/keeper/keeper.go
```diff
@@ -245,22 +245,23 @@ func (k Keeper) UpdateAttribute(ctx sdk.Context, originalAttribute types.Attribu
...
 	addrBz := originalAttribute.GetAddressBytes()
+	attrKey := types.AddrAttributeKey(addrBz, originalAttribute)
+	currentAttr := store.Get(attrKey)
 
-	store := ctx.KVStore(k.storeKey)
-	it := sdk.KVStorePrefixIterator(store, types.AddrAttributesNameKeyPrefix(addrBz, normalizedOrigName))
-	defer it.Close()
 	var found bool
-	for ; it.Valid(); it.Next() {
+	if currentAttr != nil {
```
SpicyLemon, line 254: Probably safer to use `len(currentAttr) > 0` here.

It's best to use that check for usability of a slice unless it's a situation where you want to treat a nil slice differently from an empty one. You can't always count on functions to return nil instead of an empty slice (or vice versa), and usually we only care if it returned something. So it's best practice to `len(thing) > 0` instead of `thing != nil`

x/attribute/keeper/keeper.go
```diff
@@ -438,18 +449,26 @@ func (k Keeper) PurgeAttribute(ctx sdk.Context, name string, owner sdk.AccAddres
...
+		for _, key := range attrToDelete {
+			store.Delete(key)
 			k.DecAttrNameAddressLookup(ctx, name, acct)
 		}
-		it.Close()
 	}
 	return nil
 }
 
+func (k Keeper) getAccountsToDelete(store sdk.KVStore, acctAddr sdk.AccAddress, attributeName string) (attrToDelete [][]byte) {
```
SpicyLemon, line 463: Probably want this called `getAttrsToDelete`

x/attribute/keeper/keeper.go
nullpointer0x00: I was thinking that the `nil` return here is actually a not found.  
```
// Get returns nil if key doesn't exist. Panics on nil key.
	Get(key []byte) []byte
```

A return of an empty slice would mean the key did exist with an empty slice.  This should hit the error here: 
```
attr := types.Attribute{}
		if err := k.cdc.Unmarshal(currentAttr, &attr); err != nil {
			return err
		}
```

This would indicate something bad happened else where in the code to have set this to an empty slice.  The error here would be helpful.  `len(currentAttr) > 0` would just hide it.

x/attribute/keeper/keeper.go
nullpointer0x00: Ooops, I was actually going to refactor that but forgot.  I named it to `getAccountAttributesByName` to be more specific to the actual method.  `getAttrsToDelete`  is specific to an expected caller... This method might be useful for something else in the future.

x/attribute/keeper/keeper.go
Taztingo: I just did a quick look through the codebase for the attribute module, and I don't **think** I found a case where the currentAttr could be 0. It may make the code a bit easier to understand if we use **store.Has** or a **nil check** because checking the length will imply that we store empty attributes somewhere.

Taztingo (approved):

SpicyLemon (approved):

Issue #1552, opened 2023-05-18
Add a basic metadata attribute tied to marker address


iramiller:
## Summary

Add support for a json document linked directly against an address using an attribute.

## Problem Definition

One of the common requests from users is the ability to attach some basic context information against an address to provide additional background.  This data could include information that would generally be associated with basic digital assets (such as links to off chain data) or Opensea specifications (https://docs.opensea.io/docs/metadata-standards).  

While the metadata module provides the ability to represent programmable digital assets that involve off-chain processes and multiple parties, often a single self contained digital asset is sufficient for modeling.

## Proposal

Using a new protocol owned name from the name module, create a standardized attribute that can be written by the protocol against any address.  As the attribute can only be written by the owner of the name the protocol will need to offer methods for reading and writing this particular attribute.

Using this standard metadata attribute add methods for writing this same data against `scope` and `marker` addresses.

Data being recorded should be in `JSON` format in order to promote the most accessible and interoperable standard for its use.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Just getting started on this.

I'm not exactly sure what to call it. I feel like "metadata" is a poor name choice due to both the `x/metadata` module and `denom metadata`. But the only thing I can think of right now is "Account Metadata."

That being said, here's what I'm thinking.

1. Create a new root name owned by the attribute module (or maybe the name module?). Probably `account.metadata`, but maybe just `metadata`.
2. Add an endpoint to the attribute module for normal accounts to set their metadata, e.g. `SetAccountMetadata`. It'd take in the address (signer) and the value.
3. Add an endpoint to the marker module for an admin(?) to set a marker's account metadata. I'm not sure what to call this, though since "metadata" might be easily confused with "denom metadata". But maybe `SetMarkerAccountMetadata`. It'd take in the denom, signer, and the value. I'm not sure what permission the signer would need though (admin? deposit? mint?).
4. Add an endpoint to the metadata module for setting a scope's, erm... account metadata. Possibly `SetScopeAccountMetadata`. It'd take in the scope id, signer(s) and the value. I'm guessing the signer(s) would be validated against only the value owner, but I'm not sure if that's correct.

---

So here's some outstanding questions:
1. What should I name the attribute? `account.metadata`, `metadata` or something else?
[...]

PR #1561, opened 2023-05-25
Create an attribute-owned accountdata name and endpoints for managing it

SpicyLemon:
## Description

closes: #1552 

1. Create the `accountdata` owned by the attribute account.
2. Add endpoints to the attribute module for getting/setting an account's accountdata.

---

write a little note why.


x/name/keeper/keeper.go
SpicyLemon, line 1: Once #1565 has merged and I then update this branch, the sims _should_ start passing again.

x/marker/client/cli/tx.go
```diff
@@ -544,9 +546,9 @@ func GetNewTransferCmd() *cobra.Command {
 	return cmd
 }
 
-// TODO: refactor usage comments to be provenance specific
 // GetIbcTransferTxCmd returns the command to create a GetIbcTransferTxCmd transaction
 func GetIbcTransferTxCmd() *cobra.Command {
+	// TODO: refactor ibc-transfer usage comments to be provenance specific
```
Taztingo, line 551: Should this just be an issue?

SpicyLemon: I'm not even sure what the TODO means. I just know that where it was, wasn't the correct place according to godoc standards. Godoc standards say the comment block before a function should start with the function name. The option was to either swap the two comment lines or move the TODO into the function. I chose the latter since it's not information that's worthwhile to have in the function's godoc/intellisense/info.

x/attribute/keeper/keeper.go
```diff
@@ -600,3 +601,48 @@ func (k Keeper) ValidateExpirationDate(ctx sdk.Context, attr types.Attribute) er
...
+		return fmt.Errorf("could not look up existing %s for %q: %w", types.AccountDataName, addr, err)
+	}
+	if len(existings) > 0 {
+		err = k.DeleteAttribute(ctx, addr, types.AccountDataName, nil, k.modAddr)
+		if err != nil {
+			return fmt.Errorf("could not delete existing %s for %q: %w", types.AccountDataName, addr, err)
+		}
+	}
+
+	// Just leave it deleted if the new value is an empty string.
```
Taztingo, line 633: I'm not sure how this is being used by the end user, but would they ever want an entry with an empty string? I'm guessing probably not since it would just be a waste of money.

x/marker/client/cli/tx.go
Taztingo: Whoops, I thought it was just added.

Taztingo (commented):
Overall LGTM and it had a lot of nice cleanup. I haven't approved it yet because I want to quickly go through all the tx flows again without having to worry about other stuff.

Taztingo (approved):
LGTM

x/marker/keeper/msg_server.go
```diff
@@ -734,3 +737,30 @@ func (k msgServer) UpdateForcedTransfer(goCtx context.Context, msg *types.MsgUpd
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get %s marker: %w", msg.Denom, err)
+	}
+
+	if msg.Signer == k.GetAuthority() {
+		if !marker.HasGovernanceEnabled() {
+			return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
+		}
+	} else {
+		if !marker.HasAccess(msg.Signer, types.Access_Deposit) {
```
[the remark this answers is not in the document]
SpicyLemon: I didn't use `admin` because that's supposedly only for the ability to manage permissions on a marker. It's not exactly named the best thing though primarily due to confusion like this.

The `deposit` permission is what's used in the metadata module when setting a marker as the scope's value owner. Someone signing the message for that update has to have `deposit` permission on that marker. Since that's kind of like associating data with a marker, it felt like the best fit for a marker's account data too.

It would also make sense if `withdraw` permission were needed to set a marker's account data to `""`, but I opted to keep it simple. That might be worth revisiting later, though, depending on users' needs.

SpicyLemon:
As part of this PR, I shortened the block time for the CLI tests from 2 seconds to 0.5 seconds. I just wanted to point out that the "Tests / Code Coverage" Github Action/workflow ran in 6m54s. That has usually been taking about 12 minutes. I'm really glad to see that.

x/marker/client/cli/cli_test.go
```diff
@@ -267,8 +270,23 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.Require().NoError(err)
 	genesisState[markertypes.ModuleName] = markerDataBz
 
+	// Pre-define an accountdata entry
+	attrData := attrtypes.DefaultGenesisState()
+	attrData.Attributes = append(attrData.Attributes,
+		attrtypes.Attribute{
+			Name:          attrtypes.AccountDataName,
+			Value:         []byte("Do not sell this coin."),
```
iramiller, line 278: ❤️

x/marker/keeper/msg_server.go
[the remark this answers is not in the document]
iramiller: In this context "deposit" is used for that permission flow because we are "depositing" the NFT into the marker as an asset.  Admin == Grant originally... (but was renamed away from that because grant-grants-grants was exceptionally confusing in naming).

 (╯°□°）╯︵ ┻━┻

iramiller (approved):
Overall this seems like a solid feature.  I am conflicted about the Marker/Deposit flow (as I commented on the resolved thread, deposit in the metadata side was because we were "adding a digital asset to the marker" like a pool... but yeah, there simply isn't a great role for general maintenance/ownership of the marker itself because for the most part that data is fixed on creation.

x/attribute/types/keys.go
```diff
@@ -23,6 +23,9 @@ const (
 
 	// QuerierRoute is the querier route for account
 	QuerierRoute = ModuleName
+
+	// AccountDataName is the name of the attribute used to store account data.
+	AccountDataName = "accountdata"
```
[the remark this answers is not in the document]
SpicyLemon: I thought about something like `pio` but wasn't exactly sure what to call it. It also added that tiny bit of complexity. I was also worried about competing `accountdata.<something else>` entries. I felt having it be a root name just made it a little easier and safer.

I figured that, if we end up with a couple more protocol owned ones, we could then re-evaluate and possibly add a `pio` for them.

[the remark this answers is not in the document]
iramiller: Good point. Less complexity now, more later if needed.

iramiller (approved):

Taztingo (approved):

Issue #1556, opened 2023-05-23
Inactive validator delegation cleanup process


iramiller:
## Summary

There are many inactive validators on our network that have small amounts of stake delegated to them.  We should add a process that runs on an interval and issues a batch of unbond requests for delegators that are delegated against validators that have been offline for three weeks or more.

## Problem Definition

Our inactive validator list contains many entries for validators that have been offline for a long time and are not coming back.  The chain can not remove these nodes until they have no more delegations against them.  In many cases these users are likely unaware that they have delegations against offline nodes.

## Proposal

Build a short process that runs against the list of offline validators and looks for entries that are more than `unbonding period` (3 weeks) old and then iterates their list of delegators to issue unbonding requests against any that are still bonded.

When this process has been in place for a while it will allow the list of inactive validators to be substantially reduced as well as providing a useful service for delegators. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1571, opened 2023-06-01
Inactive validator removal upgrade handler method

nullpointer0x00:
## Description

closes: #1556 

---

write a little note why.


app/upgrades.go
```diff
@@ -230,6 +232,41 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
 	}
 }
 
+// removeInactiveValidators unbonds all delegations from inactive validators, triggering their removal from the validator set.
+func removeInactiveValidators(ctx sdk.Context, app *App) {
+	unbondingTimeParam := app.StakingKeeper.GetParams(ctx).UnbondingTime
+	ctx.Logger().Info(fmt.Sprintf("removing any validator that has been inactive (unbonded) for %d hours", int64(unbondingTimeParam.Hours()/24)))
```
iramiller, line 238: status message says hours ... but you are dividing the hours param by 24?

app/upgrades.go
```diff
@@ -230,6 +232,41 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
...
+				for _, delegation := range delegations {
+					ctx.Logger().Info(fmt.Sprintf("undelegate delegator %v from validator %v of all shares (%v)", delegation.DelegatorAddress, validator.OperatorAddress, delegation.GetShares()))
+					// unbonding all of a delegators shares, method will remove validator from set once all have been removed.
+					_, err := app.StakingKeeper.Undelegate(ctx, delegation.GetDelegatorAddr(), valAddress, delegation.GetShares())
+					if err != nil {
+						panic(err)
+					}
+				}
+				// this should probably never be true, an unbonded delegator will be removed when the last delegator has called Unbond
+				if len(delegations) == 0 && validator.DelegatorShares.IsZero() {
```
iramiller, line 260: I would rather we left this out ... because if the delegator is still there once everything is unbounded that indicates a different issue and this clean up might mask it.

app/upgrades.go
```diff
@@ -230,6 +232,41 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
...
+					app.StakingKeeper.RemoveValidator(ctx, valAddress)
+				}
+				removalCount++
+			}
+		}
+	}
+	ctx.Logger().Info(fmt.Sprintf("a total of %d inactive (unbonded) validators have been removed", removalCount))
```
iramiller, line 267: more correctly this code isn't intended to "remove unbonded validators" ... but to unbond delegations from stale/unbonded validators.

nullpointer0x00: Ok, I changed the text on that along with a few other refactors to keep things consistent.

app/upgrades.go
nullpointer0x00: Thanks for input on this.  I added that not knowing if it was what we wanted to do.  That is why I left the comment explaining it.  I will remove this.

app/upgrades.go
nullpointer0x00: Good catch!

app/upgrades.go
```diff
@@ -230,6 +232,37 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
...
+			if inActiveDuration >= unbondingTimeParam {
+				ctx.Logger().Info(fmt.Sprintf("validator %v has been inactive (unbonded) for %d days and will be removed", validator.OperatorAddress, int64(inActiveDuration.Hours()/24)))
+				valAddress, err := sdk.ValAddressFromBech32(validator.OperatorAddress)
+				if err != nil {
+					panic(err)
+				}
+				delegations := app.StakingKeeper.GetValidatorDelegations(ctx, valAddress)
+				for _, delegation := range delegations {
+					ctx.Logger().Info(fmt.Sprintf("undelegate delegator %v from validator %v of all shares (%v)", delegation.DelegatorAddress, validator.OperatorAddress, delegation.GetShares()))
+					// unbonding all of a delegators shares, method will remove validator from set once all have been removed.
```
iramiller, line 253: this comment reads weird with the new streamlined implementation. 
```suggestion

```

app/upgrades.go
```diff
@@ -88,6 +88,8 @@ var upgrades = map[string]appUpgrade{
 
 			removeP8eMemorializeContractFee(ctx, app)
 
+			removeInactiveValidatorDelegations(ctx, app)
```
SpicyLemon, line 91: This needs to also be in the `rust-rc1` handler.

iramiller (approved):
We could just cut that one weird comment out and be good I think... some of those log messages are a bit odd as well but not a huge deal.

app/upgrades.go
nullpointer0x00: I'm open to suggestion on changing them.  I went back and forth between a lot of different messages.

app/upgrades.go
```diff
@@ -230,6 +232,37 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
...
+					panic(err)
+				}
+				delegations := app.StakingKeeper.GetValidatorDelegations(ctx, valAddress)
+				for _, delegation := range delegations {
+					ctx.Logger().Info(fmt.Sprintf("undelegate delegator %v from validator %v of all shares (%v)", delegation.DelegatorAddress, validator.OperatorAddress, delegation.GetShares()))
+					// unbonding all of a delegators shares, method will remove validator from set once all have been removed.
+					_, err := app.StakingKeeper.Undelegate(ctx, delegation.GetDelegatorAddr(), valAddress, delegation.GetShares())
+					if err != nil {
+						panic(err)
+					}
```
SpicyLemon, line 258: If an undelegate fails, maybe it's better to just log it and move on. Sure the validator won't get removed (at least this time), but this cleanup wouldn't then be blocking us. Same goes for the above panic too.

Also, I'm not exactly sure what the difference between a panic and error are during an upgrade, but the pattern is usually to return an error instead of panicking. I'd guess that eventually it panics, with the error, but there might be other important things happening.

app/upgrades.go
```diff
@@ -71,6 +71,8 @@ var upgrades = map[string]appUpgrade{
 
 			removeP8eMemorializeContractFee(ctx, app)
 
+			removeInactiveValidators(ctx, app)
```
SpicyLemon, line 74:
```suggestion
			removeInactiveValidatorDelegations(ctx, app)
```

app/upgrades_test.go
```diff
@@ -421,3 +429,210 @@ func (s *UpgradeTestSuite) TestRemoveP8eMemorializeContractFee() {
 		})
 	}
 }
+
+func (s *UpgradeTestSuite) TestRemoveInactiveValidatorDelegations() {
```
SpicyLemon, line 433: The `TestRust` and `TestRustRc1` should each have an expected line added to them: `"INF removing all delegations from validators that have been inactive (unbonded) for {whatever} days".

app/upgrades.go
```diff
@@ -230,6 +234,36 @@ func removeP8eMemorializeContractFee(ctx sdk.Context, app *App) {
...
 
+// removeInactiveValidatorDelegations unbonds all delegations from inactive validators, triggering their removal from the validator set.
+func removeInactiveValidatorDelegations(ctx sdk.Context, app *App) {
+	unbondingTimeParam := app.StakingKeeper.GetParams(ctx).UnbondingTime
+	ctx.Logger().Info(fmt.Sprintf("removing all delegations from validators that have been inactive (unbonded) for %d days", int64(unbondingTimeParam.Hours()/24)))
+	removalCount := 0
+	validators := app.StakingKeeper.GetAllValidators(ctx)
+	for _, validator := range validators {
+		if validator.IsUnbonded() {
+			inActiveDuration := ctx.BlockTime().Sub(validator.UnbondingTime)
```
SpicyLemon, line 245:
```suggestion
			inactiveDuration := ctx.BlockTime().Sub(validator.UnbondingTime)
```

iramiller (dismissed):

SpicyLemon (approved):
Approved. But since there's conflicts, I'll check back again later too.

iramiller (approved):

PR #1576, opened 2023-06-08
Disallow some messages and mark v1.15.2.

SpicyLemon:
## Description

Disallow some messages due to barberry.
Mark `v1.15.2` in the changelog.

This PR is only applied to `release/v1.15.x` as it won't be needed in `main`. `main` will get an updated SDK version with a better fix.

---

write a little note why.

### DIFF
--- a/internal/antewrapper/msg_blacklist.go
+++ b/internal/antewrapper/msg_blacklist.go
@@ -1,0 +1,40 @@
+package antewrapper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
+)
+
+// MsgTypeBlacklistContextDecorator is (hopefully) a temporary hard-coded antehandler that disallows certain messages.
+// Once the circuit breaker module is added to provenance, this should be removed.
+type MsgTypeBlacklistContextDecorator struct {
+	Blacklist []string
+}
+
+func NewMsgTypeBlacklistContextDecorator() MsgTypeBlacklistContextDecorator {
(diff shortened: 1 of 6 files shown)


internal/antewrapper/msg_blacklist.go
```diff
@@ -0,0 +1,40 @@
...
+// Once the circuit breaker module is added to provenance, this should be removed.
+type MsgTypeBlacklistContextDecorator struct {
+	Blacklist []string
+}
+
+func NewMsgTypeBlacklistContextDecorator() MsgTypeBlacklistContextDecorator {
+	return MsgTypeBlacklistContextDecorator{
+		Blacklist: []string{
+			// Disallow vesting account creation due to barberry: https://forum.cosmos.network/t/cosmos-sdk-security-advisory-barberry/10825
+			// Once that fix is in the SDK that we pull in, these can be removed.
```
nullpointer0x00, line 19: We should make issues for comments like these in the future and maybe put the issue link in the comment.

SpicyLemon: If this were going to `main`, I would make such an issue. But since it'll (hopefully) only be in `v1.15.2`, we won't actually have to remove it. But I put the comment there in the off chance it somehow makes its way back to `main`.

nullpointer0x00: Ok, makes sense.  Thought we would merge the deny list ante handler since it seems good to have around.

iramiller: we should keep this out of main in general because the circuit breaker stuff is coming and it does the same thing but better.

iramiller (approved):
a good interim solution until the circuit breaker is available

nullpointer0x00 (approved):

Taztingo (approved):
LGTM

PR #1594, opened 2023-06-15
Create pre-upgrade command. Fix minimum-gas-prices repacking unset.

SpicyLemon:
## Description

This PR:
1. Creates the `provenanced pre-upgrade` command as outlined by cosmovisor documentation. It will update the config files to the new formats and set `consensus.timeout_commit` to 5 seconds.
2. Adds an error message to the log if the `consensus.timeout_commit` is too low.
3. Fixes an issue where the `minimum-gas-prices` value was being set to "" when either rewriting a packed config or calling `provenanced config pack` on an already packed config.
4. Fixes the `provenanced debug pubkey-raw` command that was trying to add a flag with a `-t` shorthand that conflicted with the `--testnet`/`-t` persistent flag on the root command.

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
+TODO[1594]: Move these to correct sections once v1.16.0-rc1 is marked.
+* Add pre-upgrade command that updates config files to newest format and sets `consensus.timeout_commit` to `5s` [PR 1594](https://github.com/provenance-io/provenance/pull/1594)
+* Fix `minimum-gas-prices` from sometimes getting unset in the configs [PR 1594](https://github.com/provenance-io/provenance/pull/1594)
+
 ### Features
 
 * Add support to add/remove required attributes for a restricted marker. [#1512](https://github.com/provenance-io/provenance/issues/1512)
--- a/cmd/provenanced/cmd/pre_upgrade.go
+++ b/cmd/provenanced/cmd/pre_upgrade.go
@@ -81,12 +81,14 @@ func UpdateConfig(cmd *cobra.Command) error {
 		return err
 	}
 
-	// Update the timeout commit if it's too low.
-	timeoutCommit := config.DefaultConsensusTimeoutCommit
-	if tmCfg.Consensus.TimeoutCommit < timeoutCommit/2 {
-		cmd.Printf("Updating consensus.timeout_commit config value to %q (from %q)\n",
-			timeoutCommit, tmCfg.Consensus.TimeoutCommit)
-		tmCfg.Consensus.TimeoutCommit = timeoutCommit
+	if clientCfg.ChainID == "pio-mainnet-1" {
+		// Update the timeout commit if it's too low.
+		timeoutCommit := config.DefaultConsensusTimeoutCommit
+		if tmCfg.Consensus.TimeoutCommit < timeoutCommit/2 {
+			cmd.Printf("Updating consensus.timeout_commit config value to %q (from %q)\n",
+				timeoutCommit, tmCfg.Consensus.TimeoutCommit)
+			tmCfg.Consensus.TimeoutCommit = timeoutCommit
+		}
 	}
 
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
+TODO[1594]: Move these to correct sections once v1.16.0-rc1 is marked.
```
SpicyLemon, line 40: Unfinished TODO

cmd/provenanced/cmd/root.go
```diff
@@ -242,6 +238,11 @@ func txCommand() *cobra.Command {
 func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts servertypes.AppOptions) servertypes.Application {
 	var cache sdk.MultiStorePersistentCache
 
+	timeoutCommit := cast.ToDuration(appOpts.Get("consensus.timeout_commit"))
+	if timeoutCommit < 4*time.Second {
```
Taztingo, line 242: Should this be `5*time.Second`? It looks like 4 seconds can get around the error.

iramiller:
General comments re: 5s values and floors...

While 5 is what the SDK has used ... Tendermint recommends 1s ... and we have had some success with this lower value.  I wouldn't want to enforce something higher unless we have to.  There is a strong possibility of networks running on a private intranet which might want to leverage the lower value to improve the network performance for low latency workloads.

With this in mind it seems we should take a middle ground here and boost the value to 5 seconds for our public main net ... but preserve the ability to run the lower values (perhaps with a warning (edit: info? given the SDK has removed warn from the logger).

cmd/provenanced/cmd/root.go
SpicyLemon: Our main goal is to alert folks that are using the (old) default of `1s`. We know a value of `1s` isn't good. And we're pretty sure a value of `5s` is the way to go. But we don't really know about the values in-between. I originally chose `4s` here so that the error didn't show up for something they've chosen that is less than `5s`, but probably still okay, e.g. `4300ms`.

I've got a change coming, though, that only shows this error if it's less than half the default value, and only if running on mainnet.

cmd/provenanced/cmd/pre_upgrade.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+	clientCfg, err := config.ExtractClientConfig(cmd)
+	if err != nil {
+		return err
+	}
+
+	// Set any config values that we want to forcefully update.
+
+	timeoutCommit := 5 * time.Second
```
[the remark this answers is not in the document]
SpicyLemon: I've got a change incoming that does extract this to a global `var DefaultTimeoutCommit = 5 * time.Second`. I also made it only update the value if it's configured to less than half the default.

SpicyLemon:
> General comments re: 5s values and floors...
> 
> While 5 is what the SDK has used ... Tendermint recommends 1s ... and we have had some success with this lower value. I wouldn't want to enforce something higher unless we have to. There is a strong possibility of networks running on a private intranet which might want to leverage the lower value to improve the network performance for low latency workloads.
> 
> With this in mind it seems we should take a middle ground here and boost the value to 5 seconds for our public main net ... but preserve the ability to run the lower values (perhaps with a warning (edit: info? given the SDK has removed warn from the logger).

I tweaked the PR a bit. It now only shows the error/warning if it's a mainnet node. It also only shows it if it's less than 2500ms. Also, during pre-upgrade, it'll only get updated under the same circumstances.

I put the log message at the `Error` level since there's no `Warn` level and the SDK seems to use `Error` for lots of things that I'd consider warnings. Also, that way, it's more likely to be seen.

During `provenanced init`, you can now provide the desired timeout commit. If it's not provided, and it's for a mainnet or testnet node, it'll default to 5s, otherwise 1s.

Basically, all our local chains should still use 1s, but a fresh mainnet or testnet node will get 5s. And only `pio-mainnet-1` nodes will see the warning or get their value updated from the `pre-upgrade` command.

nullpointer0x00 (approved):

Taztingo (approved):

Issue #1518, opened 2023-04-27
Allow marker admin to prevent transfer of restricted coin with deny list on send


leeduan:
## Summary

Allow a marker admin to sanction or add specific addresses to a deny list to prevent that address from transferring that coin to another address.

## Problem Definition

With the introduction of `RequiredAttributes` in the marker module for restricted coins allowing anyone to use the bank module send functionality to any to address with those respective attributes, we can't prevent a from address that the marker admin(s) wants to sanction from transferring that coin to another address with those attributes.

Some example use cases:
* a fiat or stablecoin provider may want to prevent a blockchain address from being able to transfer coin that was minted in relation to an ACH deposit that was later claimed fraudulent.
* a court order notifies a TA to reissue a security token to a new address because old coin is held in an account that is no longer accessible due to lost keys

## Proposal

Allow some kind of marker-specific sanction or deny list so that we prevent bank send on restricted marker with required attributes on from addresses.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

nullpointer0x00:
Yes, I was wondering about this case as well.  For example, if an address is added to the deny list.  We will probably have to look at the account and see if they have any of the restricted markers and do something house keeping if they do.  If they don't, we can just add the address.  We could do this issue without that to begin with.  Just have the ability to add the deny list on marker creation.  That way the denied addresses would not have any markers yet.  Then do all the logic of denying a send.

The next issue would be to make an Update deny list tx.  That could would do all that work.  Since, the address added could have those markers.

iramiller:
After some discussion with Figure about market place restrictions for bilateral settlement we should consider a generic restriction module function for this implementation.  

In the examples discussed in that meeting it is desired that the tokens being exchanged be escrowed in the user’s wallet with a hold placed on committed tokens. This is preferable to a situation with tokens moved into a contract for escrow. 

Along this same line the ability to stop transfer of tokens out of a wallet neatly solves the vesting account setup where there are multiple vesting claims. With a proper hold function that implemented lockups as a service for other modules the marker module could issue a hold, vesting systems could do the same… even a future exchange module could issue and cancel holds for in place escrow of funds.

SpicyLemon:
[...]

PR #1604, opened 2023-06-22
Allow marker admin to prevent transfer of restricted coin with deny list on send

nullpointer0x00:
## Description

closes: #1518

---

write a little note why.


proto/provenance/marker/v1/tx.proto
```diff
@@ -291,4 +293,19 @@ message MsgSetAccountDataRequest {
...
\ No newline at end of file
+message MsgSetAccountDataResponse {}
+
+// MsgUpdateSendDenyListRequest defines a msg to add/remove addresses to send deny list for a resticted marker
+// signer must have admin authority
+message MsgUpdateSendDenyListRequest {
+  option (gogoproto.equal)      = true;
+  option (cosmos.msg.v1.signer) = "authority";
+
+  string          denom                      = 1;
```
Taztingo, line 304: Very minor, but comments before each field would be helpful.

x/marker/keeper/keeper_test.go
```diff
@@ -1415,3 +1415,119 @@ func TestRemoveIsSendEnabledEntries(t *testing.T) {
...
+			updateMsgRequest: *types.NewMsgUpdateSendDenyListRequest("blah", authUser, []string{}, []string{}),
+			expectedError:    "marker not found for blah: marker blah not found for address: cosmos1psw3a97ywtr595qa4295lw07cz9665hynnfpee",
+		},
+		{
+			name:             "should fail, not a restricted marker",
+			updateMsgRequest: *types.NewMsgUpdateSendDenyListRequest(notRestrictedMarker.Denom, authUser, []string{}, []string{}),
+			expectedError:    "marker not-restricted-marker is not a restricted marker",
+		},
+		{
+			name:             "should fail, not a restricted marker",
```
Taztingo, line 1470: I think you meant to name this test differently. "should fail, no admin access"

x/marker/keeper/msg_server.go
```diff
@@ -764,3 +764,63 @@ func (k msgServer) SetAccountData(goCtx context.Context, msg *types.MsgSetAccoun
...
+	if marker.GetMarkerType() != types.MarkerType_RestrictedCoin {
+		return nil, fmt.Errorf("marker %s is not a restricted marker", msg.Denom)
+	}
+
+	if msg.Authority == k.GetAuthority() {
+		if !marker.HasGovernanceEnabled() {
+			return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
+		}
+	} else {
+		if !marker.HasAccess(msg.Authority, types.Access_Admin) {
```
Taztingo, line 786: Should this be outside this block? It looks like it is only checking admin access of the authority when it is done by a governance proposal.

x/marker/spec/03_messages.md
```diff
@@ -358,6 +358,22 @@ This service message is expected to fail if:
 - Attributes cannot be normalized
 - Marker denom cannot be found or is not a restricted marker
 
+## Msg/UpdateSendDenyListRequest
+
+UpdateSendDenyList allows signers that have admin authority or via gov proposal to add and remove addresses to the deny send list for a restricted marker.
+
++++ https://github.com/provenance-io/provenance/blob/373d0ee8faeaa9e3b70d658e6069ab1781e6ce5e/proto/provenance/marker/v1/tx.proto #L295-L308
```
Taztingo, line 365: I think you have a space in the link where the line numbers are.

x/marker/spec/03_messages.md
```diff
@@ -358,6 +358,22 @@ This service message is expected to fail if:
...
+
++++ https://github.com/provenance-io/provenance/blob/373d0ee8faeaa9e3b70d658e6069ab1781e6ce5e/proto/provenance/marker/v1/tx.proto #L310-L311
```
Taztingo, line 367: Same as above.

x/marker/spec/03_messages.md
```diff
@@ -358,6 +358,22 @@ This service message is expected to fail if:
...
+
+This service message is expected to fail if:
+
+- Remove list has an address that does not exist in current deny list
```
Taztingo, line 371: You can probably add this as well...

`- Both add and remove lists are empty`

x/marker/spec/03_messages.md
```diff
@@ -358,6 +358,22 @@ This service message is expected to fail if:
...
+- Add list has an attribute that already exist in current deny list
+- Invalid address format in add/remove lists
+- Marker denom cannot be found or is not a restricted marker
+- Signer is does not have admin authority or is not from gov proposal
```
Taztingo, line 375: I think you meant to say `Signer does not have admin authority or is not from gov proposal`

x/marker/types/msg_test.go
```diff
@@ -689,3 +689,94 @@ func TestMsgSetAccountDataRequestGetSigners(t *testing.T) {
...
+			msg:           MsgUpdateSendDenyListRequest{Denom: denom, RemoveDeniedAddresses: []string{"invalid-address"}, AddDeniedAddresses: []string{}, Authority: addr},
+			expectedError: "decoding bech32 failed: invalid separator index -1",
+		},
+		{
+			name:          "should succeed, invalid add address",
+			msg:           MsgUpdateSendDenyListRequest{Denom: denom, RemoveDeniedAddresses: []string{removeAddr}, AddDeniedAddresses: []string{"invalid-addrs"}, Authority: addr},
+			expectedError: "decoding bech32 failed: invalid separator index -1",
+		},
+		{
+			name:          "should succeed, duplicate entries in list",
```
Taztingo, line 734: I think this is supposed to be "should fail".

x/marker/keeper/msg_server.go
```diff
@@ -764,3 +764,63 @@ func (k msgServer) SetAccountData(goCtx context.Context, msg *types.MsgSetAccoun
...
+		if err != nil {
+			return nil, err
+		}
+		if k.IsSendDeny(ctx, markerAddr, denyAddr) {
+			return nil, fmt.Errorf("%s is already on deny list cannot add address", addr)
+		}
+		k.AddSendDeny(ctx, markerAddr, denyAddr)
+	}
+
+	ctx.EventManager().EmitEvent(
```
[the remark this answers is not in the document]
nullpointer0x00: I agree it looks pretty generic, but this is correct.  This is a pattern we should be following in the msg server methods.

Taztingo (approved):

CHANGELOG.md
```diff
@@ -46,8 +50,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 * Fix ibcnet relayer creating multiple connections on restart [#1620](https://github.com/provenance-io/provenance/issues/1620).
 
----
-
```
SpicyLemon, line 50: This should still be here.

x/marker/types/key.go
```diff
@@ -32,6 +32,9 @@ const (
 var (
 	// MarkerStoreKeyPrefix prefix for marker-address reference (improves iterator performance over auth accounts)
 	MarkerStoreKeyPrefix = []byte{0x02}
+
+	// DenySendKeyPrefix prefix for adding addresses that are denied send functionality on restricted markers
+	DenySendKeyPrefix = []byte{0x02}
```
SpicyLemon, line 37: Need to have different values here (both are `0x02`).

x/marker/types/msg.go
```diff
@@ -651,3 +652,44 @@ func (msg MsgSetAccountDataRequest) GetSigners() []sdk.AccAddress {
...
+	combined := []string{}
+	combined = append(combined, msg.AddDeniedAddresses...)
+	combined = append(combined, msg.RemoveDeniedAddresses...)
+	seen := make(map[string]bool)
+	for _, addr := range combined {
+		if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+			return err
+		}
+		if seen[addr] {
+			return fmt.Errorf("denied address lists contain duplicate entries")
```
SpicyLemon, line 683:
```suggestion
			return fmt.Errorf("denied address lists contain duplicate entries: %q", addr)
```

SpicyLemon: Never mind on this. Really, it'd be handy to list all offending addresses, but that could easily become unwieldy. It's probably okay to just make the simple statement and let the user identify the offending ones on their own.

x/marker/keeper/msg_server.go
```diff
@@ -752,3 +752,63 @@ func (k msgServer) SetAccountData(goCtx context.Context, msg *types.MsgSetAccoun
...
+	if marker.GetMarkerType() != types.MarkerType_RestrictedCoin {
+		return nil, fmt.Errorf("marker %s is not a restricted marker", msg.Denom)
+	}
+
+	if msg.Authority == k.GetAuthority() {
+		if !marker.HasGovernanceEnabled() {
+			return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
+		}
+	} else {
+		if !marker.HasAccess(msg.Authority, types.Access_Transfer) {
```
SpicyLemon, line 774: I feel like `Transfer` access isn't the correct thing to use here. We probably want a separation of concerns between "people who can transfer the coin" and "people that control who can transfer the coin."

The `Admin` access might actually be correct here. It was originally for "People that can add/remove access", which is fairly in line with what's being done here.

Alternatively, we might need a new access type, and that name might be hard to iron out. My initial though is `manager`, but I worry that that will get confused with the marker field with that same name. I feel like that's what we're still trying to convey, though. There's a few endpoints now that might make more sense using such a permission.

Thoughts?

iramiller: `transfer` literally is intended as "transfer authority" ... so they can delegate/pre-approve via kyc, etc if they wish but ultimately they are responsible for controlling the transfer of the token asset.  "People that are approved to transfer" are only approved if the `transfer` holding authority has approved it.

x/marker/types/key.go
nullpointer0x00: 😬 Thought I changed that.  Good catch, that could have been an issue.

iramiller: curious why a test didn't fail with this ... definitely a good catch.

nullpointer0x00: Should not have written the test like 
```	
assert.Equal(t, DenySendKeyPrefix[0], denyKey[0], "should have correct prefix for deny send key")
` ``

Changed it to (which I was about to do in the first place...ugh): 
```
assert.Equal(t, uint8(3), denyKey[0], "should have correct prefix for deny send key")
```

File: x/marker/keeper/msg_server.go (hunk 13 above)
Comment (nullpointer0x00) [reply x/marker/keeper/msg_server.go:774]: It makes sense that it would be `transfer` instead of `admin` or something else.  Transfer authority also controls the force transfer.  Therefore, a transfer authority account could put an address on a send deny list then force transfer those markers to another address.   This would fill a requirement of recovery of a coin if the holder has lost access to their private keys.

Comment (iramiller) [reply x/marker/keeper/msg_server.go:774]: I believe they would want to use `authz` to authorize `MsgTransferRequest` which isn't going to let someone actually perform any of the operations that explicitly check for `transfer` grants held by and address...  I am not aware of anyone actually giving someone the `transfer` grant directly just so they can move the tokens.

File: x/marker/types/msg_test.go [hunk 14]
```diff
@@ -689,3 +689,94 @@ func TestMsgSetAccountDataRequestGetSigners(t *testing.T) {
...
+func TestMsgUpdateSendDenyListRequest(t *testing.T) {
+	addr := sdk.AccAddress("addr________________").String()
+	denom := "somedenom"
+	addAddr := sdk.AccAddress("addAddr________________").String()
+	removeAddr := sdk.AccAddress("removeAddr________________").String()
+
+	tests := []struct {
+		name          string
+		msg           MsgUpdateSendDenyListRequest
+		expectedError string
```
SpicyLemon, line 702: Nit: I usually try to keep these field names short so that, when the values are aligned, they're all further left.

When testing something that only returns an error (or any other single value), I usually just call the field `exp`. Some times I'll call it `expErr` though.

No need to change anything. I just wanted to share. Feel free to just resolve this comment.

x/marker/types/msg_test.go
```diff
@@ -689,3 +689,94 @@ func TestMsgSetAccountDataRequestGetSigners(t *testing.T) {
 		require.PanicsWithError(t, "decoding bech32 failed: invalid separator index -1", testFunc, "GetSigners")
 	})
 }
+
+func TestMsgUpdateSendDenyListRequest(t *testing.T) {
```
SpicyLemon, line 693: Nit 2: This function should be called `TestMsgUpdateSendDenyListRequestValidateBasic` since it's only testing the `ValidateBasic` method.

x/marker/types/msg_test.go
```diff
@@ -689,3 +689,94 @@ func TestMsgSetAccountDataRequestGetSigners(t *testing.T) {
...
+		{
+			name:          "should fail, invalid remove address",
+			msg:           MsgUpdateSendDenyListRequest{Denom: denom, RemoveDeniedAddresses: []string{"invalid-address"}, AddDeniedAddresses: []string{}, Authority: addr},
+			expectedError: "decoding bech32 failed: invalid separator index -1",
+		},
+		{
+			name:          "should succeed, invalid add address",
+			msg:           MsgUpdateSendDenyListRequest{Denom: denom, RemoveDeniedAddresses: []string{removeAddr}, AddDeniedAddresses: []string{"invalid-addrs"}, Authority: addr},
+			expectedError: "decoding bech32 failed: invalid separator index -1",
+		},
```
SpicyLemon, line 732: "should succeed" is not correct.

I'm really not a fan of including "should succeed" or "should fail" in the test case names. It's too easy to accidentally get it wrong, or forget to correct it with a change that flips it. Plus, if the test case fails, the expected "should fail" or "should succeed" concept is conveyed by the different assertion failure messages. When looking at the test case definition, whether its expected to fail is apparent with the presence or absence of an expected error.

x/marker/keeper/keeper_test.go
```diff
@@ -1415,3 +1415,114 @@ func TestRemoveIsSendEnabledEntries(t *testing.T) {
 	assert.True(t, app.BankKeeper.IsSendEnabledDenom(ctx, "testcoin"), "should not exist in table therefore default to true")
 
 }
+
+func TestMsgUpdateSendDenyListRequest(t *testing.T) {
```
SpicyLemon, line 1419:
```suggestion
func TestUpdateSendDenyList(t *testing.T) {
```

x/marker/keeper/keeper_test.go
```diff
@@ -1415,3 +1415,114 @@ func TestRemoveIsSendEnabledEntries(t *testing.T) {
...
+
+	denyAddrToAdd := testUserAddress("denyAddrToAdd")
+
+	denyAddrToAddGov := testUserAddress("denyAddrToAddGov")
+
+	testCases := []struct {
+		name             string
+		updateMsgRequest types.MsgUpdateSendDenyListRequest
+		expectedReqAttr  []string
+		expectedError    string
```
SpicyLemon, line 1457: Here too with the shorter names thing. Also, `expectedReqAttr` isn't used and should be removed.

But you could call the fields just `name`, `msg`, and `expErr`, without losing any readability.
You could also use a `*types.MsgUpdateSendDenyListRequest` since that's what the `UpdateSendDenyList` function accepts, and you're always using the constructor, which returns the reference which you then have to dereference for the test case and re-reference for the call in the test.

Also, I'm really not a fan of using the constructors in unit tests, specially when they're so simple. My problem with them is that, when we add a field to the message, we then have to add it to the constructor, and then we have to update all these test cases to include a value for the new field. In 99% of the cases, we just want the zero value, or one specific value, and it'd be better if we didn't even have to make a change to each test case in order for the tests to continue to pass. Plus, creating your own private constructor in this test (e.g. `newMsg := func(...)`) let's you use a smaller name, making it easier to read the test case definitions; it helps with the signal (msg values) to noise (the name of the constructor) ratio.

x/marker/types/msg_test.go
nullpointer0x00: I personally like the `expErr` instead of `exp` if we are going to shorten them.  I keep reading `exp` as `expression` in my head.  Since, that is what was used as at a previous job...I will get over that though.

x/marker/types/key.go
SpicyLemon: I thought this test was fine. Either way works though.

Since the key formats are different, it wouldn't have been overwriting existing data. However, iterating over all markers would have gone wrong when it hit a deny list entry and failed to marshal it.

SpicyLemon (approved):

Taztingo (approved):

Issue #1443, opened 2023-03-24
Smaller metadata query results


SpicyLemon:
## Summary

Make the `x/metadata` queries only return the entry/entries requested, without extra info.

## Problem Definition

Currently, the metadata queries put each result in a wrapper that also has address info. Those address info chunks add quite a bit of data to the results. That extra data becomes problematic with smart contracts. Including the request in the response is similarly problematic.

## Proposal

1. Only optionally include the request in the responses and default to not including them.
2. Only optionally include the id info in the responses and default to not including them. When not including id info, the entries will still be in a wrapper, but the id info fields will be nil. Handling of not-found entries should remain the same (i.e. add a wrapper entry with id info of the thing not found).
3. Create a new `GetByAddr` endpoint that takes in one or more metadata addresses and returns all that are requested. This should NOT be able to include the request or id info, just the data requested. The response will need to have a list of each of the 6 metadata types as well as a list of "not found" addresses. Either the entry goes in it's appropriate list or else the address is added to "not found."

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1614, opened 2023-07-04
GetByAddr and smaller metadata queries.

SpicyLemon:
## Description

closes: #1443

This PR:
1. Adds an `include_request` flag to appropriate metadata queries. The request is only included in the response if set to `true`. This info was previously always included, so this is a breaking change.
2. Add an `exclude_id_info` flag to appropriate metadata queries. If set to true, the `*_id_info` fields in the response will not be populated.
3. Create a `GetByAddr` endpoint that takes in one or more metadata addresses and returns all the requested items. The `query metadata get` command has been overhauled to utilize this query instead of each individual type query.

---

write a little note why.


Taztingo (approved):

Taztingo (dismissed):

iramiller:
seems like a pretty elegant solution ... I do wonder if this should be "exclude_extra_info" since the default for proto is null/zero so in this case the default behavior would be the existing one and the opt-in would be the streamlined results?  This would give smart contract authors the option of streamlining the results if they are able but would keep existing workflows working without changes...

SpicyLemon:
> seems like a pretty elegant solution ... I do wonder if this should be "exclude_extra_info" since the default for proto is null/zero so in this case the default behavior would be the existing one and the opt-in would be the streamlined results? This would give smart contract authors the option of streamlining the results if they are able but would keep existing workflows working without changes...

I'm figured we don't have any smart contracts dealing with metadata yet since most of that stuff is missing from the [x/metadata/wasm](https://github.com/provenance-io/provenance/tree/main/x/metadata/wasm) directory. But this would break anything else that runs the queries and expects the `_id_info` to be there, which might not be as insignificant as I originally thought.

I primarily wanted to, by default, shrink the size of the responses. I also wanted to stick with the `include_` pattern since there are several other fields in the queries that use `include_`, but none yet for `exclude_`.

I do have an entry under `Client Breaking` in the changelog about it. I'm not quite sure how we'd go about getting a sense of how breaking that is, though. Anything using the `_id_info` fields would break immediately at upgrade. If there's enough stuff already using them, though, it's probably best to switch to `exclude_`.

The more I think about it, though, the more I think you're right. I'll switch it to `exclude_id_info`.

iramiller (approved):

Taztingo (approved):

nullpointer0x00 (approved):

Issue #1620, opened 2023-07-05
Ibcnet relayer creates a new channel every time the relayer is restarted


Taztingo:
## Summary of Bug

Every time that the relayer is restarted with `docker stop relayer` and `docker start relayer` a new channel is created. The same channel should be used.

## Version

1.16.0

## Steps to Reproduce

`docker stop relayer`
`docker start relayer`
`provenanced -t --home ./build/ibc0-0 q ibc channel channels`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1621, opened 2023-07-05
Taztingo/1620 fix relayer restart

Taztingo:
## Description

Adds a small fix to ibcnet and prevents the relayer from spawning multiple connections. This is very helpful when debugging locally.

closes: #1620

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
-* nothing
+### Bug Fixes
+* Fix ibcnet relayer creating multiple connections on restart [#1620](https://github.com/provenance-io/provenance/issues/1620).
 
 ---
 
--- a/networks/ibc/blockchain-relayer/entrypoint.sh
+++ b/networks/ibc/blockchain-relayer/entrypoint.sh
@@ -7,8 +7,13 @@
 export RELAYER_HOME="/relayer"
 export RELAY_PATH="local_local2"
 
-# Setup the connection, client, and channel
-rly tx link "$RELAY_PATH" --home "$RELAYER_HOME"
+# Setup the connection, client, and channel on first run
+PATH_COUNT=$(rly --home "$RELAYER_HOME" q connections local | wc -l)
+if [[ $PATH_COUNT -eq 0 ]];then
+    echo "Initializing the relayer"
+    rly tx link "$RELAY_PATH" --home "$RELAYER_HOME"
+fi
 
 # Start the relayer
+echo "Starting the relayer"
 rly start "$RELAY_PATH" -p events --home "$RELAYER_HOME"


nullpointer0x00 (dismissed):

CHANGELOG.md
```diff
@@ -37,7 +37,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## Unreleased
 
-* nothing
+### Bug Fixes
+* Fix ibcnet relayer creating multiple connections on restart [#1620](https://github.com/provenance-io/provenance/issues/1620).
```
SpicyLemon, line 41:
```suggestion
### Bug Fixes

* Fix ibcnet relayer creating multiple connections on restart [#1620](https://github.com/provenance-io/provenance/issues/1620).
```

iramiller (approved):

networks/ibc/blockchain-relayer/entrypoint.sh
```diff
@@ -7,8 +7,13 @@
 export RELAYER_HOME="/relayer"
 export RELAY_PATH="local_local2"
 
-# Setup the connection, client, and channel
-rly tx link "$RELAY_PATH" --home "$RELAYER_HOME"
+# Setup the connection, client, and channel on first run
+PATH_COUNT=$(rly --home "$RELAYER_HOME" q connections local | wc -l)
+if [[ $PATH_COUNT -eq 0 ]];then
```
SpicyLemon, line 12: It's best to always double quote variables in tests, i.e. `[ ... ]` or `[[ ... ]]`. This instance is mostly safe because `wc -l` will always return a single thing (annoyingly padded with spaces, though), but always quoting them is a good habit to get into.

I'm just passing along knowledge here. There's no need to change it unless you're making other changes too.

It matters when the variable is empty, or only whitespace. With double quotes, i.e. `"$PATH_COUNT"`, it's always interpreted as a single argument. Without the quotes, it's expanded by the shell. If it's an empty string, the shell will think there' isn't an argument on the left of the `-eq` and will give an error. With the quotes, it'll be providing an empty string to the `-eq` test, which treats empty strings as `0`.

If doing a test like `==`, you will also run into problems when the variable has two or more words. Without the quotes, the shell will expand the variable and treat each word in it as an argument. Since it's only expecting one argument, though, it'll give an error. If it's quoted, it'll get interpreted as a single argument.

Also, thankfully, the arithmetic test operators (e.g. `-eq` or `-lt`) will trim leading and trailing whitespace. So even though `$PATH_COUNT` will look something like `       1`, it'll ignore the whitespace in that arg and just compare `1` to the right-hand-side. Without the quotes, the shell will strip the whitespace before doing the comparison, which has the same outcome in this case, but isn't as safe in general.

SpicyLemon (approved):

SpicyLemon (dismissed):

SpicyLemon (approved):

iramiller (approved):

Issue #1328, opened 2023-01-29
Migrate msgfees gov props to v1.


llama-del-rey:

SpicyLemon:
Work mostly done on [llama/migrate-add-msg-fee-proposal](https://github.com/provenance-io/provenance/compare/llama/migrate-add-msg-fee-proposal). I'm not sure what else needs doing in that branch, but it shouldn't be much effort to pick it up later and finish it.

PR #1627, opened 2023-07-11
Migrate msgfees gov props v1

nullpointer0x00:
## Description

closes: #1328

---

write a little note why.


Taztingo (dismissed):

x/msgfees/types/msgs_test.go
```diff
@@ -77,3 +79,298 @@ func TestMsgAssessCustomMsgFeeValidateBasic(t *testing.T) {
...
+	cases := []struct {
+		name     string
+		msg      *MsgAddMsgFeeProposalRequest
+		errorMsg string
+	}{
+		{
+			"Empty type error",
+			NewMsgAddMsgFeeProposalRequest("", sdk.NewCoin("hotdog", sdk.NewInt(10)), "", "", authority),
+			"msg type is empty",
+		},
```
SpicyLemon, line 95: **Always** use named parameters when defining test cases. **Always**.
Without them, you can't make any changes to the test case struct without having to change every single test case, even when the test case isn't actually changing. They also make the tests easier to understand lower in the list where you don't have the struct definition on the screen anymore.

x/msgfees/types/msgs_test.go
```diff
@@ -77,3 +79,298 @@ func TestMsgAssessCustomMsgFeeValidateBasic(t *testing.T) {
...
+		{
+			"Invalid fee amounts",
+			NewMsgAddMsgFeeProposalRequest("msgType", sdk.NewCoin("hotdog", sdk.NewInt(0)), "", "", authority),
```
SpicyLemon, line 98: Relatedly, these tests are significantly easier to maintain if you don't use the `New...` constructor to make the thing you're testing. When a field is added to a message, and therefore the constructor, you then have to go back through all the tests and update the constructor call. But usually, when adding a field like that, all of the existing unit tests are still valid and wouldn't need to be changed if they didn't use the constructor.

x/msgfees/types/msgs_test.go
nullpointer0x00: Dang missed this.  I usually do the named parameters.  Don't know why I missed them here.

x/msgfees/types/msgs_test.go
nullpointer0x00: I will change them.  However, I do like having the constructor for that reason of breaking test or anywhere it is used.  It forces me to review the tests and update accordingly or remove ones that might be invalid after the updated constructor.  I was usually just using the fields when I needed to create the object in a state the constructor didn't allow.  I will start using the fields now going forward, just wanted to clarify my reasoning behind it.

SpicyLemon: If the test itself doesn't break, we shouldn't have to update the code for it.

SpicyLemon (approved):

iramiller (approved):

Issue #1607, opened 2023-06-23
Coin escrow without transferring out of account


leeduan:
## Summary

Provide a way for a signer to lock or hold coin in their account, while still granting another
account access to funds when conditions meet for exchange of coins, without having to transfer
coins to a smart contract or other "escrow" address.

## Problem Definition

For more complex bilateral settlement of coins between two (or more accounts)
to occur when some condition has been met, coin must be first transferred to some escrow
account, usually a smart contract address, to guarantee availability and prevent possibility
of double spend or account holder changing their minds via removal of grant authz.

This can be considered suboptimal for the following reasons:
1. Coin leaves account that still really "owns" the asset which makes it more difficult to view or report on true ownership
2. Potentially difficult for compliance teams to explain to regulars who now "owns" the asset while coin is in escrow
3. More difficult to track true movement of asset between parties as coins are potentially aggregated in a single smart contract address bucket
4. Harder to argue this is true bilateral settlement when coins move through an address "owned" by a third party and exchange is not directly between exchanging accounts.
5. Potentially extra gas fees associated with transference of assets

## Proposal

Allow a signer to move a coin balance to a "held" bucket to lock/escrow funds in their account and designate
another address the ability to move those funds how they see fit. 

The type of lock should be configurable:
1. Locked for perpetuity until designated address moves the funds (or cancels the lock)
2. Lock with some expiration time

Ideally, you would also be able to query for all held authorizations on an account that are still active.

You would also ideally be able to query for:
1. All coin balances on account (default behavior today)
2. All coin balances on account not held for transaction purposes

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

leeduan:
Some discussions around this exists in https://github.com/provenance-io/provenance/issues/1518

iramiller:
The separate discussions around this identified some additional aspects:

## The coins in an account would not move when holds are placed.  

This is an important feature for transfer authority tracking but especially for restricted tokens that require explicit approval to move.  If the tokens are not moved then the approval process does not need to be executed when the hold is placed.

## The account with the funds being held doesn't actually have permission to create or cancel holds (directly).

[...]

PR #1638, opened 2023-07-28
Create the x/hold module.

SpicyLemon:
## Description

closes: #1607 

This PR creates a `x/hold` module which is mostly just core functionality for managing holds on funds.

Funds with a hold on them can be looked up using the `GetHolds` query (for a single address) or `GetAllHolds` query for all addresses. The bank module's `SpendableBalances` query also reflects funds that have a hold on them.

---

write a little note why.


iramiller:
It is worth pointing out that the term "escrow" has a specific meaning of a temporary transfer _to a third party_ which is exactly what we are trying to avoid doing with the in-place hold.  We likely should not be using the name "escrow" for this feature. [see more at invest-pediatric](https://www.investopedia.com/terms/i/in-escrow.asp)

go.mod
```diff
@@ -172,7 +174,8 @@ require (
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
-replace github.com/cosmos/cosmos-sdk => github.com/provenance-io/cosmos-sdk v0.46.13-pio-1
+// replace github.com/cosmos/cosmos-sdk => github.com/provenance-io/cosmos-sdk v0.46.13-pio-1
+replace github.com/cosmos/cosmos-sdk => github.com/provenance-io/cosmos-sdk v0.46.13-pio-1.0.20230728170659-9d5e97ff8efd
```
SpicyLemon, line 178: I'll want an official version here before this gets merged.

SpicyLemon:
> It is worth pointing out that the term "escrow" has a specific meaning of a temporary transfer _to a third party_ which is exactly what we are trying to avoid doing with the in-place hold. We likely should not be using the name "escrow" for this feature. [see more at investopedia](https://www.investopedia.com/terms/i/in-escrow.asp)

Good call. I changed it to the `x/hold` module, and didn't call it "escrow" anywhere.

docs/proto-docs.md
```diff
@@ -1172,6 +1190,205 @@ Msg defines the attribute module Msg service.
...
+<a name="provenance.hold.v1.EventHoldAdded"></a>
+
+### EventHoldAdded
+EventHoldAdded is an event indicating that some funds were placed on hold in an account.
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| `address` | [string](#string) |  | address is the bech32 address string of the account with the funds. |
+| `amount` | [string](#string) |  | amount is a Coins string of the funds placed on hold. |
```
iramiller, line 1209: we should consider adding a free form "reason" string to this event

SpicyLemon: Good call.

I added it to the `EventHoldAdded`. I added a `reason` argument to the `AddHold` keeper function which is passed directly into the event.

I didn't put any restrictions on it though. It can be empty or ginormous.

Arguments for no limits:
1. I expect the value to always be code generated. As such, an error for it would be really annoying from a user's standpoint.
2. Along those same lines, I didn't really want to add a new failure point to `AddHold`.
3. The max would be completely arbitrary anyway. The reason for a max is to save space, but there isn't a specific number from a technical standpoint. My first thought was 100 charcters, but I'm sure there will end up being legitimate uses that might want more than that. So then I thought about 250 or 1000. But I realized that I was trying to come up with a number that would never actually cause an error. If I never wanted it to cause an error, the answer is to not have a max.
4. I decided to allow an empty string due to the same reasons as not having a max. I.e. "What's a good minimum length." A one character reason is only barely less worthless as an empty reason; it's supposed to be human readable.
5. It'd be easy to add a limit later if some problematic reasons are being used. At that time, we'll know more about what not to allow.

iramiller: I was thinking this is similar to the reason given on gas meter usage ... something that essentially is an event from the point of use and is setup by code ... I don't see any reason to really put much structure around it, empty or otherwise it would be up to the caller to put something that makes sense in there.

go.mod
SpicyLemon: Bumped to `v0.46.13-pio-2`.

x/hold/simulation/genesis.go
```diff
@@ -0,0 +1,126 @@
...
+
+	simState.GenState[hold.ModuleName] = simState.Cdc.MustMarshalJSON(holdGenState)
+	fmt.Printf("Selected %d randomly generated holds:\n%s\n", len(holdGenState.Holds), holdsString(holdGenState.Holds))
+
+	// If we put stuff in hold, add those funds to the bank balances.
+	if len(holdGenState.Holds) > 0 {
+		bankGenState := banktypes.GetGenesisStateFromAppState(simState.Cdc, simState.GenState)
+		UpdateBankGenStateForHolds(bankGenState, holdGenState)
+		simState.GenState[banktypes.ModuleName] = simState.Cdc.MustMarshalJSON(bankGenState)
+		fmt.Printf("Bank balances after update due to randomly generated holds:\n%s\n", balancesString(bankGenState.Balances))
```
SpicyLemon, line 124: This ends up being a pretty long string. I should probably delete it.

SpicyLemon: Removed.

iramiller (approved):

SpicyLemon:
Heh. Ran some analysis on the diff:
```
Repo Directory: provenance
 Source Branch: dwedul/1607-in-place-escrow
 Target Branch: main
==============================================
Line Changes -     Code:  +1694      -4
Line Changes -    Tests:  +6590     -16
Line Changes -    Proto:   +127      -0
Line Changes - Markdown:   +173      -0
Line Changes -     Auto:  +2906      -0
========================================
Line Changes -    Total: +11490     -20
```

x/hold/spec/02_state.md
```diff
@@ -0,0 +1,22 @@
...
+
+```
+0x00 | len(<address>) | <address> | <denom> -> <amount>
+```
+
+Where:
+
+* `0x00` is the type byte, and has a value of `0` for these records.
+* `len(<address>)` is a single byte containing the length of the `<address>` as an 8-bit byte in big-endian order.
+* `<address>` is the raw bytes of the address of the account that the funds are in.
```
Taztingo, line 17: It would be helpful to know how many bytes that each section of the key would need.

x/hold/spec/03_messages.md
```diff
@@ -0,0 +1,3 @@
+# Msg Service
```
Taztingo, line 1: Can we just remove this spec?

x/hold/spec/02_state.md
SpicyLemon: There's no set length for the address. That's why there's a length byte right before it.

x/hold/spec/03_messages.md
SpicyLemon: Yeah. I originally had it because I wanted it to be clear that there aren't any Tx endpoints as opposed to their documentation just missing. I poked around in the SDK, though, and they don't have any such mostly-empty pages.

Good call. Removed.

testutil/assertions/events.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	abci "github.com/tendermint/tendermint/abci/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// PrependToEach prepends the provided prefix to each of the provide lines.
+func PrependToEach(prefix string, lines []string) []string {
```
Taztingo, line 14: This is pretty generic. Might be better to put this in general.

x/hold/events_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			event := NewEventHoldReleased(tc.addr, tc.amount)
+			assert.Equal(t, tc.exp, event, "NewEventHoldReleased")
+		})
+	}
+}
+
+func TestTypedEventToEvent(t *testing.T) {
```
Taztingo, line 111: Why do we have to test the sdk.TypedEvenToEvent functionality?

x/hold/hold_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/provenance-io/provenance/testutil/assertions"
+)
+
+func TestAccountHold_Validate(t *testing.T) {
+	cz := func(coins string) sdk.Coins {
+		rv, err := sdk.ParseCoinsNormalized(coins)
+		require.NoError(t, err, "ParseCoinsNormalized(%q)", coins)
+		return rv
+	}
+	c := func(amount int64, denom string) sdk.Coin {
```
Taztingo, line 20: This looks the same as sdk.NewInt64Coin().

x/hold/hold_test.go
```diff
@@ -0,0 +1,70 @@
...
+		return sdk.Coin{Denom: denom, Amount: sdkmath.NewInt(amount)}
+	}
+	ah := func(addr string, amount sdk.Coins) AccountHold {
```
Taztingo, line 23: You could probably make a constructor for this to make it a bit more readable.

x/hold/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,314 @@
...
+			}
+			s.Require().NotPanics(testFunc, "GetHolds")
+			s.assertErrorContents(err, tc.expErr, "GetHolds error")
+			s.Assert().Equal(tc.expResp, response, "GetHolds response")
+		})
+	}
+}
+
+func (s *TestSuite) TestKeeper_GetAllHolds() {
+	accHold := func(addr sdk.AccAddress, amount string) *hold.AccountHold {
```
Taztingo, line 88: If you end up using the suggested constructor this can replaced as well.

x/hold/keeper/keeper.go
```diff
@@ -0,0 +1,289 @@
...
+			continue
+		}
+		newAmount := onHold.Sub(toRelease.Amount)
+		if newAmount.IsNegative() {
+			errs = append(errs, fmt.Errorf("cannot release %s from hold for %s: account only has %s%s on hold", toRelease, addr, onHold, toRelease.Denom))
+			continue
+		}
+		err = k.setHoldCoinAmount(store, addr, toRelease.Denom, newAmount)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to release %s from hold for %s: %w", toRelease, addr, err))
```
Taztingo, line 161: I think you are missing a `continue` here. If `setHoldCoinAmount` fails then it will still get added to fundsReleased.

testutil/assertions/errors_test.go
```diff
@@ -0,0 +1,325 @@
...
+			expOutput: []string{"An error is expected but got nil.", "Error was expected to contain: \"not in error\""},
+		},
+		{
+			name:      "nil error two contains",
+			theError:  nil,
+			contains:  []string{"missing", "fiddlesticks"},
+			expOutput: []string{"An error is expected but got nil.", "Error was expected to contain:\n\t\t\"missing\"\n\t\t\"fiddlesticks\""},
+		},
+		{
+			name:      "with error nil contains",
```
Taztingo, line 39: Might be helpful to have a similar test, but where contains is []string{}.

x/hold/hold_test.go
nullpointer0x00: I agree here.  I'm don't really like this pattern of using an anonymous function as a constructor instead of
 using production code constructors in tests.  I can easily see mistakes being done creating these and they are a bit harder to read.
Also, the intent of using constructors was to provide a single path to initialize objects into the proper state for production side code.  Therefore, when the constructor changes due to new fields being added, all usages of the old constructor will break.  I find a lot of value in this.

x/hold/simulation/genesis.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// RandomizedGenState generates a random GenesisState for the hold module.
+func RandomizedGenState(simState *module.SimulationState) {
+	holdGenState := hold.DefaultGenesisState()
+
+	simState.AppParams.GetOrGenerate(
+		simState.Cdc, HoldAccountHolds, &holdGenState.Holds, simState.Rand,
+		func(r *rand.Rand) {
+			holdGenState.Holds = RandomAccountHolds(r, simState.Accounts)
```
Taztingo, line 90: Will any of the sims generate a bad account address? If not, it may be helpful to sometimes create a bad one to sim test validation.

x/hold/hold_test.go
SpicyLemon: Not quite. `NewInt64Coin` runs validation and panics if there's something wrong.

x/hold/hold_test.go
SpicyLemon: I didn't make a `NewAccountHold(addr, coins)` because for production code, I'd need the `addr` to be an `sdk.AccAddress`. In these unit tests, though, I need to be able to directly set the string so that I can test when it's not valid. So even if I had made `NewAccountHold`, I wouldn't be helpful in here. And in the production code, it was handier to just use struct initialization to make them.

There's four primary reasons why I use anonymous funcs in the tests like this:
1. Smaller diff when a field is added. **If I didn't break a test case, I shouldn't have to make a change to it.** When a field is added and the production constructor gets a new argument, you have to update all uses of that constructor. If you have to update all tests cases just so it'll compile, it makes it harder to notice (in the diff) which test cases actually broke and needed a logical change. Identifying those is an important part of evaluating the correctness a change. Plus, a smaller diff means that you could make your change faster, and requires less time to review.
2. Provide a constant to one or more fields. If every test case has the same value for a certain thing, it's better to move that value out of the test case definition. That way, the next person to look at the test case doesn't have to look at the value and wonder what bearing it has on that specific test case. If it's inconsequential to the case, it shouldn't be part of the case definition.
3. Error checking. The anonymous funcs can call `require.NoError` and return only the thing we want. This is preferable to using funcs that might panic because panics can disrupt an entire chunk of tests (not just the one `Test...` func or test case the panic occurs in). By removing the `error` from the return values, it makes it easier to define things specific to a test case in the test case. I.e. you don't have to create it above, handle the error, then use the variable in the test case.
4. Brevity. Test cases can be pretty complex. Removing commonly repeated, easy to deduce text can help make a test case easier to understand; it can improve the signal-to-noise ratio. Being able to identify the differences between two test cases is made easier when it's easier to see both at the same time, and when you have less to ignore. This identification is most important when a test case unexpectedly fails, and you need to know what makes it special.

Honestly, 95% of our constructors are worthless. In some cases, they actually make things worse. The only ones that are worthwhile are ones that either set private fields or ones that do some conversion or processing (other than calling `.String()`). If there's more than a few fields, they actually make things worse because you can't call functions using named parameters, so you're having to count arguments to see which argument in the call gets set to which field. This is made even worse when there's multiple arguments of the same type in a row.

Most of our stuff doesn't have anything private or require any extra processing, though. As far as I can tell, the only reason we have most of those constructors is because a lot of us came from the java world, where you don't have struct initialization like you do in go. And since go doesn't have function overloading, you're limited to a single `New...` constructor, which quite often ends up not being useful; we end up jumping through hoops to be able to use it when we could just be using struct initialization and save ourselves some effort.

In production code, brevity isn't nearly the concern it is in test cases. It's usually better to have named parameters that you get with struct initialization. Test cases are like a table of data where it's useful to be able to see several similar things at once. In the production code, it's usually not as important to compare the differences between two lines. In the cases where that is important, usually the verboseness of struct initialization with named arguments is more helpful than trying to compare function arguments, specially if the lines are far apart or when there's so many arguments, they're spread across multiple lines.

In unit tests, it's best to isolate the thing your testing as much as possible. When you use a production constructor in a test, now you're test relies on that constructor too. If there's a problem with the constructor, you only want the constructor tests to fail. You don't want other unrelated tests to also fail because it muddies the picture of what's actually been broken. While it's sometimes useful to have a case or two covering that interaction (e.g. a test case on `DefaultGenesis().Validate()`), most of the test cases should not have such a dependency.

Ultimately, the production constructors cause problems, extra work, and little-to-no benefit.

x/hold/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,314 @@
...
+	tests := []struct {
+		name        string
+		setup       func(s *TestSuite, store sdk.KVStore)
+		request     *hold.GetAllHoldsRequest
+		expHolds    []*hold.AccountHold
+		expPageResp *query.PageResponse
+		expErr      []string
+	}{
+		{
+			name:        "nil request",
```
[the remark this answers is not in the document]
SpicyLemon: Ultimately, it comes down to a separation of concerns. The code in an endpoint shouldn't concern itself with the behavior of the caller. The endpoint should only concern itself with it's own behavior, and only return errors that are insurmountable.

When all parameters of the query are optional, there's no reason to require that there's a request. I.e. there's no reason to write code that returns an error if it's not actually a problem. This is common in the `GetAll` queries that only have pagination parameters. Since there's nothing in the request that's required, there's no reason to return an error when there isn't a request.

It might be possible that all code paths that end up calling the query endpoints will always provide a non-nil request, but I can't say that for certain. There might be some legitimate code paths that don't create the request object if there's nothing to put in it. For example, maybe some http get without any http query parameters ends up coming into the endpoint func with a nil request.

x/hold/simulation/genesis.go
SpicyLemon: They'll all be valid. Introducing an invalid one would actually cause the sim to fail. There's no way to communicate that you're introducing something that's expected to error during genesis. And even if there were, such an error would then prevent any of the rest of the sim from doing anything.

x/hold/events_test.go
SpicyLemon: We aren't actually testing `sdk.TypedEvenToEvent`, we're testing our new events being provided to that function.

The only reason that `EmitTypedEvent` returns an error is if `TypedEvenToEvent` returns an error. `TypedEvenToEvent` will only return an error if there's json marshal/unmarshal problems which will only happen if there's something structurally wrong with the event type. Basically, if `TypedEvenToEvent` returns an error for one instance of a type, it'll return an error for all instances of that type.

Since these event types are designed to be provided to `EmitTypedEvent` (and thus `TypedEventToEvent`), I felt it prudent to make sure they can be safely converted.

I was even considering ignoring the error from `EmitTypedEvent` because of this, but couldn't quite bring myself to do it. I'd bet that the funcs that call `EmitTypedEvent` in here already need to return an error field anyway, so doing that probably doesn't buy me anything.

testutil/assertions/events.go
SpicyLemon: I thought about doing that, but decided to leave it here since it technically isn't even related to unit testing other than being used by the event string creation stuff. Since it was written for that event stuff, I thought it was more at home with the rest of the event stuff.

x/hold/hold_test.go
SpicyLemon: I did go through and rename the `ah` funcs to `accountHold` as well as the `c` and `cz` funcs.

testutil/assertions/errors_test.go
SpicyLemon: Added a couple test cases with an empty `contains`.

x/hold/keeper/keeper.go
SpicyLemon: Good catch. Fixed.

x/hold/keeper/grpc_query_test.go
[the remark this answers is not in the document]
Taztingo: Agreed, if I look at just the function then that input is fine.

x/hold/hold_test.go
Taztingo: My main concern was the readability of it because I was having a hard time following the tests, and I was seeing tests redefine the same anonymous functions. A constructor was just one proposed solution., but it seems like you fixed my concern. 

I find benefits in both approaches, but constructors do have some additional benefits.

1. You can ensure object creation all follows the same type of creational pattern. For example all types must follow this, all non pods must follow this, no types must follow this, etc...
2. Removal of code redundancy with New and Init functions.
3. This goes against your first point, and you cannot always guarantee a test is valid unless the objects that it uses are in a correct state. It's very easy to add a new mandatory field to a struct, and forget to update all instances of the manually initialized struct. The compiler cannot guide the user in this case because there is no constructor, and it is up to the user to update their tests to ensure all tests are in valid starting state. These passing tests may or may not be working anymore. See the following example:

```
type S struct {
	a string
	b string // Newly added mandatory field
}

func Foo(s S) int {
	// Oops, I acidentally commented this out
	/*if len(s.a) == 0 {
		return 1
	}*/ 

	// Newly added code
	if len(s.b) == 0 {
		return 1
	}

	return 0
}

// This test is now broken without the compiler telling you
// It can be very hard to find all instances of it in bigger projects
// We can easily pass an S through this without a valid value for A.
func TestAFooIsNotInvalid(t *testing.T) {
	s := S {}
	res := Foo(s)

	// This passes, but the code is not correct.
	// An invalid A will make it past the check
	assert(res, 1)
}
```

It might be better to discuss this offline as a team so that if we are choosing to go constructors or no constructors in tests, then we know all the benefits/risks we are getting from this code style.

Taztingo (approved):
LGTM

nullpointer0x00 (approved):

iramiller (approved):

Issue #1497, opened 2023-04-21
Cross Chain Query/Oracle Concept for Private Zones


iramiller:
## Summary

One of the common use cases for a private zone is maintaining detailed account records or management of digital assets.  In some cases it may be useful to selectively disclose data about these resources to another zone.  This enhancement would provide a common interface for these queries to be managed by the zone.

## Problem Definition

Within a private zone many types of data may be maintained that should not be fully public.  A useful example is a zone that maintains account balances for private equity/cap table management.  If the entire cap table was available on the public main net more information than desired could be inferred by reviewing the relative account balances.  Despite this it may still be useful to return limited information such as the existence of an account or if an account possesses sufficient balance for a given operation.  This is commonly handled by a Data Oracle which can selectively disclose information.

With the desire for a private zone to be able to run the standard Provenance binary without modifications an approach is needed for the zone operator to specify an implementation of the query processing logic in a way that doesn't require a fork of the blockchain software.

## Proposal

Add support for [cross chain queries](https://github.com/strangelove-ventures/async-icq) but modify the process for checking a query allow list such that the actual processing logic is handed off to a smart contract instance.  This instance should be configured as a parameter such that governance or genesis setup could define the address to use.  If the contract instance does not exist or is otherwise defined then the query should be rejected as unsupported.

The example [async-icq](https://github.com/strangelove-ventures/async-icq) module packages up the queries using the protobuf+ABCI wrappers.  This same method could be supported by the smart contract implementing the query endpoint. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
The [Osmosis rate limit system](https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit) uses a wasm contract for implementation and might be a useful reference...

PR #1643, opened 2023-08-03
Adds ICQHost and Oracle Module

Taztingo:
## Description

Adds the ICQHost module that allows chains to query against chains using the Provenance binary. The oracle module implements the controller portion of ICQ and allows users to query against an Oracle smart contract on another chain.

closes: #1497

---

write a little note why.


app/upgrades.go
```diff
@@ -293,3 +311,11 @@ func fixNameIndexEntries(ctx sdk.Context, app *App) {
 func setAccountDataNameRecord(ctx sdk.Context, accountK attributetypes.AccountKeeper, nameK attributetypes.NameKeeper) (err error) {
 	return attributekeeper.EnsureModuleAccountAndAccountDataNameRecord(ctx, accountK, nameK)
 }
+
+// setAccountDataNameRecord makes sure the account data name record exists, is restricted,
+// and is owned by the attribute module. An error is returned if it fails to make it so.
+func setupICQ(ctx sdk.Context, app *App) {
```
SpicyLemon, line 317: Wrong comment.

proto/provenance/oracle/v1/params.proto
```diff
@@ -0,0 +1,14 @@
...
+
+option go_package = "github.com/provenance-io/provenance/x/oracle/types";
+
+option java_package        = "io.provenance.oracle.v1";
+option java_multiple_files = true;
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+}
```
SpicyLemon, line 14: If there aren't any params, I say we just leave it out completely.

The `x/params` module is deprecated anyway, so probably best to not add something new to it.

x/oracle/types/msgs.go
```diff
@@ -0,0 +1,62 @@
...
+		return fmt.Errorf("invalid channel id")
+	}
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
+		return fmt.Errorf("invalid authority address: %w", err)
+	}
+	if err := msg.Query.ValidateBasic(); err != nil {
+		return fmt.Errorf("invalid query data: %w", err)
+	}
+	return nil
+}
```
SpicyLemon, line 62: Please keep the functions related to a specific type all together.

I.e. please move these up under `NewMsgSendQueryOracle`.

x/oracle/types/msgs.go
```diff
@@ -0,0 +1,62 @@
+package types
+
+import (
+	fmt "fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	host "github.com/cosmos/ibc-go/v6/modules/core/24-host"
+)
+
+var _, _ sdk.Msg = &MsgUpdateOracleRequest{}, &MsgSendQueryOracleRequest{}
```
SpicyLemon, line 10: Please split this onto multiple lines and put them in the same order that they're in in this file.
```suggestion
var (
	_ sdk.Msg = &MsgSendQueryOracleRequest{}
	_ sdk.Msg = &MsgUpdateOracleRequest{}
)
```

Alternatively, put them in a slice like we have in a couple modules (e.g. [Metadata](https://github.com/provenance-io/provenance/blob/main/x/metadata/types/msg.go#L49-L80)). Then that slice can be used in places where we always need to remember to add new message type definitions, and we've reduced possible pain points.

x/oracle/keeper/keeper.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// Logger returns the correctly named logger for the module
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", "x/"+types.ModuleName)
+}
+
+// BindPort stores the provided portID and binds to it, returning the associated capability
+func (k Keeper) BindPort(ctx sdk.Context, portID string) error {
```
SpicyLemon, line 72: Is this supposed to return the `capability` too or is the comment off?

x/oracle/keeper/keeper.go
```diff
@@ -0,0 +1,108 @@
...
+	host "github.com/cosmos/ibc-go/v6/modules/core/24-host"
+
+	"github.com/provenance-io/provenance/x/oracle/types"
+)
+
+type Keeper struct {
+	cdc        codec.BinaryCodec
+	storeKey   storetypes.StoreKey
+	memKey     storetypes.StoreKey
+	paramstore paramtypes.Subspace
```
SpicyLemon, line 24: The `x/params` module is deprecated. It's "replacement" is to have each module do their own thing for params. E.g. in the `x/sanction` module, they're stored as `<type byte> | <field name> -> <field value>`. See the stuff below https://github.com/provenance-io/cosmos-sdk/blob/main-pio/x/sanction/keeper/keeper.go#L287.

app/upgrades_test.go
```diff
@@ -415,6 +415,19 @@ func (s *UpgradeTestSuite) TestRust() {
...
+
+	expInLog := []string{
+		"INF Starting module migrations. This may take a significant amount of time to complete. Do not restart node.",
+		"INF Updating ICQ params",
+		"INF Done updating ICQ params",
+	}
+
+	s.AssertUpgradeHandlerLogs("saffron", expInLog, nil)
+}
+
```
SpicyLemon, line 430: We'll want a `TestSaffronRC1` in here too.

app/upgrades.go
```diff
@@ -112,6 +115,21 @@ var upgrades = map[string]appUpgrade{
...
+	"saffron": { // upgrade for v1.17.0,
+		Handler: func(ctx sdk.Context, app *App, vm module.VersionMap) (module.VersionMap, error) {
+			var err error
+			vm, err = runModuleMigrations(ctx, app, vm)
+			if err != nil {
+				return nil, err
+			}
+
+			setupICQ(ctx, app)
+
```
SpicyLemon, line 127: We probably want `removeInactiveValidatorDelegations(ctx, app)` in here too.

app/upgrades.go
```diff
@@ -112,6 +115,21 @@ var upgrades = map[string]appUpgrade{
...
+			return vm, nil
+		},
+		Added: []string{icqtypes.ModuleName, oracletypes.ModuleName},
+	},
+	"saffron-rc1": {},
```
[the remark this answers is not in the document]
SpicyLemon: Also, the `-rc1` upgrade should be defined above the non-rc entry since that's how they get used chronologically.

x/oracle/keeper/msg_server_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			res, err := s.msgServer.UpdateOracle(s.ctx, tc.req)
+
+			if tc.event != nil {
+				s.Assert().Equal(1, len(s.ctx.EventManager().Events()), "should emit the correct number of events")
+				s.Assert().Equal(*tc.event, s.ctx.EventManager().Events()[0], "should emit the correct event")
+			} else {
+				s.Assert().Equal(0, len(s.ctx.EventManager().Events()), "should not emit events")
```
SpicyLemon, line 53: `.Empty` would be a better assertion here. That way, if it fails, the failure message will include the events which will likely help identify why it failed.

x/oracle/keeper/msg_server_test.go
```diff
@@ -0,0 +1,132 @@
...
+			if tc.mockChannel {
+				s.app.OracleKeeper.SetMockChannelKeeper(&keeper.MockChannelKeeper{})
+			}
+			res, err := s.msgServer.SendQueryOracle(s.ctx, tc.req)
+
+			if tc.event != nil {
+				s.Assert().Equal(1, len(s.ctx.EventManager().Events()), "should emit the correct number of events")
+				s.Assert().Equal(*tc.event, s.ctx.EventManager().Events()[0], "should emit the correct event")
+			} else {
+				s.Assert().Equal(0, len(s.ctx.EventManager().Events()), "should not emit events")
```
SpicyLemon, line 120: Same here with `.Empty`.

x/oracle/keeper/query_server.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/provenance-io/provenance/x/oracle/types"
+)
+
+var _ types.QueryServer = Keeper{}
+
+// QueryAddress returns the address of the module's oracle
+func (k Keeper) OracleAddress(goCtx context.Context, req *types.QueryOracleAddressRequest) (*types.QueryOracleAddressResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
```
SpicyLemon, line 21: Since the request doesn't have anything in it, we don't really care whether or not one comes in.
```suggestion
func (k Keeper) OracleAddress(goCtx context.Context, _ *types.QueryOracleAddressRequest) (*types.QueryOracleAddressResponse, error) {
```

x/oracle/keeper/query_server_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			if tc.mockEnabled {
+				s.app.OracleKeeper.SetWasmQueryServer(keeper.MockWasmServer{})
+			}
+
+			if len(tc.oracle) > 0 {
+				s.app.OracleKeeper.SetOracle(s.ctx, sdk.MustAccAddressFromBech32(tc.oracle))
+			}
```
SpicyLemon, line 116: A word of caution on this: these lines are doing things that will affect the tests cases to follow.

There's two workarounds:
1. Change the setters into `With...` functions, create a keeper variable to hold the keeper to use for the test, and, if mocking, call the `With...` function and asign the result back to the keeper variable.
2. Before calling the setter, get the current thing and create a `defer` that sets it back to its original value.

Example of what `WithWasmQueryServer` would look like:
```golang
func (k Keeper) SetWasmQueryServer(server wasmtypes.QueryServer) Keeper {
	k.wasmQueryServer = server
	return k
}
```

Since it's got a concrete receiver (no `*`), it doesn't end up updating the existing keeper; it ultimately creates a copy of it with the new `wasmQueryServer` value.

Relatedly, with mocks, it's usually best to always mock if you're going to mock any test cases. The place where it might cause problems is if you've got a test that you expect to not need the mocked stuff, but suddenly it does. Now you're making those calls for real which might be causing other problems.

app/upgrades.go
```diff
@@ -3,6 +3,8 @@ package app
 import (
 	"fmt"
 
+	icqtypes "github.com/strangelove-ventures/async-icq/v6/types"
```
SpicyLemon, line 6: Ignore this comment. Github is being dumb.

x/oracle/keeper/icq.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	reqs := []abcitypes.RequestQuery{
+		{
+			Path: "/provenance.oracle.v1.Query/Oracle",
+			Data: k.cdc.MustMarshal(&q),
+		},
+	}
+
+	timeoutTimestamp := ctx.BlockTime().Add(time.Minute).UnixNano()
```
SpicyLemon, line 38: It surprises me that nano is used here. It also annoys me that they don't document anywhere that it should be nanoseconds. I had to dig down to where it's used; it's ultimately compared to `consensusState.GetTimestamp()` which is documented to be in nano.

I guess max uint64 as nanoseconds since epoch is still `2554-07-21` (a Sunday), so we've got more time than I expected.

(just musing, feel free to resolve this comment)

x/oracle/keeper/query_server.go
```diff
@@ -0,0 +1,47 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/provenance-io/provenance/x/oracle/types"
+)
+
+var _ types.QueryServer = Keeper{}
+
+// QueryAddress returns the address of the module's oracle
+func (k Keeper) OracleAddress(goCtx context.Context, req *types.QueryOracleAddressRequest) (*types.QueryOracleAddressResponse, error) {
```
SpicyLemon, line 18: I bet the linter is complaining about `req` not being used. You'll want to change it to `_`.

x/oracle/types/errors.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import (
+	cerrs "cosmossdk.io/errors"
+)
+
+var (
+	ErrSample               = cerrs.Register(ModuleName, 2, "sample error")
```
SpicyLemon, line 8: Probably don't need this one.

x/oracle/types/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+
+// DefaultGenesis returns the default oracle genesis state
+func DefaultGenesis() *GenesisState {
+	return NewGenesisState(PortID, 1, "")
+}
+
+// Validate performs basic genesis state validation returning an error upon any
+// failure.
+func (gs GenesisState) Validate() error {
+	if gs.Sequence == 0 {
```
SpicyLemon, line 26: This `Sequence` value is the "last sequence value used" right? If that's the case, it seems like a value of `0` should be allowed.

x/oracle/module/module_ibc.go
```diff
@@ -0,0 +1,167 @@
...
+// OnTimeoutPacket implements the IBCModule interface
+func (am AppModule) OnTimeoutPacket(
+	ctx sdk.Context,
+	modulePacket channeltypes.Packet,
+	_ sdk.AccAddress,
+) error {
+	return am.keeper.OnTimeoutPacket(ctx, modulePacket)
+}
+
+func (am AppModule) NegotiateAppVersion(
```
nullpointer0x00, line 158: Add method comment.

x/oracle/simulation/operations.go
```diff
@@ -0,0 +1,198 @@
...
+		),
+		simulation.NewWeightedOperation(
+			weightMsgSendOracleQuery,
+			SimulateMsgSendQueryOracle(k, ak, bk, ck),
+		),
+	}
+}
+
+// SimulateMsgCreateTrigger sends a MsgUpdateOracle.
+func SimulateMsgUpdateOracle(_ keeper.Keeper, ak authkeeper.AccountKeeperI, bk bankkeeper.Keeper) simtypes.Operation {
```
nullpointer0x00, line 65: Wrong comment method name

x/oracle/types/expected_keepers.go
```diff
@@ -0,0 +1,40 @@
...
+		chanCap *capabilitytypes.Capability,
+		sourcePort string,
+		sourceChannel string,
+		timeoutHeight clienttypes.Height,
+		timeoutTimestamp uint64,
+		data []byte,
+	) (sequence uint64, err error)
+}
+
+// Channel defines the expected IBC channel keeper
```
nullpointer0x00, line 23: ChannelKeeper

x/oracle/simulation/operations.go
```diff
@@ -0,0 +1,198 @@
...
+	return func(
+		r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simtypes.Account, chainID string,
+	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
+		raccs, err := RandomAccs(r, accs, uint64(len(accs)))
```
SpicyLemon, line 69:
```suggestion
		raccs, err := RandomAccs(r, accs, 2)
```

SpicyLemon: Eh. Nevermind. I guess it doesn't really matter.

x/oracle/client/cli/query.go
```diff
@@ -0,0 +1,61 @@
...
+	queryCmd := &cobra.Command{
+		Use:                        types.ModuleName,
+		Short:                      "Querying commands for the oracle module",
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+	queryCmd.AddCommand(
+		GetQueryOracleAddressCmd(),
+	)
```
SpicyLemon, line 27: Missing a command for the `Oracle` query.

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/cosmos/cosmos-sdk/version"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	govcli "github.com/cosmos/cosmos-sdk/x/gov/client/cli"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+
+	"github.com/provenance-io/provenance/x/oracle/types"
+)
+
+// The flag to specify that the command should be ran as a gov proposal
+const FlagGovProposal = "gov-proposal"
```
SpicyLemon, line 26: You should only need this flag if there are endpoints that are sometimes used as a normal `Tx`, and sometimes as a governance proposal.

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,172 @@
...
+			if govErr != nil {
+				return govErr
+			}
+			proposal.Messages = anys
+			var req sdk.Msg = proposal
+
+			if err := msg.ValidateBasic(); err != nil {
+				return err
+			}
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), req)
```
SpicyLemon, line 83: Since you've already got the `govErr` variable created, you can just assign directly to `proposal.Messages`. Also, I don't think you need to create the `req` variable since the proposal is already an `sdk.Msg` and can be provided to  GenerateOrBroadcastTxCLI.

```suggestion
			proposal.Messages, govErr = sdktx.SetMsgs([]sdk.Msg{msg})
			if govErr != nil {
				return govErr
			}

			if err := msg.ValidateBasic(); err != nil {
				return err
			}
			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), proposal)
```

x/oracle/client/cli/query.go
Taztingo: I opted not to expose it since it's the same as doing a wasm query. The only reason I have a separate message for it is to make querying the oracle easier from another chain. The user doesn't need to look up an address on a remote chain.

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,172 @@
...
+	case 0:
+		return asciiDecodeString(s)
+	case 1:
+		return hex.DecodeString(s)
+	case 2:
+		return base64.StdEncoding.DecodeString(s)
+	default:
+		return a.dec(s)
+	}
+}
```
SpicyLemon, line 168: I'm not sure what the purpose of the `asciiF`, `hexF`, and `b64F` bools is. They're never assigned to, so they're always false.

I feel like there's a missing flag on the command that would let the user tell this how to try to decode it. As it is, it'll only ever use `asciiDecodeString`.

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,172 @@
...
+	cmd.Flags().Bool(FlagGovProposal, false, "Run transaction as gov proposal")
+	govcli.AddGovPropFlagsToCmd(cmd)
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// GetCmdSendQuery is a command to send a query to another chain's oracle
+func GetCmdSendQuery() *cobra.Command {
+	decoder := newArgDecoder(asciiDecodeString)
```
SpicyLemon, line 96: This should be something created inside the cmd, and probably have it pay attention to a flag value so it knows how it's supposed to try to decode the thing.

x/oracle/types/genesis.go
Taztingo: I actually don't think the sequence is needed in state anymore. Good catch.

x/oracle/client/cli/tx.go
Taztingo: This was taken from wasmd, but I can simplify it for our use case.

x/oracle/spec/01_concepts.md
```diff
@@ -0,0 +1,31 @@
...
+
+# Concepts
+
+The oracle module is very minimal, but users should understand what the `Oracle` is and how it interacts with `ICQ`.
+
+<!-- TOC 2 -->
+- [Concepts](#concepts)
+  - [Oracle](#oracle)
+  - [Interchain Queries (ICQ)](#interchain-queries-icq)
+    - [Note](#note)
```
SpicyLemon, line 13: The TOCs should be auto-generated using the `make update-tocs` command.

That command looks for a `<!-- TOC -->` comment in an `.md` file and generates the table of contents from the section headings. One or two numbers can be provided to control the levels of headers to involve. By default, entries are made for the level 2 and 3 headers. Providing a single number (e.g. `<!-- TOC 2 -->`) changes the maximum level from 3 to whatever was provided. So with `TOC 2` you only get the level 2 headers; with `TOC 4` you'd get all 2, 3, and 4 levels. Providing two numbers (e.g. `<!-- TOC 1 5 -->`) sets both the min and max levels to use. In that example, you'd get entries for all level 1, 2, 3, 4, and 5 headers.

Usually you don't want the level 1 header because there's only one in the file that applies to the whole file and is also above the TOC.

For this one, `TOC 2` is probably what's wanted since we don't really need an entry for that note.

It looks like you've got the `<!-- TOC ... -->` comments in the files just fine, but manually created them. So you'll want to run `make update-tocs` and make sure you've got all the entries you think should be in the TOC for that file.

SpicyLemon: Note that `make update-tocs` will also fix a missing entry in one of the marker spec docs. That's what's nice about having them auto-generated. Go ahead and include that change in this PR.

x/oracle/client/cli/tx.go
SpicyLemon: I think something is missing from that copy then, because those flags it loops through are never anything other than `false`.

SpicyLemon: Ah! I think, maybe, there's some flag bindings that should be added to that command that bind a flag value to the fields in the decoder.  That would explain why the decoder is created outside of the command runner.

x/oracle/spec/01_concepts.md
Taztingo: Hm, I ran the update-toc.sh script manually. I'll run it through make from now on.

x/oracle/client/cli/tx.go
Taztingo: I ended up removing this entirely and used the default decoder since we didn't have a use case for multiple types. Thanks for pointing this out.

x/oracle/client/cli/tx.go
Taztingo: Comment from above applies here.

x/oracle/keeper/query_server_test.go
Taztingo: I'm happy you mentioned this. This is exactly what I wanted to do.

x/oracle/keeper/keeper_test.go
```diff
@@ -0,0 +1,60 @@
+package keeper_test
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/99designs/keyring"
```
iramiller, line 8: It seems like we want to avoid this reference and instead rely on the Cosmos SDK replacement ... is it possible to create this test without this?

Taztingo: Sure, I can replace it with the Cosmos SDK one.

app/app.go
```diff
@@ -1215,10 +1267,12 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(wasm.ModuleName)
 	paramsKeeper.Subspace(rewardtypes.ModuleName)
 	paramsKeeper.Subspace(triggertypes.ModuleName)
+	paramsKeeper.Subspace(oracletypes.ModuleName)
```
SpicyLemon, line 1270: Don't need this since the oracle module doesn't use the (deprecated) params module.

x/oracle/keeper/msg_server_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			if tc.mockChannel {
+				s.app.OracleKeeper = s.app.OracleKeeper.WithMockChannelKeeper(&keeper.MockChannelKeeper{})
+			}
+			res, err := s.msgServer.SendQueryOracle(s.ctx, tc.req)
+
+			if tc.event != nil {
+				s.Assert().Equal(1, len(s.ctx.EventManager().Events()), "should emit the correct number of events")
```
SpicyLemon, line 117: Nit for the future: If you use the `.Len` assertion, the failure message will contain the contents of the slice when it fails. E.g. in this case, if an extra event gets emitted, the failure message will state "The length was expected to be 1, but was 2, and here's the contents of the thing." It makes it easier to decide whether some code is broken or the test needs updating.

Also, it's helpful in the long run if you assign things to variables and run the assertion on those variables. Primarily, it makes it easier when you have to run the test in a debugger because you've got direct access to the value.

Just wanted to share. Feel free to resolve this comment.

x/oracle/keeper/query_server_test.go
```diff
@@ -0,0 +1,128 @@
...
+				Query: []byte("{}"),
+			},
+			err: "missing oracle address",
+		},
+		{
+			name: "success - should handle error from contract",
+			req: &types.QueryOracleRequest{
+				Query: []byte("{}"),
+			},
+			err: "missing oracle address",
```
SpicyLemon, line 85: I think this is a failure cause, right?

x/oracle/spec/02_state.md
```diff
@@ -0,0 +1,27 @@
+<!--
+order: 2
+-->
+
+# State
+
+The oracle module manages the address of the Oracle and the ICQ state.
+
+<!-- TOC 2 -->
+- [State](#state)
```
SpicyLemon, line 10:
```suggestion
```

The top-level header doesn't need to be in the TOC, and isn't included when you `make update-tocs`.

nullpointer0x00 (approved):

SpicyLemon (approved):

nullpointer0x00 (approved):

Issue #1626, opened 2023-07-10
Accepting Quarantined Restricted Marker Errors


tbrooks-figure:
## Summary of Bug

When a restricted marker gets quarantined due to the recipient having opted in, the restricted marker cannot be accepted unless the restricted marker grants transfer auth to the quarantine module.

## Version

v1.16.0

## Steps to Reproduce
1. Create a restricted marker
2. Set up a wallet and opt in to quarantine
3. Send some of the restricted marker coin to the quarantined address
4. Attempt to accept the quarantined funds. Ex: 
```
provenanced tx quarantine accept quarantine-tt4 tp1j82e50f7qp5slv8fu9ymffapf7vxhtsr5xcka6 -t --node http://localhost:26657 --chain-id testing --gas auto --gas-prices 1905nhash --gas-adjustment 1.3
Error: rpc error: code = Unknown desc = rpc error: code = Unknown desc = failed to execute message; message index: 0: tp1cttxw40x6z6z77j5rp7qyr6vsrmhjcs2czz9lg does not have transfer permissions [provenance-io/provenance/vendor/github.com/cosmos/cosmos-sdk/baseapp/baseapp.go:859] With gas wanted: '18446744073709551615' and gas used: '59673' : unknown request
```

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
It make sense to me to have some hard-coded whitelist addresses for the marker module. These addresses wouldn't be required to have the required attributes, and wouldn't be required to have transfer authority. We'd want to do some analysis on all the aspects of the marker send restriction to figure out which can be bypassed here. As part of that analysis, we should consider pros/cons of applying the marker module's `SendRestrictionFn` before/after the quarantine module's.

The addresses we'd want to whitelist are the quarantine module, fee collector, reward keeper, and maybe the bonded and not-bonded pools. The bonded and not-bonded pools can only ever hold the bond denom. So for mainnet, those wouldn't need to be whitelisted, but if a private zone wanted to use a restricted token as the bond denom, we'd want them whitelisted too.

PR #1652, opened 2023-08-15
Allow restricted coins to be quarantined

SpicyLemon:
## Description

closes: #1626 

This PR fixes a bug related to quarantine of restricted coins, and also enables the use of restricted coins for other things that require an intermediary holding account. It also prevents forced transfers from module accounts.

Quarantine of restricted coins is fixed by adding a hard-coded list of addresses that bypass the required attributes checking. That list consists of the following module account addresses: 

- `authtypes.FeeCollectorName` - Allows paying fees with restricted coins.
- `reward` - Allows reward programs to use restricted coins.
- `quarantine` - Allows quarantine and acceptance of quarantined coins.
- `gov` - Allows deposits to have quarantined coins.
- `distribution` - Allows collection of delegation and fee rewards in restricted coins.
- `stakingtypes.BondedPoolName` - Allows delegation of restricted coins.
- `stakingtypes.NotBondedPoolName` - Allows delegation of restricted coins.

For restricted markers without required attributes:
* If the `toAddr` is a bypass account, the `fromAddr` must have transfer authority.
* If the `fromAddr` is a bypass account, it's assumed that the funds got where they currently are because someone with transfer authority got them there, so this transfer is allowed.

For restricted markers with required attributes:
* If the `toAddr` is a bypass account, the transfer is allowed regardless of whether the `fromAddr` has transfer authority. It's assumed that the next destination's attributes will be properly checked before allowing the funds to leave the bypass account.
* If the `fromAddr` is a bypass account, the `toAddr` must have the required attributes.

Transfers of restricted markers between bypass accounts is allowed (e.g. bonded to not-bonded, or fee collector to distribution).

There's a special case when the `toAddr` is a marker. In that case, no matter what, the `fromAddr` must have deposit permissions on that marker. That will almost certainly not be the case for these bypass accounts, so a "deposit" from, e.g. a rewards program will not work.

---

When a transfer of restricted coins is being sent to an account that's opted into quarantine, there's some possibly confusing flow behavior.

The marker's `SendRestrictionFn` is applied before quarantine's.

The initial send will get into the marker's send restriction with the same `to` and `from` as the send request. I.e. neither will be a bypass account. Any restricted coins without required attributes will require transfer authority. Any restricted coins with required attributes will require that the `to` has them. If there isn't a problem, it'll then get to the quarantine restriction which will change the destination to the quarantine module.

Later, when the user attempts to accept the funds, the `from` will be the quarantine module, which is a bypass account. Any restricted coins without required attributes are okayed. Any restricted coins with required attributes will again require the `to` to have them.

---

write a little note why.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -479,8 +479,19 @@ func New(
 		appCodec, keys[metadatatypes.StoreKey], app.GetSubspace(metadatatypes.ModuleName), app.AccountKeeper, app.AuthzKeeper, app.AttributeKeeper,
 	)
 
+	markerWhitelistAddrs := []sdk.AccAddress{
+		authtypes.NewModuleAddress(authtypes.FeeCollectorName),     // Allow collecting fees in restricted coins.
+		authtypes.NewModuleAddress(rewardtypes.ModuleName),         // Allow rewards to hold onto restricted coins.
+		authtypes.NewModuleAddress(quarantine.ModuleName),          // Allow quarantine to hold onto restricted coins.
+		authtypes.NewModuleAddress(govtypes.ModuleName),            // Allow restricted coins in deposits.
+		authtypes.NewModuleAddress(distrtypes.ModuleName),          // Allow bond denom to be a restricted coin.
+		authtypes.NewModuleAddress(stakingtypes.BondedPoolName),    // Allow bond denom to be a restricted coin.
+		authtypes.NewModuleAddress(stakingtypes.NotBondedPoolName), // Allow bond denom to be a restricted coin.
+	}
 	app.MarkerKeeper = markerkeeper.NewKeeper(
-		appCodec, keys[markertypes.StoreKey], app.GetSubspace(markertypes.ModuleName), app.AccountKeeper, app.BankKeeper, app.AuthzKeeper, app.FeeGrantKeeper, app.AttributeKeeper, app.NameKeeper, app.TransferKeeper,
+		appCodec, keys[markertypes.StoreKey], app.GetSubspace(markertypes.ModuleName),
+		app.AccountKeeper, app.BankKeeper, app.AuthzKeeper, app.FeeGrantKeeper,
+		app.AttributeKeeper, app.NameKeeper, app.TransferKeeper, markerWhitelistAddrs,
 	)
 
 	pioMessageRouter := MessageRouterFunc(func(msg sdk.Msg) baseapp.MsgServiceHandler {
--- a/app/app.go
+++ b/app/app.go
@@ -484,7 +484,7 @@ func New(
 		authtypes.NewModuleAddress(rewardtypes.ModuleName),         // Allow rewards to hold onto restricted coins.
 		authtypes.NewModuleAddress(quarantine.ModuleName),          // Allow quarantine to hold onto restricted coins.
 		authtypes.NewModuleAddress(govtypes.ModuleName),            // Allow restricted coins in deposits.
-		authtypes.NewModuleAddress(distrtypes.ModuleName),          // Allow bond denom to be a restricted coin.
+		authtypes.NewModuleAddress(distrtypes.ModuleName),          // Allow fee denoms to be restricted coins.
 		authtypes.NewModuleAddress(stakingtypes.BondedPoolName),    // Allow bond denom to be a restricted coin.
 		authtypes.NewModuleAddress(stakingtypes.NotBondedPoolName), // Allow bond denom to be a restricted coin.
 	}
--- a/app/app.go
+++ b/app/app.go
@@ -479,7 +479,7 @@ func New(
 		appCodec, keys[metadatatypes.StoreKey], app.GetSubspace(metadatatypes.ModuleName), app.AccountKeeper, app.AuthzKeeper, app.AttributeKeeper,
 	)
 
-	markerWhitelistAddrs := []sdk.AccAddress{
+	markerReqAttrBypassAddrs := []sdk.AccAddress{
 		authtypes.NewModuleAddress(authtypes.FeeCollectorName),     // Allow collecting fees in restricted coins.
 		authtypes.NewModuleAddress(rewardtypes.ModuleName),         // Allow rewards to hold onto restricted coins.
 		authtypes.NewModuleAddress(quarantine.ModuleName),          // Allow quarantine to hold onto restricted coins.
@@ -491,7 +491,7 @@ func New(
 	app.MarkerKeeper = markerkeeper.NewKeeper(
 		appCodec, keys[markertypes.StoreKey], app.GetSubspace(markertypes.ModuleName),
 		app.AccountKeeper, app.BankKeeper, app.AuthzKeeper, app.FeeGrantKeeper,
-		app.AttributeKeeper, app.NameKeeper, app.TransferKeeper, markerWhitelistAddrs,
+		app.AttributeKeeper, app.NameKeeper, app.TransferKeeper, markerReqAttrBypassAddrs,
 	)
 
 	pioMessageRouter := MessageRouterFunc(func(msg sdk.Msg) baseapp.MsgServiceHandler {
--- a/x/marker/keeper/keeper_test.go
+++ b/x/marker/keeper/keeper_test.go
@@ -1493,8 +1493,8 @@ func TestUpdateSendDenyList(t *testing.T) {
 	}
 }
 
-func TestWhitelistAddrs(t *testing.T) {
-	// Tests both GetWhitelistAddrs and IsWhitelistAddr.
+func TestReqAttrBypassAddrs(t *testing.T) {
+	// Tests both GetReqAttrBypassAddrs and IsReqAttrBypassAddr.
 	expectedNames := []string{
 		authtypes.FeeCollectorName,
 		rewardtypes.ModuleName,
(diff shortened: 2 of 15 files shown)


app/app.go
```diff
@@ -479,8 +479,19 @@ func New(
 		appCodec, keys[metadatatypes.StoreKey], app.GetSubspace(metadatatypes.ModuleName), app.AccountKeeper, app.AuthzKeeper, app.AttributeKeeper,
 	)
 
+	markerReqAttrBypassAddrs := []sdk.AccAddress{
+		authtypes.NewModuleAddress(authtypes.FeeCollectorName),     // Allow collecting fees in restricted coins.
+		authtypes.NewModuleAddress(rewardtypes.ModuleName),         // Allow rewards to hold onto restricted coins.
+		authtypes.NewModuleAddress(quarantine.ModuleName),          // Allow quarantine to hold onto restricted coins.
+		authtypes.NewModuleAddress(govtypes.ModuleName),            // Allow restricted coins in deposits.
+		authtypes.NewModuleAddress(distrtypes.ModuleName),          // Allow bond denom to be a restricted coin.
```
iramiller, line 487: Technically this is the _fee_ denom and not the _bond_ denom.  In some cases (our mainnet) these two are the same but this is not always true.

app/app.go
```diff
@@ -479,8 +479,19 @@ func New(
...
+		authtypes.NewModuleAddress(stakingtypes.BondedPoolName),    // Allow bond denom to be a restricted coin.
+		authtypes.NewModuleAddress(stakingtypes.NotBondedPoolName), // Allow bond denom to be a restricted coin.
+	}
```
nullpointer0x00, line 490: I like this pattern.

x/marker/keeper/keeper_test.go
```diff
@@ -1488,3 +1492,91 @@ func TestUpdateSendDenyList(t *testing.T) {
...
+	}
+	t.Run("get: only has expected entries", func(t *testing.T) {
+		// This assumes each expectedNames test passed. This is designed to fail if a new entry
+		// is added to the list (in app.go). When that happens, update the expectedNames with
+		// the new entry so it's harder for it to accidentally go missing.
+		actual := app.MarkerKeeper.GetReqAttrBypassAddrs()
+		assert.Len(t, actual, len(expectedNames), "GetReqAttrBypassAddrs()")
+	})
+
+	t.Run("get: called twice equal but not same", func(t *testing.T) {
```
iramiller, line 1532: This seems like a pretty odd test. In what situations would the endpoint return two different sets of addresses when called back to back?

x/marker/keeper/keeper_test.go
```diff
@@ -1488,3 +1492,91 @@ func TestUpdateSendDenyList(t *testing.T) {
...
+		if assert.Equal(t, expected, actual, "GetReqAttrBypassAddrs()") {
+			if assert.NotSame(t, expected, actual, "GetReqAttrBypassAddrs()") {
+				for i := range expected {
+					assert.NotSame(t, expected[i], actual[i], "GetReqAttrBypassAddrs()[%d]", i)
+				}
+			}
+		}
+	})
+
+	t.Run("get: changes to result not reflected in next result", func(t *testing.T) {
```
iramiller, line 1544: It feels like these addresses should probably be "sealed" when the marker module is created.

iramiller (commented):
I was worried about the send restriction order ... glad to see a test that verifies that is wired up correctly.  I feel like the bypass address list might deserve to be sealed on creation given its sensitivity.

x/marker/keeper/keeper_test.go
SpicyLemon: This uses "equal" and "same" in relation to the `assert.Equal` and `assert.Same` tests.

It's asserting that `GetReqAttrBypassAddrs()` isn't just returning `k.reqAttrBypassAddrs` directly, but is instead, returning a deep copy of that slice. Each invocation should return the same content (be `.Equal`), but should be at different memory locations (`.NotSame`).

x/marker/keeper/keeper_test.go
SpicyLemon: They almost are since there's no keeper functionality for changing them. The only way they could get changed after keeper creation would be to alter the list provided to the constructor, but that's pretty limiting.

Even still, I updated `NewKeeper` to store a deep copy of the provided slice instead of the provided slice directly. Now, the only way to change it after keeper creation would be to add functionality to the keeper.

x/marker/spec/12_transfers.md
```diff
@@ -0,0 +1,264 @@
...
+
+During such transfers several things are checked using a `SendRestrictionFn` injected into the bank module. This restriction is applied in almost all instances when funds are being moved between accounts. The exceptions are delegations, undelegations, minting, burning, and marker withdrawals.
+
+<!-- TODO: Add notes about IBC movement too -->
+
+## Definitions
+
+### Transfer Permission
+
+One permission that can be granted to an address is `transfer`. An address with `transfer` permission can utilize `MsgTransferRequest` to move restricted funds from one account to another. If the marker allows forced transfer, the source account can be any account, otherwise, it must be the admin's own account.
```
iramiller, line 29:
```suggestion
One permission that can be granted to an address is `transfer`.  The `transfer` permission is granted to accounts that represent a "Transfer Agent" or "Transfer Authority" for restricted marker tokens. An address with `transfer` permission can utilize `MsgTransferRequest` to move restricted funds from one account to another. This authority can be granted (see `authz`) to another address allowing them to use the `MsgTransferRequest` endpoint.

An additional transfer capability is the "Forced Transfer" permission.  This permission changes the nature of the token itself such that accounts no longer fully self-custody the marker token. The Transfer Agent on a "Forced Tranfer" marker has the ability to move the funds from one account to another unilaterally.  Note that the source account must be a "standard signing account".  Smart contracts or other module and protocol controlled accounts are not valid sources to force transfers from.
```

x/marker/spec/12_transfers.md
```diff
@@ -0,0 +1,264 @@
...
+
+`MsgSend` and `MsgMultiSend` can also be used by an address with `transfer` permission to move funds out of their own account.
+
+### Required Attributes
+
+Required attributes are something that can be configured for a restricted marker. Accounts with the required attributes are allowed to receive those marker's coins regardless of whether the sender has `transfer` authority.
```
iramiller, line 35:
```suggestion
Required attributes allow a marker Transfer Authority to define a set of account attestations created with the name/attribute modules to certify an account as an approved holder of the token.  Accounts that possess all of the required attributes are considered authorized by the Transfer Authority to receive the token from normal bank send operations without a specific Transfer Authority approval.

Note: Required attributes control restricted transfers and are only supported on restricted markers.
```

app/app.go
SpicyLemon: Updated.

x/marker/spec/12_transfers.md
SpicyLemon: I left off the
> This authority can be granted (see `authz`) to another address allowing them to use the `MsgTransferRequest` endpoint.

It's not really the authority that's being granted with `authz`, just the ability to use a `MsgTransferRequest`. That's just normal `authz` operation, though, so I'm not sure it needs to be called out. You can't use `authz`, though, to convey transfer permission in a way that `MsgSend` could be used. There could also be some confusion with the `MarkerTransferAuthorization` which is issued by someone with some restricted coins, but without transfer permission to someone with transfer permission. But that's discussed in `11_authorization.md`.

x/marker/spec/12_transfers.md
```diff
@@ -0,0 +1,264 @@
...
+
+```
+
+If the `Send` is allowed, the requested funds are transferred from `QFH` to `Receiver`.
+
+If the `Send` is denied, the funds remain with `QFH`.
+
+An important subtle part of this process is the rechecking of `Receiver` attributes. It's possible for the initial send to be okay (causing funds to be quarantined), then later, during this `Accept`, the send is not okay, and the quarantined funds are effectively locked with`QFH` until the `Receiver` gets the required attributes.
+
+If the marker does not have required attributes though, it's assumed that they were originally sent by someone with transfer authority, so they are allowed to continue to from here too.
```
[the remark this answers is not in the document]
SpicyLemon: I fixed the first paragraph, but left off the note.

Removing funds from the `QFH` account without updating quarantine records will break an invariant.

`authz` cannot be used like that to move funds out of the `QFH`. The only way funds can leave the `QFH` is through the quarantine `Accept` endpoint. Even if `authz` is used to allow a third party to call `Accept`, the marker restrictions only know about the source account (`QFH`) and the destination (the `toAddr`).

Funds can become stuck in the `QFH` if there are required attributes but the `toAddr` doesn't have them. The only workarounds would be to give the `toAddr` the required attributes or grant `QFH` transfer permission. The latter is probably a very bad idea.

If these stuck funds become a problem, we could either add some functionality to `MsgTransferRequest` or create a new `Rescue` endpoint somewhere. Either way, quarantine records need to be updated as when the funds are moved out of `QFH`.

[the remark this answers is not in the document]
iramiller: This brings up a good point… at the moment the authz is for the msgtransfer endpoint, but we also have attribute checks on acceptance and potentially disabled ability on source to send without a corresponding ability to express these rights in authz.  That ability would be needed here for the above as the intent wasn’t to move the funds from the quarantine unilaterally but instead to satisfy the recipient eligibility criteria.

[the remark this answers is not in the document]
SpicyLemon: The required attributes and deny-list checks are specific to the account receiving funds and account the funds are coming out of (respectively). The former can be bypassed using the transfer permission. The latter is only bypassable by using the `MsgTransferRequest` endpoint, though I wonder if even that shouldn't be allowed.

If an account does not have the required attributes, but funds should still be sent to it, there's a few ways to do it:
1. The current holder of the funds is given transfer permission. They can then use any method of sending the funds that they want.
2. The current holder of the funds is conveyed transfer permission through authz using a `GenericAuthorization` on the `MsgTransferRequest` messages. They then submit an authz `Exec` with a `MsgTransferRequest` identifying the granter as the `admin`.
3. The current holder of the funds issues a `MarkerTransferAuthorization` to someone with transfer permission who then sends a `MsgTransferRequest` on their behalf.

The deny list check isn't done during a `MsgTransferRequest`. It's only applied when the `Send` is initiated by other means (e.g. `MsgSend`). That check is done before checking for transfer authority though, so if someone is on the deny list, they only way the funds can move out of their account is through a `MsgTransferRequest`. So option 2 or 3 above would work there too.

All authz does is allow a message to have a different signer from the one contained in the message. Usually, by the time we're processing a message, that checking as already been done and we don't know the actual signer, just the "signer" identified in the msg. During a quarantine `Accept`, the signer is the `toAddr`. However, for an `Accept` to be processed, there must be a record of the quarantined funds to that address, and the funds can only go to that address. There's no way for it to know of another address that might have tried to convey transfer permission in order to allow the `Accept` to go through. During a `MsgSend`, the same is true. There's no way for a third address to be involved conveying transfer permission.

Currently, the only way to get a third address involved like that is through `MsgTransferRequest`. But that can't be used for quarantined funds because `QFH` is a module account (plus, doing so would break an invariant). If we don't mind the marker module having a dependency on the quarantine module, we can add a check in `MsgTransferRequest` to identify when the `from` is `QFH` and properly tell the quarantine module about the fund movement. Otherwise, somewhere, we'd have to create a new `Rescue` endpoint that would provide the ability to get stuck funds out of quarantine, applying the appropriate restrictions and properly updating the quarantine module.

---

Should we be doing the deny-list check during a `MsgTransferRequest`? I feel like we should since transfer permission does not bypass that check during other sends. Or maybe transfer permission should bypass the deny-list check during other sends too. But I feel like the spirit of the deny-list is the same as a hold; i.e. those funds cannot leave the account until they're off the deny list.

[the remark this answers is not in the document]
iramiller: No, we should not be restricting `MsgTransferRequest` as it is more correctly `MsgTransfer[Authority]Request` meaning you must be a holder of the `transfer` grant (directly or indirectly) and you have complete authority to perform transfers.  Literally the Transfer Authority sets the rules and can make exceptions to them.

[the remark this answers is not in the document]
SpicyLemon: Summarizing an outside conversation in here for the record.

Let's say `chickens` are a restricted coin, and I have a bunch of them. I have transfer permission for them, but am also on the deny list for them.
* If I `MsgSend` from: me, to: you, amount: `5chickens`, it will fail because I'm on the deny list.
* If I `MsgTransferRequest` from: me, to: you, admin: me, amount: 5chickens, it will go through because it doesn't check the deny list.

This is expected behavior because the `MsgTransferRequest` is sort of a `sudo bank send`. Someone with transfer but on the deny list needs to recognize that they're possibly bypassing some restrictions to do their transfer.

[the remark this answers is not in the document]
SpicyLemon: As for the issue with funds getting stuck in quarantine, I feel like that would be better handled with a separate issue and PR since there will be some design and implementation decisions to make.

iramiller:
This is a very good PR.  The documentation updates in `12_transfers.md` are exceptionally useful for understanding the myriad of options that the marker module offers for restricted tokens.

x/marker/types/immutable_addresses.go
```diff
@@ -0,0 +1,48 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// ImmutableAccAddresses holds some AccAddresses in a way that they cannot be changed after creation.
+type ImmutableAccAddresses struct {
```
iramiller, line 6: A bit of overkill as it adds a bit of code that fees like it should be useful as part of the base SDK directly into a module ... but still a worthwhile change. (feel free to dismiss)

iramiller (approved):

nullpointer0x00 (approved):

x/marker/spec/12_transfers.md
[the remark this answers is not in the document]
SpicyLemon: I created #1654 for the stuck-in-quarantine thing.

x/marker/types/immutable_addresses.go
SpicyLemon: Yeah... I mostly just wanted to see what it'd look like, but kind of liked it.

It has the nice benefit of turning `IsReqAttrBypassAddr` from O(7) into O(1), which probably actually doesn't even matter in the long run, but oh well.

Issue #1658, opened 2023-08-22
Epic: Exchange Module


SpicyLemon:
## Summary

Build an `x/exchange` module to facilitate the buying and selling of assets.

## Problem Definition

Currently, exchanges of assets can only be done with either a smart contract or the groups module.

Each of those has downsides, though. Notably, the assets are sent to an intermediary account while waiting for finalization/clearing/settlement.

## Proposal

Create an `x/exchange` module that utilizes in-place holds and allows owners to put their assets up for sale, and for buyers to express desire to buy.

The module will allow creation of markets that are responsible for match making (among other things). Users can then submit asks and bids using their assets, which are placed in hold until settled or cancelled.

The markets will be able to collect order submission fees and settlement fees, and the exchange (blockchain) will receive a small cut of that.

## Sub-tasks

- [x] #1698
- [ ] #1699
- [ ] #1700
- [ ] #1701
- [ ] #1702
- [ ] #1703

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Closed since this shipped with `v1.17.0`, even though there's still some stuff to add later.

PR #1661, opened 2023-08-23
Create the exchange module

SpicyLemon:
## Description

closes: #1698

related: #1658 

This PR creates the exchange module and adds it to the app.

---

write a little note why.


SpicyLemon:
I'm dropping this into ready-for-review.

I've created issues to fill in the remaining `TODO`s. Those issues can be found [under sub-tasks](https://github.com/provenance-io/provenance/issues/1658).

I haven't yet done the code review myself, though (which I usually do before dropping it into read-for-review). I'm expecting cleanup of unused/unneeded stuff to happen with the work for #1699.

SpicyLemon:
diff analysis:
```
Line Changes -     Code:  +7577     -56
Line Changes -    Tests: +22979      -1
Line Changes -    Proto:  +1073      -0
Line Changes - Markdown:  +1572      -0
Line Changes -     Auto: +23884      -0
========================================
Line Changes -    Total: +57085     -57
```

iramiller:
@jtalis this is the upcoming exchange module... when this merges we will have it available in dev

x/exchange/params.go
```diff
@@ -0,0 +1,66 @@
...
+	"errors"
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+const (
+	// DefaultDefaultSplit is the default value used for the DefaultSplit parameter.
+	// TODO[1658]: Discuss what this should be with someone who would know.
+	DefaultDefaultSplit = uint32(500)
```
SpicyLemon, line 13: Anyone have any thoughts on this value?

proto/provenance/exchange/v1/orders.proto
```diff
@@ -0,0 +1,72 @@
...
+import "gogoproto/gogo.proto";
+
+// Order associates an order id with one of the order types.
+message Order {
+  option (gogoproto.goproto_getters) = false;
+
+  // order_id is the numerical identifier for this order.
+  uint64 order_id = 1;
+  // order is the specifics of this order.
+  oneof order {
```
Taztingo, line 20: I wonder if we should try to stick with Any based on this documentation from cosmos-sdk.

https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-019-protobuf-state-encoding.md#usage-of-any-to-encode-interfaces

proto/provenance/exchange/v1/orders.proto
```diff
@@ -0,0 +1,72 @@
...
+  // proportional fee (and possibly the settlement flat fee) is taken out of the amount the seller receives,
+  // so it's possible that the seller will still receive less than this price.
+  cosmos.base.v1beta1.Coin price = 4 [(gogoproto.nullable) = false];
+  // seller_settlement_flat_fee is the flat fee for sellers that will be charged during settlement. If this denom is the
+  // same denom as the price, it will come out of the actual price received. If this denom is different, the amount must
+  // be in the seller's account and a hold is placed on it until the order is filled or cancelled.
+  cosmos.base.v1beta1.Coin seller_settlement_flat_fee = 5;
+  // allow_partial should be true if partial fulfillment of this order should be allowed, and should be false if the
+  // order must be either filled in full or not filled at all.
+  bool allow_partial = 6;
```
Taztingo, line 49: How do partials actually work?

proto/provenance/exchange/v1/orders.proto
```diff
@@ -0,0 +1,72 @@
...
+  // price is the amount that the buyer will pay for the assets.
+  // A hold is placed on this until the order is filled or cancelled.
+  cosmos.base.v1beta1.Coin price = 4 [(gogoproto.nullable) = false];
+  // buyer_settlement_fees are the fees (both flat and proportional) that the buyer will pay (in addition to the price)
+  // when the order is settled. A hold is placed on this until the order is filled or cancelled.
+  repeated cosmos.base.v1beta1.Coin buyer_settlement_fees = 5
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+  // allow_partial should be true if partial fulfillment of this order should be allowed, and should be false if the
+  // order must be either filled in full or not filled at all.
+  bool allow_partial = 6;
```
Taztingo, line 71: Same as above.

proto/provenance/exchange/v1/params.proto
```diff
@@ -0,0 +1,28 @@
...
+import "gogoproto/gogo.proto";
+
+// Params is a representation of the exchange module parameters.
+message Params {
+  // default_split is the default proportion of fees the exchange receives in basis points.
+  // It is used if there isn't an applicable denom-specific split defined.
+  // E.g. 100 = 1%. Min = 0, Max = 10000.
+  uint32 default_split = 1;
+  // denom_splits are the denom-specific amounts the exchange receives.
+  repeated DenomSplit denom_splits = 2 [(gogoproto.nullable) = false];
```
Taztingo, line 18: How does this interact with the fees in the other protos?

x/exchange/module/module.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/provenance-io/provenance/x/exchange/simulation"
+)
+
+var (
+	_ module.AppModule           = AppModule{}
+	_ module.AppModuleBasic      = AppModuleBasic{}
+	_ module.AppModuleSimulation = AppModule{}
+)
+
+type AppModule struct {
```
Taztingo, line 33: This should be near the AppModule methods.

x/exchange/module/module.go
```diff
@@ -0,0 +1,156 @@
...
+	AppModuleBasic
+	keeper keeper.Keeper
+}
+
+func NewAppModule(cdc codec.Codec, exchangeKeeper keeper.Keeper) AppModule {
```
Taztingo, line 38: This should also be with the AppModule methods.

proto/provenance/exchange/v1/orders.proto
SpicyLemon: My problem with using an `any` here is that a codec is then required in order to deal with an `Order` object. Since the codec is an app-level thing that's only available in the keepers, it's significantly more difficult to keep all this logic in this package. Plus, I'm not trying to have this hold an interface that any message can then implement, I only want these two specific types possible in here.

Further, this `Order` message is only used in query results and genesis. Its not used in any tx, signing, or state encoding. In genesis, they'll be json which will just omit the unused type(s). In query results, the chain just uses standard protobuf encoding for the result and clients can use standard protobuf decoding. It doesn't suffer from the same problems discussed in that link (and the issue it references) because all of the possible types are known ahead of time (right now). This isn't designed to be extensible outside of this module. I.e., it's not like a `MsgSubmitProposal` where the content is expected to be defined in a different module/repo/organization.

All that being said, after looking at the `UnpackInterfaces` thing, I might be overblowing the codec complications. And a `oneof` field has a forward compatibility consideration if we ever want to add a type. I.e. If a type is added to this `oneof`, any client using the old proto won't be able to handle results containing the new type. But as an `Any`, it should still be able to decode the content, (just maybe not know what to do with it).

Let me think on this and look into the amount of work that'd be needed to change it.

iramiller: the forward compatibility piece seems like a big deal.. I am suspecting that we already have new order types in the pipeline for the peer to peer use cases...

proto/provenance/exchange/v1/params.proto
SpicyLemon: This is applied to fees that markets receive in order to determine how much the exchange should receive. For each coin in the fee, if there's a `denom_splits` entry with that coin's denom, that split value is applied to the coin; otherwise, (if there aren't any `denom_splits` with the coin's denom), the `default_split` value is applied to the coin. Fractional amounts are rounded up to the next whole number.

### Example

Say the default split is 500 (i.e. 5%), and there's a denom split for `nhash` that's 777 (i.e. 7.77%), and a market is collecting a fee of `444444nhash,1000usdf`.

* `nhash` has a specific denom split of 777.
  444,444 * 777 / 10,000 = 34,533, and 444,444 * 777 % 10,000 = 2,988, so we round up and the exchange will get `34534nhash` (of the `444444nhash`).
* `usdf` doesn't have a specific denom split, so the default 500 is used. 
  1,000 * 500 / 10,000 = 50, and 1,000 * 500 % 10,000 = 0, so we don't need to round up, and the exchange will get `50usdf` (of the `1000usdf`).

When fees are collected, the full fee amount is sent to the market account, then the exchange's cut is transferred from the market to the exchange (fee collector).

So, `444444nhash,1000usdf` is transferred from the buyers/sellers to the market, and then `34534nhash,50usdf` is then transferred from the market account to the fee collector, leaving the market with `409910nhash,950usdf`.

x/exchange/module/module.go
SpicyLemon: Good call. Moved.

x/exchange/module/module.go
SpicyLemon: Moved this too.

proto/provenance/exchange/v1/query.proto
```diff
@@ -0,0 +1,295 @@
...
+  rpc QueryValidateManageFees(QueryValidateManageFeesRequest) returns (QueryValidateManageFeesResponse) {
+    option (google.api.http).get = "/provenance/exchange/v1/validate/manage_fees";
+  }
+}
+
+// QueryOrderFeeCalcRequest is a request message for the QueryOrderFeeCalc endpoint.
+// Exactly one of ask_order or bid_order must be provided.
+message QueryOrderFeeCalcRequest {
+  // ask_order is the ask order to calculate the fees for.
+  AskOrder ask_order = 2;
```
Taztingo, line 92: Would it be easier for this message to contain a Order message? That would enforce the one of these types.

proto/provenance/exchange/v1/market.proto
```diff
@@ -0,0 +1,143 @@
...
+  // An entry of exactly "*" will match any attribute, which is equivalent to leaving this list empty.
+  repeated string req_attr_create_ask = 12;
+  // req_attr_create_ask is a list of attributes required on an account for it to be allowed to create a bid order.
+  // An account must have all of these attributes in order to create a bid order in this market.
+  // If the list is empty, any account can create bid orders in this market.
+  //
+  // An entry that starts with "*." will match any attributes that end with the rest of it.
+  // E.g. "*.b.a" will match all of "c.b.a", "x.b.a", and "e.d.c.b.a"; but not "b.a", "xb.a", "c.b.x.a", or "c.b.a.x".
+  //
+  // An entry of exactly "*" will match any attribute, which is equivalent to leaving this list empty.
```
nullpointer0x00, line 105: Not a big deal, but why would we want to allow `*` and just not have an empty list as the option for `any`?  It would be consistent with Marker's required attributes then.

proto/provenance/exchange/v1/events.proto
```diff
@@ -0,0 +1,135 @@
...
+// EventMarketEnabled is an event emitted when a market is enabled.
+message EventMarketEnabled {
+  // market_id is the numerical identifier of the market.
+  uint32 market_id = 1;
+  // updated_by is the account that enabled the market.
+  string updated_by = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+}
+
+// EventMarketDisabled is an event emitted when a market is disabled.
+message EventMarketDisabled {
```
Taztingo, line 83: We might be able to reduce the amount of events by changing the name to encompass both cases. The following is just an example:

```
message EventMarketStatusChange {
  // market_id is the numerical identifier of the market.
  uint32 market_id = 1;
  // updated_by is the account that enabled the market.
  string updated_by = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  // enabled is the status of the Market
  string enabled = 3 [(cosmos_proto.scalar) = "cosmos.AddressString"];
}
```

proto/provenance/exchange/v1/market.proto
Taztingo: I actually had the same thought, and it might simplify things if there is only one of these inputs.

SpicyLemon: That thought occurred to me too. I didn't realize markers didn't allow simply `*` though. I'll take this functionality out of here.

x/exchange/events.go
```diff
@@ -0,0 +1,131 @@
...
+func NewEventMarketDisabled(marketID uint32, updatedBy sdk.AccAddress) *EventMarketDisabled {
+	return &EventMarketDisabled{
+		MarketId:  marketID,
+		UpdatedBy: updatedBy.String(),
+	}
+}
+
+// NewEventMarketUserSettleUpdated returns a new EventMarketUserSettleEnabled if isAllowed == true,
+// or a new EventMarketUserSettleDisabled if isActive == false.
+func NewEventMarketUserSettleUpdated(marketID uint32, updatedBy sdk.AccAddress, isAllowed bool) proto.Message {
```
Taztingo, line 82: If you squash this event into one, then you will only need 1 constructor.

x/exchange/events.go
```diff
@@ -0,0 +1,131 @@
...
+func NewEventMarketDetailsUpdated(marketID uint32, updatedBy sdk.AccAddress) *EventMarketDetailsUpdated {
+	return &EventMarketDetailsUpdated{
+		MarketId:  marketID,
+		UpdatedBy: updatedBy.String(),
+	}
+}
+
+// NewEventMarketActiveUpdated returns a new EventMarketEnabled if isActive == true,
+// or a new EventMarketDisabled if isActive == false.
+func NewEventMarketActiveUpdated(marketID uint32, updatedBy sdk.AccAddress, isActive bool) proto.Message {
```
Taztingo, line 59: If you squash this event into one, then you will only need 1 constructor.

proto/provenance/exchange/v1/events.proto
SpicyLemon: I waffled on this quite a bit. I ultimately decided to go with the separate events because it puts the most important piece of information (whether it was enabled or disabled) in the event type which is more prominent than in an attribute. I also liked that, as separate event types, the events are a little smaller. These events probably won't be emitted often enough for the size differences to matter all that much, but it's still something to consider.

In general, you want events to be as small as possible. Usually, you don't want to include things (as attributes) that can be found with a query. You generally want the bare minimum amount of data that would allow you to a) recognize that something happened and b) be able to query for the details if you need them. With that in mind, I didn't really want a `bool enabled` in the event since you can query for the market to get its value. But I still really wanted the event to convey whether it was enabled or disabled, so I went with separate events whose names convey that information.

I also originally didn't have the `updated_by` field in there. Technically, it'd be possible to query for the info by getting the block txs, finding the right msg, and getting the field from it, but I decided that was a little too cumbersome. Plus, I imagine we (or someone) will need that `updated_by` information for an audit eventually.

But it's also nice to put things into events that aren't as easy to query for. Originally, the `EventOrderFilled` only had the order id. But for ask orders, the price received and fees paid depend on the bid(s) it was settled with and isn't something that can be actually queried. The actual price received might be split across multiple events generated by the bank module (since it might be coming from multiple accounts). So I felt it prudent to include those two fields in that event. I then added the `assets` field to it since the other info was there anyway, and now, something can use just those events to keep track of what things are being traded for.

iramiller: Events are in block results which is straight forward to process without needing to understand any of the proto msg formats or even run the "provenance app" (reading straight from tendermint/comet).

In my view you want events to be aggregated to some extent based on what you expect the listener to filter on ... a category of use.  It can be fine to add differential in json as empty/null attributes are omitted which doesn't cause excessive state bloat.  

Forcing related data to be queried breaks this model as then you need to query the "app" side of the blockchain process and you have associated issues with data retention (pruning intervals).  Relying on queries is a very bad place to be if you are event processing.

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+	k.emitEvent(ctx, exchange.NewEventMarketDetailsUpdated(marketID, updatedBy))
+
+	return nil
+}
+
+// storeMarket writes all the market fields to the state store (except MarketDetails which are in the account).
+func storeMarket(store sdk.KVStore, market exchange.Market) {
+	marketID := market.MarketId
+	setMarketKnown(store, marketID)
+	setCreateAskFlatFees(store, marketID, market.FeeCreateAskFlat)
```
Taztingo, line 1152: Can we store some of these together? Is this being done for optimization purposes?

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+		k.accountKeeper.SetAccount(ctx, marketAcc)
+	}
+
+	storeMarket(store, market)
+}
+
+// CreateMarket saves a new market to the store with all the info provided.
+// If the marketId is zero, the next available one will be used.
+func (k Keeper) CreateMarket(ctx sdk.Context, market exchange.Market) (uint32, error) {
+	// Note: The Market is passed in by value, so any alterations to it here will be lost upon return.
```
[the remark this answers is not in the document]
SpicyLemon: It's there to prevent someone from changing the arg to a `*exchange.Market` without realizing theres stuff in there depending on it not being a reference.

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+	setCreateBidFlatFees(store, marketID, market.FeeCreateBidFlat)
+	setSellerSettlementFlatFees(store, marketID, market.FeeSellerSettlementFlat)
+	setSellerSettlementRatios(store, marketID, market.FeeSellerSettlementRatios)
+	setBuyerSettlementFlatFees(store, marketID, market.FeeBuyerSettlementFlat)
+	setBuyerSettlementRatios(store, marketID, market.FeeBuyerSettlementRatios)
+	setMarketActive(store, marketID, market.AcceptingOrders)
+	setUserSettlementAllowed(store, marketID, market.AllowUserSettlement)
+	setAccessGrants(store, marketID, market.AccessGrants)
+	setReqAttrsAsk(store, marketID, market.ReqAttrCreateAsk)
+	setReqAttrsBid(store, marketID, market.ReqAttrCreateBid)
```
Taztingo, line 1162: Can we store some of these together? Is this being done for optimization purposes? It just seems like a lot to have getters, setters, and iterators for.

x/exchange/keeper/market.go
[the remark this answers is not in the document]
Taztingo: Hm, if that's a concern then would you want to change that param to a pointer and manually make a copy of it? The code would speak for itself and we wouldn't have the concern of someone changing that param type.

x/exchange/market.go
```diff
@@ -0,0 +1,662 @@
...
+		// Nothing to check for with the AcceptingOrders and AllowUserSettlement booleans.
+		ValidateReqAttrs("create-ask", m.ReqAttrCreateAsk),
+		ValidateReqAttrs("create-bid", m.ReqAttrCreateBid),
+	)
+}
+
+// ValidateFeeOptions returns an error if any of the provide coin values is not a valid fee option.
+func ValidateFeeOptions(field string, options []sdk.Coin) error {
+	var errs []error
+	denoms := make(map[string]bool)
```
Taztingo, line 55: This is very minor, and really up to you if you want to include it. You can specify a minimum size for the map so that it potentially won't have to reallocate.

denoms := make(map[string]bool, len(options))
dups := make(map[string]bool, len(options))

x/exchange/keeper/market.go
SpicyLemon: It's done for optimization purposes so that as we're processing things, we only have to look up the pieces of data that are needed.

x/exchange/market.go
SpicyLemon: I've never really looked into the guts of how maps grow and when it's beneficial to include a length with a `make(map...)`. I've known it's possible, but haven't ever seen any discussions about when its best to include it. By contrast, I've seen lots of discussions about including the length and optional capacity when creating slices. I guess I've just assumed the make function needed to allow the lengths for slices, and therefore also allows a length for maps even if it doesn't really do anything (even though I don't actually know it doesn't really do anything).

Taztingo: Yeah, the map hashes items into an array and each of the array elements are a list of buckets. As those buckets are filled the map automatically double sthe number of buckets and redistributes the keys. I doubt we will see a noticeable difference (depending on how big options can get), but pointed it out since I saw it.

SpicyLemon: Heh. From https://go.dev/ref/spec#Making_slices_maps_and_channels

> Calling make with a map type and size hint n will create a map with initial space to hold n map elements. The precise behavior is implementation-dependent.

They say in a few places that `make(T, n)` where `T` is a `map` means "map of type T with initial space for approximately n elements"

x/exchange/msg.go
```diff
@@ -0,0 +1,464 @@
...
+
+	if _, err := sdk.AccAddressFromBech32(m.Admin); err != nil {
+		errs = append(errs, fmt.Errorf("invalid administrator %q: %w", m.Admin, err))
+	}
+
+	if m.MarketId == 0 {
+		errs = append(errs, fmt.Errorf("invalid market id: cannot be zero"))
+	}
+
+	// Nothing to validate for the AcceptingOrders field.
```
Taztingo, line 262: We can probably get rid of this comment.

x/exchange/msg.go
```diff
@@ -0,0 +1,464 @@
...
+
+	if _, err := sdk.AccAddressFromBech32(m.Admin); err != nil {
+		errs = append(errs, fmt.Errorf("invalid administrator %q: %w", m.Admin, err))
+	}
+
+	if m.MarketId == 0 {
+		errs = append(errs, fmt.Errorf("invalid market id: cannot be zero"))
+	}
+
+	// Nothing to validate for the AllowUserSettlement field.
```
Taztingo, line 283: Probably don't need this comment.

x/exchange/msg.go
SpicyLemon: It's nice to have it so that it's clear that validation on that field wasn't simply forgotten.

x/exchange/market.go
SpicyLemon: All that being said, I did provide suggested sizes where it made sense to when making maps in all this exchange stuff.

x/exchange/msg.go
SpicyLemon: Here too, this tells the next person that that field wasn't forgotten, it's just that there's nothing to do for it.

x/exchange/msg.go
nullpointer0x00: I don't think these are necessary.  The worst case is that a developer will look into it themselves and discover it isn't needed.  However, there may be a future change that could require a validation to happen to a field that wasn't required in the previously this would then cause a little confusion.

x/exchange/market.go
```diff
@@ -0,0 +1,662 @@
...
+		field += " "
+	}
+	_, err := sdk.AccAddressFromBech32(a.Address)
+	if err != nil {
+		return fmt.Errorf("invalid %saccess grant: invalid address %q: %w", field, a.Address, err)
+	}
+	if len(a.Permissions) == 0 {
+		return fmt.Errorf("invalid %saccess grant: no permissions provided for %s", field, a.Address)
+	}
+	seen := make(map[Permission]bool)
```
Taztingo: Can potentially add a size heere.

x/exchange/market.go
```diff
@@ -0,0 +1,662 @@
...
+}
+
+// ValidateAccessGrantsField returns an error if any of the provided access grants are invalid.
+// The provided field is used in error messages.
+func ValidateAccessGrantsField(field string, accessGrants []AccessGrant) error {
+	if len(field) > 0 && !strings.HasSuffix(field, " ") {
+		field += " "
+	}
+	errs := make([]error, len(accessGrants))
+	seen := make(map[string]bool)
```
Taztingo: Can potentially add a size here.

x/exchange/msg.go
```diff
@@ -0,0 +1,464 @@
...
+			toRev, ok := toRevokeByAddr[ag.Address]
+			if ok {
+				for _, perm := range ag.Permissions {
+					if toRev.Contains(perm) {
+						errs = append(errs, fmt.Errorf("address %s has both revoke and grant %q", ag.Address, perm.SimpleString()))
+					}
+				}
+			}
+		}
+	} else {
```
Taztingo, line 344: You can actually get rid of this else statement and adding an error check before this branch.

```
if !m.HasUpdates() {
  errs = append(errs, errors.New("no updates"))
}
```

x/exchange/msg.go
```diff
@@ -0,0 +1,464 @@
...
+					ag.Address))
+			}
+			toRevokeByAddr[ag.Address] = ag
+		}
+
+		if err := ValidateAccessGrantsField("to-grant", m.ToGrant); err != nil {
+			errs = append(errs, err)
+		}
+
+		for _, ag := range m.ToGrant {
```
Taztingo, line 334: There is 4 levels of indentation here. Is it possible to restructure this a little bit to make it easier to read? You could add some additional functions or rewrite some of the loop.

```
if !ok {
  continue
}
Happy path...
```

proto/provenance/exchange/v1/tx.proto
```diff
@@ -0,0 +1,367 @@
...
+message MsgMarketManagePermissionsRequest {
+  option (cosmos.msg.v1.signer) = "admin";
+
+  // admin is the account with "permissions" permission requesting this change.
+  string admin = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // market_id is the numerical identifier of the market to update required attributes for.
+  uint32 market_id = 2;
+
+  // revoke_all are addresses that should have all their permissions revoked.
+  repeated string revoke_all = 3;
```
Taztingo, line 264: Is there a difference between `revoke_all` and adding all permissions to `to_revoke`? It would get rid of a lot of extra validation and logic if we only used one.

proto/provenance/exchange/v1/tx.proto
```diff
@@ -0,0 +1,367 @@
...
+  option (cosmos.msg.v1.signer) = "admin";
+
+  // admin is the account with "attributes" permission requesting this change.
+  string admin = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // market_id is the numerical identifier of the market to update required attributes for.
+  uint32 market_id = 2;
+
+  // create_ask_to_add are the attributes that should now also be required to create an ask order.
+  repeated string create_ask_to_add = 3;
+  // create_ask_to_add are the attributes that should no longer be required to create an ask order.
```
Taztingo, line 285: Very minor typo:

create_ask_to_**remove**

proto/provenance/exchange/v1/tx.proto
```diff
@@ -0,0 +1,367 @@
...
+  repeated string create_ask_to_remove = 4;
+  // create_ask_to_add are the attributes that should now also be required to create a bid order.
```
Taztingo, line 287: Very minor typo:

create_**bid**_to_add

proto/provenance/exchange/v1/tx.proto
```diff
@@ -0,0 +1,367 @@
...
+  repeated string create_bid_to_add = 5;
+  // create_ask_to_add are the attributes that should no longer be required to create a bid order.
```
Taztingo, line 289: Very minor typo:

create_**bid**_to_**remove**

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+
+	var updatedAttrs []string
+	for _, attr := range curAttrs {
+		if !exchange.ContainsString(toRemove, attr) {
+			updatedAttrs = append(updatedAttrs, attr)
+		}
+	}
+
+	for _, attr := range toAdd {
+		if !exchange.ContainsString(curAttrs, attr) {
```
Taztingo, line 969: If you flip the logic here then you can prevent the happy path from being indented.

```
if exchange.ContainsString(curAttrs, attr) {
  errs = append(errs, fmt.Errorf("cannot add %s required attribute %q: attribute already required", field, attr))
  continue
}
updatedAttrs = append(updatedAttrs, attr)
```

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+	if marketAcc == nil {
+		return nil
+	}
+	return &marketAcc.MarketDetails
+}
+
+// UpdateMarketDetails updates a market's details. It returns an error if the market account
+// isn't found or if there aren't any changes provided.
+func (k Keeper) UpdateMarketDetails(ctx sdk.Context, marketID uint32, marketDetails exchange.MarketDetails, updatedBy sdk.AccAddress) error {
+	if err := marketDetails.Validate(); err != nil {
```
Taztingo, line 1128: It looked like this was already validated in the ValidateBasic. Does this need to be done twice?

x/exchange/keeper/market.go
nullpointer0x00: I reviewed some of the msg_server endpoints to follow the logic through.  Some of them are doing a lot OF individual reads from the store, i.e. `CreateAsk` and `CreateBid`.  I understand what @Taztingo means when asking if grouping these are better.  
Also, my other concern is maintainability, I didn't really understand the relationship of the different parts until I stepped through the keeper.  Grouping them would have made it a little easier to understand that `AskFlatFees`, `SellerSettlementFlatFees`,  `setReqAttrsAsk` are going to be used together.  Also, maybe it would reduce the amount of code/tests to maintain as well?

x/exchange/keeper/orders.go
```diff
@@ -0,0 +1,711 @@
...
+}
+
+// GetOrder gets an order. Returns nil, nil if the order does not exist.
+func (k Keeper) GetOrder(ctx sdk.Context, orderID uint64) (*exchange.Order, error) {
+	return k.getOrderFromStore(k.getStore(ctx), orderID)
+}
+
+// CreateAskOrder creates an ask order, collects the creation fee, and places all needed holds.
+func (k Keeper) CreateAskOrder(ctx sdk.Context, askOrder exchange.AskOrder, creationFee *sdk.Coin) (uint64, error) {
+	if err := askOrder.Validate(); err != nil {
```
Taztingo, line 567: Does this need to be validated again? I think it's done in ValidateBasic.

x/exchange/keeper/market.go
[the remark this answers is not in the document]
SpicyLemon: No. That's a whole lot of extra code just to get rid of a one line comment. It'd also be significantly less efficient than the automatic way that go does it. The code actually does speak to itself as long as you're familiar with some golang intricacies. This comment is a defensive measure to hopefully prevent someone less familiar with golang from unknowingly making a mistake.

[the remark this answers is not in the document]
nullpointer0x00: I usually just put those things into tests.  If a developer changes something and it breaks then mistake is apparent.

x/exchange/keeper/market.go
SpicyLemon: Even though creating an ask will require lookup of required create-ask attributes, and create-ask flat fees, we only need to check one fee denom (the one provided). Yes, there are more lookups, but the amount of data looked up is significantlly less than if things were bundled together.

We have a problem with markers and how they're starting to have too much data in them. E.g. in order to check the permissions of one address, you have to read all of the marker's permissions for all addresses first. We've already had to address a similar problem in the SDK with the send-enabled list.

So yes, it is an extra burden too build, but it's a necessary burden. As for maintenance, it's pretty simple: Each aspect has its own getters/setters; when you need to get something, you get just that thing, when you need to update something, you update just that thing. There isn't any duplicated logic code either. So when a thing needs changing, you can change just that thing.

Since I haven't written the unit tests on most of the keeper stuff, I also haven't audited what's there vs what's actually being used. So there's a good chance I'll be able to clean up some of that stuff, but It still needs to be broken out like this.

x/exchange/msg.go
SpicyLemon: Why is it bad to have it here? Why is it bad to save a developer the trouble of re-investigating something?
If there's a future change that could require validation to happen, they'd be looking at stuff related to the field anyway, knowing there's a change and would know immediately that the comment was no longer accurate.

x/exchange/msg.go
SpicyLemon: I thought about that, but wanted only the one return at the end of the function.

Taztingo: That makes sense. I was just pointing it out because it goes against the [go convention](https://go.dev/doc/effective_go).

x/exchange/msg.go
SpicyLemon: I get the desire to not get too tabbed in, but with every best practice, there are exceptions. That short-circuit adds 2 lines and saves a tab on 5 lines. My cutoff is usually around 8-10 tabbed-in lines for doing that. I feel like having only a few lines tabbed in is fine, specially when they're short, and specially when the alternative is either a clunky function extraction or a loop control statement. The one line tabbed in the most is the longest, but it's clear in the first 20 chars that it's just adding an error to the list which is usually all someone needs to know about that line.

Ultimately, I don't think the extra tabs hurt readability here.

proto/provenance/exchange/v1/tx.proto
SpicyLemon: In `to_revoke` you can only include permissions that they have. So you'd have to look up their permissions before you could revoke them. Most of the time, when permissions need to be revoked, it needs to be all their permissions, and its often time critical. So it's good to make it easy to do that.

If I got rid of the `to_revoke` list (and just had `revoke_all` and `to_grant`), it becomes painful to revoke just a single permission since, again, you'd have to look up their existing permissions first.

x/exchange/keeper/market.go
```diff
@@ -0,0 +1,1303 @@
...
+func (k Keeper) UpdatePermissions(ctx sdk.Context, msg *exchange.MsgMarketManagePermissionsRequest) error {
+	admin := sdk.MustAccAddressFromBech32(msg.Admin)
+	marketID := msg.MarketId
+	store := k.getStore(ctx)
+	var errs []error
+
+	for _, addrStr := range msg.RevokeAll {
+		addr := sdk.MustAccAddressFromBech32(addrStr)
+		perms := getUserPermissions(store, marketID, addr)
+		if len(perms) > 0 {
```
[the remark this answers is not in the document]
SpicyLemon: Loop control statements like `continue` and `break` make reading and understanding code more difficult. I'm not saying they should never be used, heck, I even know of valid uses of `goto`, but it's good to try to avoid them.

That being said, I did flatten this a bit:
```
		if len(perms) == 0 {
			errs = append(errs, fmt.Errorf("account %s does not have any permissions for market %d", addrStr, marketID))
		}
		if len(errs) == 0 {
			revokeUserPermissions(store, marketID, addr)
		}
```

x/exchange/keeper/market.go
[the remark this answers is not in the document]
Taztingo: I guess I'm not really understanding the problem because it should be the same amount of lines just handling both our concerns. I'm pretty sure go does pass by value and doesn't do a deep copy, which would be equivalent to this... We don't need the comment, and we aren't concerned about the user changing the param type.

```
func (k Keeper) CreateMarket(ctx sdk.Context, *m exchange.Market) (uint32, error) {
	market := *m
```

x/exchange/keeper/market.go
SpicyLemon: I much prefer an else to a continue.

x/exchange/msg.go
```diff
@@ -0,0 +1,464 @@
...
+
+	if _, err := sdk.AccAddressFromBech32(m.Admin); err != nil {
+		errs = append(errs, fmt.Errorf("invalid administrator %q: %w", m.Admin, err))
+	}
+
+	if m.MarketId == 0 {
+		errs = append(errs, fmt.Errorf("invalid market id: cannot be zero"))
+	}
+
+	if m.HasUpdates() {
```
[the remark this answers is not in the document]
iramiller: Returning early means you don't have to parse/scan the rest of the method and is a major reason why the go convention of return on error is used.

This is bordering on bike shedding at this point and I fear we are getting bogged down in details that are not helpful ...

x/exchange/keeper/market.go
SpicyLemon: This is a public keeper function. So if someone is trying to do things with the exchange in some other module/place, they might be using this, in which case, we do need to make sure the market details are valid before we save them. Otherwise, someone might accidentally (or purposefully) put a giant amount of data in there and cause problems.

Honestly, I'm on the fence about most of the public keeper functions for this module. I could probably clean up a lot of stuff if we didn't want to support other modules that want to interact with the exchange. I'm pretty sure there's several public keeper funcs (e.g. `GetAccessGrants`) that aren't called anywhere, but exist because we usually have such public funcs in keepers in case something else needs access to that info.

Once I dig into the keeper unit tests, I'll have a better grasp of how I feel about those.

x/exchange/keeper/orders.go
SpicyLemon: Yup. As a public keeper function, other modules/stuff might try to use this function, and we don't want anyone to be able to create an invalid order.

x/exchange/keeper/market.go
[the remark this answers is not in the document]
SpicyLemon: This is exactly what comments are made for. It's explaining something that might not be obvious.

Everywhere that calls `CreateMarket` already has it as a value (not a reference), so I'd then have to update those calls to send in the reference to it. And then in here, I'd have to make sure it's not nil before doing anything with it, and also need an extra unit tests for that case, and I'd still have to have the unit tests make sure it's not being changed.

This one comment saves lots of code.

When I write the unit tests on it, I'll probably add a check that it's not being changed, and at that time, I'll consider removing this comment. I'm expecting questions about, "Why are you making sure this doesn't change when you're passing it in by value?" though, since I've gotten things like that in the past when I unit tested similar situations.

x/exchange/msg.go
SpicyLemon: Go convention doesn't say that you should use early returns and loop control statements, it just says that if you do, put them first and leave the `else` clause off. In this case, I don't want an early return and there isn't a loop control statement. I only want the one return at the end so that it's clear that everything is validated even if there's an already known problem.

I could put a label before the return and `if !m.HasUpdates() { errs = ...;  goto ...; }`, but I imagine there'd be more complaints about that than this.

x/exchange/msg.go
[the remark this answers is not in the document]
SpicyLemon: From a usability standpoint, it's great when I get a list of all errors rather than just the first error encountered. When you get just one error, you fix it, try again, then get another, and you get bogged down in a tedious fix+try loop. It's like when unit tests stop at the first broken test and don't run any other tests; you're only getting part of the picture. That's why I did it like this for all the message validation, and in a few other places where an error from one thing doesn't mean the next thing is impossible to check, and where painting a better picture would be helpful.

Early returns are good to use when an error is encountered that makes it impossible to move forward. E.g., if `msg.Validate()` returns an error, there's probably nothing else that can be done, so return that error. But inside `Validate()`, it's significantly more user friendly to identify as many issues as possible. There'll be cases where an error from one field means you don't need to check another, but most of the time, that's not the case.

x/exchange/msg.go
Taztingo: The main reason why I suggested getting rid of this comment is because it's self explanatory. This is the value we are updating and there are only two acceptable values. The only subset of this is one value, and if that were the case the boolean would not be included. The message would either be titled with Enable or Disable. I think comments are great, but I think we get the idea from the message and its contents.

x/exchange/keeper/orders.go
```diff
@@ -0,0 +1,711 @@
...
+		return 0, err
+	}
+
+	k.emitEvent(ctx, exchange.NewEventOrderCreated(order))
+	return orderID, nil
+}
+
+// CreateBidOrder creates a bid order, collects the creation fee, and places all needed holds.
+func (k Keeper) CreateBidOrder(ctx sdk.Context, bidOrder exchange.BidOrder, creationFee *sdk.Coin) (uint64, error) {
+	if err := bidOrder.Validate(); err != nil {
```
Taztingo, line 611: Same as previous keeper validates.

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	if !isUserSettlementAllowed(store, marketID) {
+		return fmt.Errorf("market %d does not allow user settlement", marketID)
+	}
+	return nil
+}
+
+// FillBids settles one or more bid orders for a seller.
+func (k Keeper) FillBids(ctx sdk.Context, msg *exchange.MsgFillBidsRequest) error {
+	if err := msg.ValidateBasic(); err != nil {
```
Taztingo, line 38: Same as the other keeper validation comments.

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+			if !known {
+				j = len(feeInputs)
+				feeAddrIndex[buyer] = j
+				feeInputs = append(feeInputs, banktypes.Input{Address: buyer})
+			}
+			feeInputs[j].Coins = feeInputs[j].Coins.Add(buyerSettlementFees...)
+		}
+
+		filledOrders = append(filledOrders, exchange.NewFilledOrder(order, price, buyerSettlementFees))
+	}
```
Taztingo, line 112: This whole block seems pretty self contained and you can probably extract it out to make FillBids shorter and easier to understand.

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	totalAssets, totalPrice := sumAssetsAndPrice(orders)
+
+	if !exchange.CoinsEquals(totalAssets, msg.TotalAssets) {
+		return fmt.Errorf("total assets %q does not equal sum of bid order assets %q", msg.TotalAssets, totalAssets)
+	}
```
Taztingo, line 122: Would it make sense to do this before getting the filled orders?

x/exchange/keeper/keeper.go
```diff
@@ -0,0 +1,270 @@
...
+func (k Keeper) iterate(ctx sdk.Context, pre []byte, cb func(key, value []byte) bool) {
+	iterate(k.getStore(ctx), pre, cb)
+}
+
+// DoTransfer facilitates a transfer of things using the bank module.
+func (k Keeper) DoTransfer(ctxIn sdk.Context, inputs []banktypes.Input, outputs []banktypes.Output) error {
+	// We bypass the quarantine module here under the assumption that someone creating
+	// an order counts as acceptance of the stuff to receive (that they defined when creating the order).
+	ctx := quarantine.WithBypass(ctxIn)
+	if len(inputs) == 1 && len(outputs) == 1 {
```
Taztingo, line 163: You can get rid of this indentation by flipping the logic of this clause.
```
if len(inputs) != 1 || len(outputs) != 1 {
  return k.bankKeeper.InputOutputCoins(ctx, inputs, outputs)
}
Happy path...
```

x/exchange/keeper/keeper.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/cosmos/cosmos-sdk/x/quarantine"
+
+	"github.com/provenance-io/provenance/x/exchange"
+)
+
+var (
+	// OneInt is an sdkmath.Int of 1.
+	OneInt = sdkmath.NewInt(1)
+	// TenKInt is an sdkmath.Int of 10,000.
+	TenKInt = sdkmath.NewInt(10_000)
```
Taztingo, line 27: Is there a stronger name that we can use? I'm asking because TenKInt is not really that much different than just using sdkmath.NewInt(10_000), and I had trouble deciphering what was going on at first:

`splitAmt, splitRem := exchange.QuoRemInt(coin.Amount.Mul(sdkmath.NewInt(split)), TenKInt)`

x/exchange/keeper/keeper.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// CollectFees will transfer the inputs to the market account,
+// then the exchange's cut from the market to the fee collector.
+// If there is only one input, CollectFee is used.
+func (k Keeper) CollectFees(ctx sdk.Context, marketID uint32, inputs []banktypes.Input) error {
+	if len(inputs) == 0 {
+		return nil
+	}
+	if len(inputs) == 1 {
```
Taztingo, line 239: Is the only difference here the type of events? It might make it easier for listeners if they are guaranteed to get the same type of events, and it would be nice to remove this if it's redundant.

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+		events[i] = exchange.NewEventOrderFilled(order)
+	}
+
+	k.emitEvents(ctx, events)
+	return nil
+}
+
+// FillAsks settles one or more ask orders for a buyer.
+func (k Keeper) FillAsks(ctx sdk.Context, msg *exchange.MsgFillAsksRequest) error {
+	if err := msg.ValidateBasic(); err != nil {
```
Taztingo, line 166: Same comment on validation.

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+
+	events := make([]proto.Message, len(orders))
+	for i, order := range filledOrders {
+		deleteAndDeIndexOrder(store, *order.GetOriginalOrder())
+		events[i] = exchange.NewEventOrderFilled(order)
+	}
+
+	k.emitEvents(ctx, events)
+	return nil
+}
```
Taztingo, line 293: A lot of the logic here is very similar to what is in FillBids. Is there any logic that they can share just to reduce the amount of code?

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+		if rerr != nil {
+			errs = append(errs, fmt.Errorf("error calculating seller settlement ratio fee for order %d: %w",
+				order.OrderId, rerr))
+			continue
+		}
+		if err := k.releaseHoldOnOrder(ctx, order); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
```
Taztingo, line 86: This was one of the few differences from FillAsks.

```
if sellerSettlementFlatFee != nil && !sellerSettlementFlatFee.IsZero() {
			totalSellerFee = totalSellerFee.Add(*sellerSettlementFlatFee)
		}
```

Just wanted to check there is no settlementFlatFee for the buyer?

x/exchange/keeper/fulfillment.go
```diff
@@ -0,0 +1,375 @@
...
+		filledOrders = append(filledOrders, exchange.NewFilledOrder(order, price, totalSellerFee))
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	totalAssets, totalPrice := sumAssetsAndPrice(orders)
```
Taztingo, line 252: Same as FillBids, does it make sense to move this up before filledOrders since its validation.

x/exchange/msg.go
```diff
@@ -0,0 +1,485 @@
...
+	if err := ValidateOrderIDs("bid", m.BidOrderIds); err != nil {
+		errs = append(errs, err)
+	}
+
+	inBoth := IntersectionUint64(m.AskOrderIds, m.BidOrderIds)
+	if len(inBoth) > 0 {
+		errs = append(errs, fmt.Errorf("order ids duplicated as both bid and ask: %v", inBoth))
+	}
+
+	// Nothing to validate now for the ExpectPartial flag.
```
Taztingo, line 190: Probably don't need this.

x/exchange/keeper/keeper.go
SpicyLemon: Yes, it's events. And using `InputOutputCoins` for a 1-input 1-output transfer has caused confusion in the past, mostly because `MultiSend` (or `InputOutputCoins`) is almost never used. People are more used to the `SendCoins` events, and I expect the norm here to be 1-to-1. If I ever update `InputOutputCoins` to again allow many to many, it'll actually clean up a whole lot of logic, and I'd always use it since there'd always be at least two inputs and outputs; the tricky thing there is the `SendRestriction`s.

x/exchange/keeper/fulfillment.go
SpicyLemon: Again, this is a public keeper function. Anyone else calling this shouldn't be allowed to use an invalid message.

I'm really leaning towards not having many public keeper functions (unlike most keepers). I'd make that change with the keeper unit tests PR.

x/exchange/keeper/fulfillment.go
SpicyLemon: Possibly, but a lot of it is similar but different, e.g. what's an input and what's an output, or what type of ratio fee to use. It'd end up taking quite a bit of scaffolding to generalize.

x/exchange/keeper/fulfillment.go
SpicyLemon: For the most part, you should only send funds out of an account if they've signed a TX acknowledging/accepting the loss of those funds.

The `AskOrder` struct has the `SellerSettlementFlatFee *sdk.Coin` field which is only for the flat fee. And the `BidOrder` struct has `BuyerSettlementFees sdk.Coins` which is for both the flat and ratio.

When a `BidOrder` is settled, the buyer pays their price plus the buyer settlement fees, and they receive their assets. When an `AskOrder` is settled, the seller pays the assets and receives a price (that's at least their ask price). They then pay the seller settlement fees which can come out of the price they just received.

The seller settlement flat fee can be known ahead of time, but the seller settlement ratio fee is based on the bid(s) used to settle it, so it's not known until settlement (or `FillAsks`). The seller settlement flat fee might be in a denom different from the price, but the seller ratios have to have the same price and fee denom, so the seller settlement ratio fee will always be the same denom as the price they receive. This allows us to have the sellers pay that fee out of the price that they just received. It's a gray area since they didn't actually sign a tx to have a specific fee amount removed from their account, but since they also just received the funds and (hopefully) know ahead of time that a bit of fee will come out of the price they receive.

This setup also allows someone with only `2cows` in their account (and no `chickens`) to sell them for `50chickens` while still paying the required fees for that trade. But the buyer would have to have the `50chickens` plus the required fee (flat + ratio) in their account since it's highly likely that whatever fees they're paying cannot come out of the assets they're receiving.

The `FillAsks` endpoint is similar to creating a `BidOrder` and having it immediately settled by a set of ask orders. The buyer settlement fees are paid on top of the price they're paying. So all of it must be provided in the tx, and we just need to verify that they've provided enough fee based on the total price they're paying. Each of the sellers will then get their price and pay their fees out of that.

The `FillBids` endpoint is similar to creating an `AskOrder` and having it immediately settled by a set of bid orders. Even though, with a `FillBids`, we do actually know the total settlement price, the seller ratio fee still comes out of the price they're receiving, so it's not necessary that they be indicated in the tx. So, for parity with `CreateAsk`, only the flat fee needs to be provided with a `FillBids`.

Lastly, it's possible for the seller settlement flat fee to be a denom different from the price they're receiving. When such an `AskOrder` is created, the seller must also have that flat fee in their account already, and a hold is placed on it. But if the seller settlement flat fee is the same denom as the price, we don't place a hold on it so that the `chickens` example above is still possible. During `CreateAsk`, the order is rejected if the flat fee denom is the same as the price, and flat fee + ratio fee at the ask price is more than the ask price. E.g. if the seller flat fee is `10nhash` and their price is `12nhash`, and the ratio fee ends up being `3nhash`, settlement would result in the seller receiving `12nhash`, then having to pay `13nhash` for the pleasure. So we prevent that.

x/exchange/keeper/fulfillment.go
SpicyLemon: Yup. I think I put it down there because for the validation, only one of them is needed, and the other isn't needed until later, but that's a good piece of validation to put up front.

x/exchange/keeper/fulfillment.go
SpicyLemon: Also, I wrote `FillAsks` and `FillBids` before `SettleOrders` (and all the stuff needed by it). I can probably have these create a `Settlement` and have a function that processes everything for a settlement. Then all three would be doing the same thing, at least for all the transfers.

I'll work on that and get back to you.

SpicyLemon: Update: I did overhaul `FillAsks` and `FillBids` to shorten them up quite a bit. They're still really similar, but just different enough that extracting more to generic helpers would have lots of "If asks, do this, if bids, do that". But I also split a `closeSettlement` method out of `SettleOrders` and reused that for all the transferring stuff with `FillAsks` and `FillBids` too.

x/exchange/keeper/fulfillment.go
SpicyLemon: In other words, if I don't do this here, I have to either redo it in more individual pieces or put up with the possibility of panics or other weird behavior.

E.g. because I know everything in the message is valid, I can use `sdk.MustAccAddressFromBech32(msg.Buyer)` below without fear of it causing a panic. Without calling `ValidateBasic` here, I'd need to use `sdk.AccAddressFromBech32` and handle the possible error specifically.

x/exchange/keeper/market.go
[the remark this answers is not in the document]
Taztingo: I'm not sure that I agree that comments are a replacement for error checks unless there is a very specific reason why that the error check cannot be there (performance). However, now that I have put more thought into this I believe that this comment is actually doing more harm than good. The param is a shallow copy so the contents of Market aren't necessarily safe because it has pointers. This can actually cause more confusion than passing in by pointer because with the pointer they could expect modifications to persist. I would recommend one of the following:

1. Keep Market a value, remove the comment, and don't modify the values on the param.
2. Create a read only interface that exposes the methods needed and use that as the param type, and remove the comment.

An example of modifying a pass by value struct: https://go.dev/play/p/8TEYj0qd0on

x/exchange/msg.go
Taztingo: Whoops my mistake on that, thought I saw reference to "the happy path is left-aligned" in there too. That quote is what I'm referring to though. Are you trying to have a single point of return to aggregate errors? I don't think there are any errors to aggregate in the rainy day path for this function.

x/exchange/genesis.go
```diff
@@ -0,0 +1,76 @@
...
+			maxOrderID = order.OrderId
+		}
+	}
+
+	if g.LastOrderId < maxOrderID {
+		errs = append(errs, fmt.Errorf("last order id %d is less than the largest id in the provided orders %d",
+			g.LastOrderId, maxOrderID))
+	}
+
+	// No validation to do on LastMarketId.
```
Taztingo, line 73: We probably don't need this comment.

x/exchange/msg_test.go
```diff
@@ -0,0 +1,2237 @@
...
+)
+
+const (
+	emptyAddrErr = "empty address string is not allowed"
+	bech32Err    = "decoding bech32 failed: "
+)
+
+func TestAllMsgsGetSigners(t *testing.T) {
+	// getTypeName gets just the type name of the provided thing, e.g. "MsgGovCreateMarketRequest".
+	getTypeName := func(thing interface{}) string {
```
Taztingo, line 24: Can rename this to getMsgTypeName to make it a bit clearer. Can we make the type of thing stricter? I think it's only interacting with sdk.Msg.

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+	}{
+		{name: "Ask", str: OrderTypeAsk, b: OrderTypeByteAsk},
+		{name: "Bid", str: OrderTypeBid, b: OrderTypeByteBid},
+	}
+
+	ot := func(name string) string {
+		return "OrderType" + name
+	}
+	otb := func(name string) string {
+		return "OrderTypeByte" + name
```
Taztingo, line 183: Are these both needed? They are only used in a string formatter.

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+			}
+			assertions.RequirePanicEquals(t, testFunc, tc.expPanic, "GetHoldAmount()")
+			assert.Equal(t, tc.expected.String(), actual.String(), "GetHoldAmount() result")
+		})
+	}
+}
+
+func TestOrder_Validate(t *testing.T) {
+	// Annoyingly, sdkmath.Int{} (the zero value) causes panic whenever you
+	// try to do anything with it. So we have to give it something.
```
Taztingo, line 1163: This is nice to know, but probably isn't needed since it was fixed.

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.Order.Validate()
+			assertions.AssertErrorContents(t, err, tc.exp, "Validate() error")
+		})
+	}
+}
+
+func TestOrder_Split(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
```
Taztingo, line 1206: Is this the same as sdk.NewInt64Coin?

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+}
+
+func TestAskOrder_CopyChange(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
+		return sdk.Coin{Denom: denom, Amount: sdkmath.NewInt(amount)}
+	}
+	coinP := func(amount int64, denom string) *sdk.Coin {
+		rv := coin(amount, denom)
+		return &rv
+	}
```
Taztingo, line 1888: Same as above. Not entirely sure if needed since we do have NewInt64Coin.

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+				actual = tc.order.GetHoldAmount()
+			}
+			require.NotPanics(t, testFunc, "GetHoldAmount()")
+			assert.Equal(t, tc.exp, actual, "GetHoldAmount() result")
+		})
+	}
+}
+
+func TestBidOrder_Validate(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
```
Taztingo, line 2272: Same as above. It would be helpful to reuse existing functions so we don't have to redefine this.

x/exchange/orders_test.go
```diff
@@ -0,0 +1,2786 @@
...
+			err := tc.order.Validate()
+			assertions.AssertErrorContents(t, err, tc.exp, "Validate() error")
+		})
+	}
+}
+
+func TestBidOrder_CopyChange(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
+		return sdk.Coin{Denom: denom, Amount: sdkmath.NewInt(amount)}
+	}
```
Taztingo, line 2503: Same as above.

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+			if !assert.Equal(t, orig, tc.receiver, "receiver before and after GetAsInputs()") {
+				t.Logf("Before: %s", indexedAddrAmtsString(orig))
+				t.Logf(" After: %s", indexedAddrAmtsString(tc.receiver))
+			}
+		})
+	}
+}
+
+func TestIndexedAddrAmts_GetAsOutputs(t *testing.T) {
+	coins := func(coins string) sdk.Coins {
```
Taztingo, line 1474: Does this has to be redefined multiple times?

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+			}
+			require.NotPanics(t, testFunc, "GetAssetsUnfilled()")
+			assert.Equal(t, tc.exp.String(), actual.String(), "GetAssetsUnfilled() result")
+		})
+	}
+}
+
+func TestOrderFulfillment_GetPriceApplied(t *testing.T) {
+	coin := func(amt int64) sdk.Coin {
+		return sdk.Coin{Denom: "price", Amount: sdkmath.NewInt(amt)}
```
Taztingo, line 2191: Same as above

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+				actual = tc.f.GetPriceApplied()
+			}
+			require.NotPanics(t, testFunc, "GetPriceApplied()")
+			assert.Equal(t, tc.exp.String(), actual.String(), "GetPriceApplied() result")
+		})
+	}
+}
+
+func TestOrderFulfillment_GetPriceLeft(t *testing.T) {
+	coin := func(amt int64) sdk.Coin {
```
Taztingo, line 2229: Same as above.

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+				actual = tc.f.GetOwner()
+			}
+			require.NotPanics(t, testFunc, "GetOwner()")
+			assert.Equal(t, tc.exp, actual, "GetOwner() result")
+		})
+	}
+}
+
+func TestOrderFulfillment_GetAssets(t *testing.T) {
+	coin := func(amt int64) sdk.Coin {
```
Taztingo, line 2426: Same as above for all of these. You could even consider a factory for all your different types of anons.

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+				actual = tc.f.GetPrice()
+			}
+			require.NotPanics(t, testFunc, "GetPrice()")
+			assert.Equal(t, tc.exp.String(), actual.String(), "GetPrice() result")
+		})
+	}
+}
+
+func TestOrderFulfillment_GetSettlementFees(t *testing.T) {
+	coin := func(amt int64) *sdk.Coin {
```
Taztingo, line 2506: Same as above.

x/exchange/fulfillment_test.go
```diff
@@ -0,0 +1,6220 @@
...
+				actual = tc.f.PartialFillAllowed()
+			}
+			require.NotPanics(t, testFunc, "PartialFillAllowed()")
+			assert.Equal(t, tc.exp, actual, "PartialFillAllowed() result")
+		})
+	}
+}
+
+func TestOrderFulfillment_GetExternalID(t *testing.T) {
+	askOrder := func(externalID string) orderFulfillment {
```
Taztingo, line 2588: Same as above.

x/exchange/market_test.go
```diff
@@ -0,0 +1,4198 @@
...
+			}
+			require.NotPanics(t, testFunc, "ValidateAddRemoveFeeOptions")
+
+			assertions.AssertErrorValue(t, err, tc.expErr, "ValidateAddRemoveFeeOptions")
+		})
+	}
+}
+
+func TestValidateAddRemoveFeeOptionsWithExisting(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
```
Taztingo, line 2102: sdk.NewInt64Coin.

x/exchange/market_test.go
```diff
@@ -0,0 +1,4198 @@
...
+			assertions.AssertErrorValue(t, err, tc.expErr, "ValidateAddRemoveFeeRatiosWithExisting error")
+		})
+	}
+}
+
+func TestValidateRatioDenoms(t *testing.T) {
+	coin := func(amount int64, denom string) sdk.Coin {
+		return sdk.Coin{Denom: denom, Amount: sdkmath.NewInt(amount)}
+	}
+	ratio := func(price, fee sdk.Coin) FeeRatio {
```
Taztingo, line 1831: Is this just a constructor?

Taztingo:
We will need to update internal/stargate_whitelist.go to contain the exchange queries we want provwasm to support.

x/exchange/keeper/grpc_query.go
```diff
@@ -0,0 +1,413 @@
...
+	if err := k.ValidateMarket(ctx, msg.MarketId); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
+		resp.Error = errors.Join(errs...).Error()
+	}
+
+	return resp, nil
+}
```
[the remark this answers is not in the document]
SpicyLemon: The `CreateMarket` and `ManageFees` endpoints are gov prop only. So the stateful validation only happens after the proposal passes and voting period ends, which is an annoying time to find out you've got a problem. So there isn't much stateful validation that happens with those, which can result in a market being in a state that can make it difficult or impossible to actually use.

The `QueryValidateCreateMarket` and `QueryValidateManageFees` exist so some extra validation can be done prior to submitting a gov prop (or after it's submitted, but before voting ends). The `QueryValidateMarket` exists to help identify any problems that weren't noticed as a gov prop.

SpicyLemon:
> We will need to update internal/stargate_whitelist.go to contain the exchange queries we want provwasm to support.

Oops! Good call. I need to add the hold module queries too.

x/exchange/market_test.go
SpicyLemon: * `sdk.NewInt64Coin` is 20 characters long, `coin` is 4. So a) the important bits are a larger percent of the text, and b) one or two extra calls can be put on a single line before it gets long enough to be split up.
* `sdk.NewInt64Coin` runs validation and panics if the denom doesn't match the regex or the amount is negative. That might be handy for production code, but it's often detrimental to unit tests.
* `sdk.NewInt64Coin` has the arguments backwards. Whenever anyone talks or thinks about coins, it's `<amount>` `<denom>`, but for some dumb reason, the SDK decided they wanted the denom first. So this `coin` function is easier to read.

x/exchange/fulfillment_test.go
SpicyLemon: These are defined individually in each unit test in a way that makes sense for the test.

E.g., this one takes in a string to convert, while another might take in a single amount and denom.

x/exchange/orders_test.go
SpicyLemon: They're redefined in each test because 
* I want the function's name to be short so that there's less noise to sift through while reading the test cases
* I don't want a `coin` function in the package since it's often handy to have a variable named `coin`. So if it's a package-level function, it needs to be longer.
* Different tests have different needs, and it's not always handy to have `(amount, denom)` for the args. E.g. maybe it's easier in some test to have a `(string)`.
* By defining it in the test, I don't have to have `t *testing.T` as an arg for any of them that need to run a check (e.g. make sure an error is nil).

x/exchange/orders_test.go
SpicyLemon: `NewInt64Coin` returns a concrete `Coin` struct, and you can't take the reference of a function result. E.g. you can't do `&sdk.NewInt64Coin(...)`.

x/exchange/msg_test.go
```diff
@@ -0,0 +1,2237 @@
...
+	"github.com/provenance-io/provenance/testutil/assertions"
+)
+
+const (
+	emptyAddrErr = "empty address string is not allowed"
+	bech32Err    = "decoding bech32 failed: "
+)
+
+func TestAllMsgsGetSigners(t *testing.T) {
+	// getTypeName gets just the type name of the provided thing, e.g. "MsgGovCreateMarketRequest".
```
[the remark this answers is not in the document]
SpicyLemon: It's a function comment, just like any other function should have. And the comment inside the function doesn't give you an example of what the result will look like, just what a specific step might look like. And having the example also saves you the time of having to read and understand the function in order to know what the results look like.

I don't get why you all are so hellbent on not having comments.

x/exchange/msg_test.go
SpicyLemon: It's an anonymous function used only for the purposes of this unit test. One benefit of them is it's easier to keep them shorter since they only need to be know in the context of this one unit test. In here, we are only concerned about the message type names. And there's no real reason to make it stricter since the extra limitations don't buy us anything .

x/exchange/orders_test.go
SpicyLemon: `sdkmath.Int{}.Equal(sdkmath.Zero())` will panic, as will `sdkmath.ZeroInt().Equal(sdkmath.Int{})`. Same with all the comparison and math methods including `.IsZero()`, `.IsPositive()`, `.IsNegative()`, and `.Add(...)`.

Unrelatedly, but also annoying as hell, `coinA.IsEqual(coinB)` will also panic if the two coins have different denoms (regardless of the amounts). Yay! And `coinsA.IsEqual(sdk.Coin{Denom: <same as A>})` will also panic because of the empty int comparison.

x/exchange/orders_test.go
SpicyLemon: See https://github.com/provenance-io/provenance/pull/1661#discussion_r1362633596

x/exchange/fulfillment_test.go
SpicyLemon: A factory is significant overkill for most of this stuff and would then also require extra testing to make sure the factory is making what you're expecting it to make.

x/exchange/msg_test.go
Taztingo: The benefit is clarity, and it makes it easier to understand as you trace through the test.

x/exchange/fulfillment_test.go
iramiller: These don't seem like a big impact on readability or maintainability to me ...

x/exchange/keeper/market.go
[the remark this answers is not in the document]
SpicyLemon: It's not the comment that's replacing the nil-case error check, it's that it can't possibly be nil here, so it doesn't make sense to check if it's nil.

A read-only interface is yet even more code that would need to be written and tested just to replace two lines of comments.

I'll consider removing the comment once there's a unit test ensuring the provided market doesn't change during this call. But again, the comment is pointing out behavior that isn't obvious.

x/exchange/msg.go
SpicyLemon: But it's not self explanatory. In order to know that there isn't any validation involving that field, you have to have intimate knowledge of the containing struct and how it should behave.

x/exchange/msg_test.go
SpicyLemon: Also, as it's written, it's not specific to messages. It's available to use on anything. E.g. It then can be used on things that haven't yet implemented `sdk.Msg`, but will need to at some point.

x/exchange/keeper/keeper.go
SpicyLemon: There's no need to recreate the exact same thing over and over again, specially when it's immutable. Calling a function (e.g. `NewInt`) requires more resources than accessing a variable (even when the function is just returning a variable (e.g. `GetAmount() string { return f.amount }`)).

Yes, the 10,000 is for converting from basis points to whatever it's called as an actual value. Usually, you'd think of it as amount * (split/10,000), but with integer math, we need to do that multiplication before the division.

As it is the line is, "give me the quotient and remainder when I multiply the amount times the split and divide it by 10,000." Calling the variable anything other than "ten thousand" would make that line less readable because you then don't actually know what math is being done; "what am I dividing by here?"

`QuoRem` is a function that `math.BigInt` has, but the sdk didn't see fit to also have that method for an `sdkmath.Int`, which is why I had to make that `exchange.QuoRemInt` function.

x/exchange/params.go
SpicyLemon: After some offline discussion, we think this is a decent value to have until we know better. At which point, it'll probably already be in use and will be updated with a param change.

proto/provenance/exchange/v1/events.proto
Taztingo: Just to make sure we understand my concern... This was more about reducing the amount of code we have since we don't have any requirements by the consumer.

nullpointer0x00 (approved):

Taztingo (approved):

Issue #1660, opened 2023-08-23
Add genesis export/import for Deny Send Addresses in Marker module


nullpointer0x00:
## Summary

The export and init genesis for marker module is missing the Deny Send Addresses. 

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1662, opened 2023-08-24
Add genesis export/init for marker deny send lists

nullpointer0x00:
## Description

closes: #1660 

---

write a little note why.


Taztingo (approved):
LGTM

x/marker/types/key_test.go
```diff
@@ -34,3 +34,13 @@ func TestDenySendKey(t *testing.T) {
...
 }
+
+func TestGetDenySendAddresses(t *testing.T) {
+	addr, err := MarkerAddress("nhash")
+	require.NoError(t, err)
+	denyAddr := sdk.AccAddress("cosmos1v57fx2l2rt6ehujuu99u2fw05779m5e2ux4z2h")
+	denyKey := DenySendKey(addr, denyAddr)
+	mAddr, dAddr := GetDenySendAddresses(denyKey)
+	assert.Equal(t, mAddr, addr)
+	assert.Equal(t, dAddr, denyAddr)
```
SpicyLemon, line 45: Your expected and actual are backwards in this call: `Equal(t TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{})`. Additionally, **ALWAYS** provide a message with your assertions. That way, when it fails, the failure message tells you what you're looking at.

```suggestion
	assert.Equal(t, addr, mAddr, "module address")
	assert.Equal(t, denyAddr, dAddr, "deny address")
```

proto/provenance/marker/v1/genesis.proto
```diff
@@ -19,4 +19,18 @@ message GenesisState {
...
+
+// DenySendAddress defines addresses that are denied sends for marker denom
+message DenySendAddress {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // marker_address is the marker's address for denied address
+  string marker_address = 1;
+  // deny_address defines all wallet addresses that are denied sends for the marker
+  string deny_address = 3;
```
iramiller, line 35: what happened to position 2 in this proto?

x/marker/simulation/operations.go
```diff
@@ -368,6 +371,32 @@ func SimulateMsgSetAccountData(k keeper.Keeper, args *WeightedOpsArgs) simtypes.
...
+	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
+		msg := &types.MsgUpdateSendDenyListRequest{}
+
+		marker, signer := randomMarkerWithAccessSigner(r, ctx, k, accs, types.Access_Transfer)
+		if marker == nil {
+			return simtypes.NoOpMsg(sdk.MsgTypeURL(msg), sdk.MsgTypeURL(msg), "unable to find marker with a transfer signer"), nil, nil
+		}
+
+		rDenyAccounts := simtypes.RandomAccounts(r, 10)
+		addDenyAddresses := make([]string, len(rDenyAccounts))
```
iramiller, line 387: If you are randomly adding addresses to this list (but never removing them) have you noted that all addresses for senders are eventually blocked? What kind of simulation is intended here and have you observed/proven that it properly occurs?  

How many send operations (overall) occur, how many get blocked and what percentage of the time does this occur (not how often is this method invoked which is noted as 10 in the weights above).

I am also surprised there is not a corresponding remove deny list operation.

proto/provenance/marker/v1/genesis.proto
nullpointer0x00: Avoid merge conflict with Nav PR. This was on purpose.

x/marker/types/key_test.go
nullpointer0x00: Oops, mistake on the order. I will add the messages, but I don't think that is always necessary. The test name, line number variable names, and expected values messages are usually enough.

SpicyLemon: Instances where a msg/args isn't necessary are very rare. The line number only helps you when you're looking at the test code and when no code has changed yet in the file (to alter the line number). When you rely on the line number, you have to do this back and forth comparing the failure output and the code. That becomes even more painful when the error message isn't available in your IDE (e.g. it's in a Github Action).

When you include a message, you're putting that information in the test failure message with all the other information about the failure; it's all in one place. It also makes it easier to identify exactly what is wrong which often lets you skip digging into the unit test and instead go directly to the production code to fix the problem.

In this case, you've got two `Equals` tests on byte slices. The values in the failure message are not going to provide any clue as to which byte slices you're looking at. Knowing which one isn't as expected tells you exactly what to look at in `GetDenySendAddresses`.

In the rare cases where a msg/args might not be necessary, having them is often still helpful, and are only detrimental if what they contain isn't accurate. Once you're used to always adding them, they take almost no time to add. And having them will save you plenty of time in the end.

Always including them in your assertions is a good habit to get into.

SpicyLemon: Also, the `.NoError` above should have a message too since the error isn't coming from the `GetDenySendAddresses` that the test is about. As it is, if there's an error there, you'll probably waste time looking at `GetDenySendAddresses` first, only to find it doesn't return an error. Then, you'd go to the test code to look up what exactly was called that returned the error.

If you include the msg "MarkerAddress(nhash)" in that assertion, when it fails, you'll know exactly what's broken, and that that unit test will (probably) be fixed when you fix `MarkerAddress`. You don't even have to look at the unit test code. You just go to `MarkerAddress`, identify and fix the problem, and rerun unit tests (since lots of things probably failed as well).

Taztingo (approved):

proto/provenance/marker/v1/genesis.proto
iramiller: I don't see the update from the NAV PR which has been merged in...

nullpointer0x00: 🤦 ugh, I had my wires crossed.  I did an increment jump in the GenesisState proto in prep for merge.  This one was a mistake.  Thought I was looking at the GenesisState here in the 4 lines the comment show.  Sorry.  Fixed.

iramiller (approved):

Taztingo (approved):

Issue #1127, opened 2022-10-05
Investigate Marker movement(restricted markers especially)via  ibc protocol(marker-ics20 for now) based on cosmwasm design 


arnabmitra:
## Summary
This is a task to POC/investigate send markers(including restricted markers) via ics-20 between chains
Based on this writeup 

Several threads here i think:
* https://github.com/CosmWasm/cosmwasm/blob/main/IBC.md#writing-new-protocols seems like this is similar to what we want with a marker-ics20 protocol
* In order to enable IBC communication, a contract must expose the following 6 entry points. Upon detecting such an "IBC-Enabled" contract, the [x/wasm runtime](https://github.com/CosmWasm/wasmd) will automatically bind a port for this contract (wasm.<contract-address>), which allows a relayer to create channels between this contract and another chain. Once channels are created, the contract will process all packets and receipts.
Seems this is what we want for the `marker-ics20` protocol i think

These hold even if we want to do the same via `ica` i think

## Problem Definition

 Why do we need this feature? 
IBC transfers will be completely impossible after v1.13 and there is some desire to move them around between chains, see #1093 
Also markers should be created with certain restrictions on the destination chains for e.g MINT/BURN should be disabled on destination chain marker etc

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

arnabmitra:
hackmd based on discussions https://hackmd.io/3fwzGtlbRRGiaBAUdZyyGg

iramiller:
Also #1425

Taztingo:
It may be worth investigating if it's better to do something like this as Middleware.

PR #1678, opened 2023-09-11
Marker ics20 hooks 

nullpointer0x00:
## Description

closes: #1127 

---

write a little note why.


x/ibchooks/wasm_hook.go
```diff
@@ -340,7 +338,7 @@
 
 	sudoMsg := []byte(fmt.Sprintf(
 		`{"ibc_lifecycle_complete": {"ibc_ack": {"channel": "%s", "sequence": %d, "ack": %s, "success": %s}}}`,
-		sanitizedSourceChannel, packet.Sequence, ack, success))
+		sanitizedSourceChannel, packet.Sequence, ackAsJSON, success))
```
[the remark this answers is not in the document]
iramiller: This is seemingly a false positive ... but annoying as hell.  We need to find a way to either obfuscate this to make the code ql test happy or explicitly exclude it.  Afaict there is no way to cause valid json (which is marshaled as such above) injected as an attribute value which is not quoted to break out ...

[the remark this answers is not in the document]
nullpointer0x00: I think I found a way.  Use structs and just serialize the json instead of doing this sprintf way.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,226 @@
...
+				sdktypes.NewInt64Coin(ibcDenom, amount),
+				nil,
+				nil,
+				markertypes.StatusActive,
+				coinType,
+				false, // supply fixed
+				false, // allow gov
+				false, // allow force transfer
+				[]string{},
+			)
```
nullpointer0x00, line 75: The logic in here needs to be discussed and understood by the team.  I took a best first guess at how these should be setup, but decided to defer it to a larger discussion.

x/ibchooks/ibc_hooks.go
```diff
@@ -0,0 +1,87 @@
...
+		cdc:                     cdc,
+		ibcKeeper:               ibcKeeper,
+		ibcHooksKeeper:          ibcHooksKeeper,
+		wasmHooks:               wasmHooks,
+		markerHooks:             markerHooks,
+		SendPacketPreProcessors: preSendPacketDataProcessingFns,
+	}
+}
+
+// ProperlyConfigured returns false if either wasm or marker hooks are configured properly
```
Taztingo, line 36: Minor typo. I think you mean this:

ProperlyConfigured returns false if either wasm or marker hooks are configured **incorrectly**.

x/ibchooks/types/types_test.go
```diff
@@ -0,0 +1,35 @@
...
+	s.Require().NoError(err)
+	s.Require().Equal(`{"ibc_lifecycle_complete":{"ibc_ack":{"channel":"channel-1","sequence":100,"ack":{"packet_sequence":100,"source_channel":"channel"},"success":true}}}`, string(sudoMsg))
+}
+
+func (s *IbcHooksTypesTestSuite) TestIbcLifecycleCompleteTimeoutJsonSerialization() {
+	ibcLifecycleCompleteAck := NewIbcLifecycleCompleteTimeout("channel-1", 100)
+	sudoMsg, err := json.Marshal(ibcLifecycleCompleteAck)
+	s.Require().NoError(err)
+	s.Require().Equal(`{"ibc_lifecycle_complete":{"ibc_timeout":{"channel":"channel-1","sequence":100}}}`, string(sudoMsg))
+}
```
Taztingo, line 35: A few basic tests for NewMarkerPayload would be helpful.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,226 @@
...
+			if err = ResetMarkerAccessGrants(transferAuthAddrs, marker); err != nil {
+				return err
+			}
+
+			if err = h.MarkerKeeper.AddMarkerAccount(ctx, marker); err != nil {
+				return err
+			}
+			chainID := h.GetChainID(ctx, packet, ibcKeeper)
+			markerMetadata := banktypes.Metadata{
```
Taztingo, line 84: This is really nice, and will be super helpful.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// AddUpdateMarker will add or update ibc Marker with transfer authorities
+func (h MarkerHooks) AddUpdateMarker(ctx sdktypes.Context, packet exported.PacketI, ibcKeeper *ibckeeper.Keeper) error {
+	var data transfertypes.FungibleTokenPacketData
+	if err := json.Unmarshal(packet.GetData(), &data); err != nil {
+		return err
+	}
+	ibcDenom := MustExtractDenomFromPacketOnRecv(packet)
+	if strings.HasPrefix(ibcDenom, "ibc/") {
```
Taztingo, line 45: I would actually consider this the happy path because it contains the most logic. So you can remove one level of indentation by making this 

```
if !strings.HasPrefix(ibcDenom, "ibc/") {
  return nil
}
```

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,226 @@
...
+		marker, err := h.MarkerKeeper.GetMarker(ctx, markerAddress)
+		if err != nil {
+			return err
+		}
+		var transferAuthAddrs []sdktypes.AccAddress
+		transferAuthAddrs, coinType, err := ProcessMarkerMemo(data.GetMemo())
+		if err != nil {
+			return err
+		}
+		if marker == nil {
```
Taztingo, line 59: Same comment as above. You can remove this indentation by making this the happy path. It's not super important, but will help improve clarity.

```
if marker != nil {
  if err = ResetMarkerAccessGrants(transferAuthAddrs, marker); err != nil {
    return err
  }
h.MarkerKeeper.SetMarker(ctx, marker)
return nil
}
```

x/ibchooks/marker_hooks.go
Taztingo: This logic makes sense to me. I haven't looked at the send portion yet.

x/ibchooks/types/types_test.go
nullpointer0x00: Done.

x/ibchooks/marker_hooks.go
nullpointer0x00: Thanks!  We could probably extract it to a general function in the ibc-hooks later when it might need to be shared.

x/ibchooks/marker_hooks.go
nullpointer0x00: True, I also made a few more refactors to the method.  Extracted the rest of the logic into it's own method.  This makes it clearer what is happening after `marker != nil`, because the method name says what it does along with method comments instead adding comments in the middle of method. 
It seemed a little confusing after moving things around and the method was getting a bit long.  I hope this makes it easier to understand.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,235 @@
...
+	marker := markertypes.NewMarkerAccount(
+		authtypes.NewBaseAccountWithAddress(markertypes.MustGetMarkerAddress(ibcDenom)),
+		sdktypes.NewInt64Coin(ibcDenom, amount),
+		nil,
+		nil,
+		markertypes.StatusActive,
+		coinType,
+		false, // supply fixed
+		false, // allow gov
+		false, // allow force transfer
```
tbrooks-figure, line 88: Bringing this out of a slack conversation. Would it be possible to reflect the `allow_forced_transfer` from the source chain to the destination? Figure/FES has some use cases around that.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			return nil, markertypes.MarkerType_Unknown, err
+		}
+		transferAuths[i] = accAddr
+	}
+	return transferAuths, markertypes.MarkerType_RestrictedCoin, nil
+}
+
+// ProcessMarkerMemoFn processes a ics20 packets memo part to have `marker` setup information for receiving chain
+func (h MarkerHooks) ProcessMarkerMemoFn(
```
Taztingo, line 183: This name is really similar to the previous one. I would recommend changing one of them.

x/ibchooks/marker_hooks.go
```diff
@@ -0,0 +1,235 @@
...
+	ctx sdktypes.Context,
+	data []byte,
+	_ map[string]interface{},
+) ([]byte, error) {
```
Taztingo, line 187: The only concern I have is the denom of the ICS20 token being sent. Will it correctly be reverted to its original ibc/.. format?

x/ibchooks/marker_hooks_test.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+		{
+			name:        "successfully process with non json memo",
+			denom:       "fiftyfivehamburgers",
+			memo:        "55 burger 55 fries...",
+			expErr:      "",
+			expIbcDenom: "ibc/F3F4565153F3DD64470F075D6D6B1CB183F06EB55B287CCD0D3506277A03DE8E",
+		},
+		{
+			name:        "successfully process with non json memo",
```
Taztingo, line 111: This name is the same as the previous one. Do you mean invalid json memo?

x/ibchooks/marker_hooks_test.go
```diff
@@ -0,0 +1,360 @@
...
+			transferAuths: []sdk.AccAddress{address1, address2},
+			markerAcct:    *markertypes.NewEmptyMarkerAccount("jackthecat", address1.String(), []types.AccessGrant{}),
+			expErr:        "",
+		},
+		{
+			name:          "successfully reset marker access grants and remove other transfer grant",
+			transferAuths: []sdk.AccAddress{address1, address2},
+			markerAcct:    *markertypes.NewEmptyMarkerAccount("jackthecat", address1.String(), []types.AccessGrant{*types.NewAccessGrant(address3, []types.Access{types.Access_Transfer})}),
+			expErr:        "",
+		},
```
Taztingo, line 252: I would probably add a case where transferAuths is nil and one when its empty.

x/ibchooks/marker_hooks_test.go
```diff
@@ -0,0 +1,360 @@
...
+		{
+			name:    "mal-formed json should be moved to memo",
+			memo:    `{"marker":{"transfer-auths":["123", "345"]}`,
+			expMemo: "{\"memo\":\"{\\\"marker\\\":{\\\"transfer-auths\\\":[\\\"123\\\", \\\"345\\\"]}\"}",
+		},
+		{
+			name:    "correct json should not modify memo",
+			memo:    `{"marker":{"transfer-auths":["address"]}}`,
+			expMemo: `{"marker":{"transfer-auths":["address"]}}`,
```
Taztingo, line 292: Adding when memo is nil OR empty will give you coverage for when that branch isn't hit inside SanitizeMemo.

x/ibchooks/marker_hooks.go
nullpointer0x00: I'm not sure I follow your question?  The code isn't changing the denom being sent or received.

Taztingo: You can disregard this. I saw the Name and Display on the marker and thought users would be transferring with those names.

x/ibchooks/marker_hooks_test.go
nullpointer0x00: Good call.

x/ibchooks/marker_hooks.go
nullpointer0x00: Yes, I can have the source chain add the force transfer flag along with the transfer auths.

Taztingo (dismissed):
LGTM

iramiller (approved):

x/ibchooks/marker_hooks.go
nullpointer0x00: Done and ready to go.

Taztingo (approved):

Issue #1706, opened 2023-10-13
Ensure IBC marker has matching supply


iramiller:
## Summary of Bug

When a marker is created via ICS20 transfer for an existing marker, the supply amount must match existing values

## Version

main/HEAD

## Steps to Reproduce

When an existing ibc token has a marker created for it the desired supply must match the existing total on the chain.

When a marker is created the desired supply will mint/burn amounts as required to adjust current supply against desired supply regardless of if fixed supply is selected or not.  This can result in issues where there are existing tokens of an IBC denom on the network since the desired supply will not match.

## Remediation

The desired supply must be set to the total existing supply when the marker is created.  Further checks should be put in place to ensure that the supply after the token transfer is completed align correctly.  In other words the mint associated with the inbound transfer must be taken into account.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1708, opened 2023-10-13
Ensure IBC marker has matching supply 

nullpointer0x00:
…efore marker

## Description

closes: #1706 

---

write a little note why.


x/ibchooks/marker_hooks.go
```diff
@@ -96,6 +96,7 @@ func (h MarkerHooks) createNewIbcMarker(ctx sdktypes.Context, data transfertypes
 		allowForceTransfer,
 		[]string{},
 	)
+	h.adjustSupply(ctx, marker)
```
iramiller, line 99: the terminology used here is confusing ... it would be more clear if this was structured as "get existing supply" then add the incoming amount to it.  Existing supply would be zero or more...

nullpointer0x00: Thanks, I was waffling on what to call it that described the action better.

x/ibchooks/marker_hooks.go
```diff
@@ -96,6 +96,7 @@ func (h MarkerHooks) createNewIbcMarker(ctx sdktypes.Context, data transfertypes
 		allowForceTransfer,
 		[]string{},
 	)
+	h.getExistingSupply(ctx, marker)
```
iramiller, line 99: This should return the existing supply so it can be added here in this method to make it clear what is going on.  This current approach hides that the existing supply is being added to the amount that is inbound from the transfer.

Taztingo (approved):
LGTM

iramiller (approved):

x/ibchooks/marker_hooks.go
nullpointer0x00: Yeah, I refactored it a minute ago.  I realized I mis-understood your first comment on my other refactor.

Issue #1634, opened 2023-07-25
ARM64 Container Builds


hyperschwartz:
## Summary

The produced docker containers for this repository currently only support AMD64 architectures.  It would be amazing to get ARM64 multi-platform builds for the containers in the future.

## Problem Definition

Figure does a large amount of testing with locally-built Provenance containers.  With the introduction of the M1 Macbook and later generations, local development has decreased in efficiency due to the need for these machines to do AMD64 emulation when running the containers.

## Proposal

Please modify the docker build/publication to include ARM64 multi-platform images.  Docker buildx is already used, so I believe this should not require a large number of changes to the build configurations.
 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
A bit ago, we removed the ARM64 build because of problems signing it (with Apple). Building and publishing of the related docker container was also removed the since it needed that executable.

Because of the Apple signing requirement (and required hurdles and fees), we probably don't want to publish the mac executable as a release artifact again. But maybe published as a docker container, Mac's won't incessantly complain about it not being signed.

iramiller:
Correct ... you can verify this by running the heighliner container.
```
 docker run --rm -it ghcr.io/strangelove-ventures/heighliner/provenance
 
 ~ $ provenanced version
v1.12.0-rc1
~ $ arch
aarch64
```

PR #1709, opened 2023-10-17
Add arm64 to docker publish workflow

nullpointer0x00:
## Description

closes: #1634

---

write a little note why.


.github/workflows/docker.yml
```diff
@@ -56,24 +58,28 @@ jobs:
...
+
       - name: Publish to Docker Hub
         uses: docker/build-push-action@v5
         with:
           context: .
           target: run
           build-args: |
             VERSION=${{ steps.prep.outputs.version }}
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
```
iramiller, line 82: Is that really all we needed?

nullpointer0x00: I believe so, but I'm still testing it.  I will turn this into review mode if it succeeds.

Taztingo: Does this need qemu defined to emulate arm64? I don't really know too much about the machine running this, but saw this in the documentation:

https://github.com/docker/build-push-action#usage

nullpointer0x00: I don't think so.  I believe the Figure team just wanted to be able to pull down an arm64 docker image.  We only had the amd64.  

I tested this by modifying the workflow to execute and publish from my branch https://hub.docker.com/layers/provenanceio/provenance/nullpointer0x00-1634-arm64-container-builds/images/sha256-e6189e7ab7d8b4181dc2c7fd996d3de062d633a789f625c70d1b998182a87322?context=explore. We can delete that image after this is merged.

docker/blockchain/Dockerfile
Taztingo, line 1: I think there is still an issue with this file. I looked at the outputted docker file and it still is attempting to copy libwasmvm.x86_64.so instead of the arm version of it.

I think ARCH as to be correctly updated and then it should work.

nullpointer0x00: Yeah you are correct. I'm fixing that.  The arch is being sent in from the uname command from the host machine... However, that is always a x86_64 when running the gh workflow.  I had to move the uname command to the Dockerfile since it is running on the docker contain for the specific platform.

SpicyLemon:
The [docker publish action](https://github.com/provenance-io/provenance/actions/runs/6566918022/job/17838621292?pr=1709) is failing:

```
 > [linux/amd64 run 4/6] COPY --from=build /go/src/github.com/provenance-io/provenance/vendor/github.com/CosmWasm/wasmvm/internal/api/libwasmvm..so /usr/local/lib:
------
Dockerfile:39
--------------------
  37 |         rm -rf /var/lib/apt/lists/
  38 |     
  39 | >>> COPY --from=build /go/src/github.com/provenance-io/provenance/vendor/github.com/CosmWasm/wasmvm/internal/api/libwasmvm.$ARCH.so /usr/local/lib
  40 |     COPY --from=build /go/bin/provenanced /usr/bin/provenanced
  41 |     
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref og41bfipp9bxrbsg3ryn46uv5::ko8qnspaa5pgrom4x3vn57x2c: "/go/src/github.com/provenance-io/provenance/vendor/github.com/CosmWasm/wasmvm/internal/api/libwasmvm..so": not found
Error: buildx failed with: ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref og41bfipp9bxrbsg3ryn46uv5::ko8qnspaa5pgrom4x3vn57x2c: "/go/src/github.com/provenance-io/provenance/vendor/github.com/CosmWasm/wasmvm/internal/api/libwasmvm..so": not found
```

nullpointer0x00:
@SpicyLemon I know it is failing.  Have been in the middle of fixing it.  Meant to move it back to draft mode.

docker/blockchain/Dockerfile
```diff
@@ -27,17 +26,25 @@ RUN make VERSION=${VERSION} WITH_CLEVELDB=true install
```
Taztingo, line 26: I think this could be an issue as well. It looks like the binary will be built on an x86_64 machine, and could potentially cause issues on an arm64 machine.

nullpointer0x00: RUN is a shell executed on the container not the host.  So, when github actions build a arm64 that container will be building for the arm64 arch. In this case, the `make` is being executed on the container.

I removed the whole ARG passing for builds, because we were passing the arch of the host machine not the container platform.

Taztingo (dismissed):

iramiller (dismissed):

SpicyLemon (approved):

Taztingo (approved):

iramiller (approved):

Issue #1699, opened 2023-10-09
Exchange: Finish keeper unit tests


SpicyLemon:
Finish all the keeper unit tests in the exchange module.

PR #1713, opened 2023-10-18
Dwedul/1699 exchange keeper tests

SpicyLemon:
## Description

closes: #1699 

This PR:
* Add the (empty) `saffron-rc2` upgrade.
* Adds unit tests on the exchange keeper methods.
* During `FillBids`, the seller settlement fee is now calculated on the total price instead of each order individually.
* In the `OrderFeeCalc` query, ensure the market exists.
* In several endpoints, prevent the addresses from being converted to/from string/`AccAddress` when one or the other isn't needed.
* Fixes `InitGenesis`, to ensure `LastOrderId` is at least the largest order id.
* Properly populate the permissions lists when reading access grants from state.
* Fixes the paginated order queries to properly read orders from state.

---

write a little note why.


x/exchange/keeper/grpc_query_test.go
```diff
@@ -1,27 +1,4017 @@
...
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			runQueryTestCase(s, testDef, tc)
+		})
+	}
+}
+
+func (s *TestSuite) TestQueryServer_GetAllOrders() {
+	testDef := queryTestDef[exchange.QueryGetAllOrdersRequest, exchange.QueryGetAllOrdersResponse]{
+		queryName: "GetAllOrders",
```
Taztingo, line 2941: How large can the order list get? Will this become unusable without pagination after a certain length?

x/exchange/keeper/grpc_query_test.go
```diff
@@ -1,27 +1,4017 @@
...
+
+func (s *TestSuite) TestQueryServer_GetAllMarkets() {
+	briefIDStringer := func(brief *exchange.MarketBrief) string {
+		if brief == nil {
+			return "<nil>"
+		}
+		return fmt.Sprintf("%d", brief.MarketId)
+	}
+	testDef := queryTestDef[exchange.QueryGetAllMarketsRequest, exchange.QueryGetAllMarketsResponse]{
+		queryName: "GetAllMarkets",
```
Taztingo, line 3242: I doubt this has the same issue as orders, but will this always be usable without pagination?

x/exchange/keeper/grpc_query_test.go
SpicyLemon: It is paginated. And yes, it'd be unusable at some length. But that exact length would depend on the processing power available to the node fielding the query.

The default limit is 100, which is used if no pagination parameters are set in the query, but there's no hard-coded maximum on the `limit`. If it's too high, the node will time-out the query after a few seconds.

x/exchange/keeper/grpc_query_test.go
SpicyLemon: This one also has pagination on it already. It'll probably be a long time until there's enough markets for it to be unusable without pagination, but any time there's a variable number of results in a query, it's best to include pagination.

Taztingo (approved):

iramiller (approved):

Issue #1736, opened 2023-11-07
Update Marker NAV with exchange settlement


iramiller:
## Summary

During settlement of orders in the exchange module the price paid for an asset should be set on the associated marker.

## Problem Definition

In order to track accurate pricing information for NAV on a marker updates to pricing information should be made when an asset trades.  The NAV data on a marker support price ratios and volume which should be published when exchange orders are successfully matched.

## Proposal

Add a step to record the price paid during the settlement step for the market as part of the event output:
https://github.com/provenance-io/provenance/blob/4aabad21b7bcb0d138d8dda0cd70859d9ef39d45/x/exchange/keeper/fulfillment.go#L302

Marker values is recorded with:
https://github.com/provenance-io/provenance/blob/d050e9701ddb1f5df7f23ab8b0ca6b0d1ed7e60e/x/marker/keeper/keeper.go#L240
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1741, opened 2023-11-09
Have exchange module update marker net-asset-value entries

SpicyLemon:
## Description

closes: #1736 

During a settlement, update the marker's net-asset-value.

---

write a little note why.


x/exchange/fulfillment.go
```diff
@@ -881,3 +881,58 @@ func getPriceTransfer(f *orderFulfillment) (*Transfer, error) {
 	// This is here in case a new SubTypeI is made that isn't accounted for in here.
 	panic(fmt.Errorf("%s order %d: unknown order type", f.GetOrderType(), f.GetOrderID()))
 }
+
+// NetAssetValue is a total of assets and the price they sold for.
+type NetAssetValue struct {
```
Taztingo, line 886: I can see this easily getting mixed up with the markertypes.NetAssetValue especially since it uses the MarkerKeeper.

iramiller (dismissed):
This looks like it will be a good first step in this process

x/exchange/fulfillment.go
SpicyLemon: I'm open to suggestions. I'm having trouble thinking of an accurate other name for it.

My original thinking was to call it `NetAssetValue` since it's intimately related to the marker's version. I couldn't just use the marker module's version because it doesn't have the asset denom in it, though. The only time the marker's version is used in the exchange module is the one function that makes the calls to `AddSetNetAssetValues`.

Taztingo: Works for me. It's only in one spot and we can rename it at a later time if needed.

Taztingo (dismissed):

nullpointer0x00 (approved):

iramiller (dismissed):

iramiller (approved):

iramiller (approved):

Taztingo (approved):

Issue #1757, opened 2023-11-20
Add a market to the genesis files created by our various make targets.


SpicyLemon:
## Summary

In our various `make` targets that create a new chain, add a default market to the genesis file that gets used.

## Problem Definition

Creating a market requires a governance proposal, so creating one after the chain has started can be time consuming. This makes use and testing of the exchange module more difficult than it needs to be.

## Proposal

1. Create a `add-genesis-market` command that adds a market to the genesis file.
2. Update all of the make targets that generate a default genesis file to add an appropriate market.

Here's some example `make` targets that should get this addition (this list might not be exhaustive).
* `make run-config`
* `make localnet-generate`
* `make ibcnet-generate`
* `make devnet-generate`

This can probably re-use the setup & msg-gen that the `gov-create-market` command uses, but add it to the genesis file instead of broadcasting it.

By default, it should have a name and description, 5% fees (e.g. `20<denom>:1<denom>`), and some amount of flat fees (not exactly sure what amount makes sense). The validators should get all permissions. It should not have any required attributes, and it should be enabled and allow user-settlement.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1759, opened 2023-11-22
Add a market to our test/dev chains.

SpicyLemon:
## Description

closes: #1757 

This PR adds some genesis commands for adding a market to genesis. It also uses those to add a market to our various local environments (e.g. `make run`).

---

write a little note why.

### DIFF
--- a/cmd/provenanced/cmd/genaccounts.go
+++ b/cmd/provenanced/cmd/genaccounts.go
@@ -763,7 +763,7 @@ $ ` + version.AppName + ` add-genesis-custom-market \
 
 			// Create the market and add it to the app state.
 			market := MakeDefaultMarket(feeDenom, addrs)
-			return addMarketToAppState(clientCtx, appState, market)
+			return addMarketsToAppState(clientCtx, appState, market)
 		}),
 	}
 
@@ -811,7 +811,7 @@ func AddGenesisCustomMarketCmd(defaultNodeHome string) *cobra.Command {
 
 			// Now that we've read all the flags and stuff, no need to show usage with any errors anymore in here.
 			cmd.SilenceUsage = true
-			return addMarketToAppState(clientCtx, appState, msg.Market)
+			return addMarketsToAppState(clientCtx, appState, msg.Market)
 		}),
 	}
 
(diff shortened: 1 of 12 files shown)


cmd/provenanced/cmd/pre_upgrade_test.go
```diff
@@ -211,7 +210,6 @@ func makeDummyCmd(t *testing.T, home string) *cobra.Command {
 }
 
 func TestPreUpgradeCmd(t *testing.T) {
-	app.SetConfig(true, false)
```
SpicyLemon, line 214: I removed this line because I couldn't see a reason why it would matter for this test and having it here was causing the `SimulateTestSuite` (in `cmd/provenanced/cmd/simulate_test.go`) to fail in setup.

I couldn't get it to happen locally, but it happened thrice in a row on the github action that runs our tests. Somehow, the `authority` address created in the bank module's `NewBaseKeeper` was getting one HRP, but by the time it was being checked in `NewBaseSendKeeper`, the other HRP was expected, so it was panicking. I suspect that the reason I couldn't get the failure locally was that the ordering of the tests in `github.com/provenance-io/provenance/cmd/provenanced/cmd` ends up being different in the github action vs locally.

The `AccAddress.String()` function uses some caching. At some point in these package tests, with the default HRP, something gets the gov module account address bech32 string. Later, this `app.SetConfig(true, false)` is called, changing the (global) HRP. But that `.String()` cache still has the old HRP for the given address bytes. So in the bank's `NewBaseKeeper`, the call to `authtypes.NewModuleAddress(govModuleName).String()` recognizes the bytes and returns the previously generated bech32 (with the now incorrect HRP). A few lines later, in `NewBaseSendKeeper`, it tries to decode that address, and finds the wrong HRP which makes it panic.

This was the only place where `SetConfig` was being called in this package's tests. Once I removed it, the tests started passing in the github action again.

cmd/provenanced/cmd/genaccounts.go
```diff
@@ -640,3 +690,188 @@ func checkMsgTypeValid(registry types.InterfaceRegistry, msgTypeURL string) erro
...
+				return fmt.Errorf("error reading --%s value: %w", flagDenom, err)
+			}
+			if len(feeDenom) == 0 {
+				var stGen stakingtypes.GenesisState
+				err = clientCtx.Codec.UnmarshalJSON(appState[stakingtypes.ModuleName], &stGen)
+				if err == nil {
+					feeDenom = stGen.Params.BondDenom
+				}
+			}
+			if len(feeDenom) == 0 {
```
Taztingo, line 757: Does this branch condition need to be evaluated twice? Can probably move `feeDenom = "nhash"` up to line `751` to achieve the same effect.

Taztingo (approved):

cmd/provenanced/cmd/genaccounts.go
SpicyLemon: If I put it on line 751, then, when `stGen.Params.BondDenom` is an empty string, `feeDenom` would end up empty too. To account for that, there'd need to be a `len` check on it before setting `feeDenom` from it anyway.

The nice thing about having it like this, is that it puts things in priority: "Look here for the thing. If it wasn't there, look over there for it. If we still don't have it, use this default." If I moved it up to line 751, it'd mix together the last two parts: "Look here for the thing. If it wasn't there, use this default, but overwrite it with what's over there if that's got something."

iramiller (approved):

iramiller (approved):

Taztingo (approved):

Issue #1687, opened 2023-09-15
Provenance Auto Completion


Taztingo:
## Summary

It would be very helpful to have auto completion in Provenance. The cobra module provides a way to generate static and dynamic auto completion

## Problem Definition

One barrier that the Provenance Blockchain faces is ease of use. It can be very hard for new users to remember what the commands are, and it can be frustrating to learn the valid types of args and flags. This problem not only affects new users, but also the development and devops teams. All these users are forced to write out entire commands and query the chain for valid arguments.

## Proposal

Introduce a command to the cli to generate the auto completion scripts. Cobra supports auto completion generation for Bash, ZSH, FISH, and PowerShell. We should at least support Bash and ZSH as they are the most popular. It is probably best to start with static auto completion and deal with dynamic creation at a later time since that involves handling asynchronous results.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

iramiller:
Not a bad idea to expand the use of the other features Cobra/Viper support.  I would recommend including Fish as well since it is nearly as popular as ZSH (more so before Apple forced the switch to ZSH in the past couple of years).

A point to consider though is that the Cosmos SDK has restructured how command line integration is done in the SDK to greatly reduce the amount of code required.  We would probably be smart to ensure any CLI updates we make are compliant with where the SDK is going.

PR #1762, opened 2023-11-22
Add auto complete.

Taztingo:
## Description

Adds Osmosis' autocomplete generation command to the Provenance binary.

closes: #1687

---

write a little note why.

### DIFF
--- a/cmd/provenanced/cmd/cmd_test.go
+++ b/cmd/provenanced/cmd/cmd_test.go
@@ -74,11 +74,7 @@ func TestGenAutoCompleteCmd(t *testing.T) {
 			rootCmd, _ := cmd.NewRootCmd(false)
 			rootCmd.SetArgs(args)
 			err := cmd.Execute(rootCmd)
-			if len(tc.err) > 0 {
-				assertions.AssertErrorValue(t, err, tc.err, "should have correct error")
-			} else {
-				require.NoError(t, err, "should not output error on command execution")
-			}
+			assertions.AssertErrorValue(t, err, tc.err, "should have the correct output value")
 		})
 	}
 }
--- a/cmd/provenanced/cmd/cmd_test.go
+++ b/cmd/provenanced/cmd/cmd_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/genutil/client/cli"
 
 	"github.com/provenance-io/provenance/cmd/provenanced/cmd"
+	"github.com/provenance-io/provenance/testutil/assertions"
 )
 
 func TestInitCmd(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cmd/provenanced/cmd/root.go
```diff
@@ -180,6 +181,39 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
...
+		DisableFlagsInUseLine: true,
+		ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+		Args:                  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			switch args[0] {
+			case "bash":
+				_ = cmd.Root().GenBashCompletion(os.Stdout)
+			case "zsh":
+				_ = cmd.Root().GenZshCompletion(os.Stdout)
+			case "fish":
```
iramiller, line 208: Good to see the best shell is included.

iramiller (dismissed):

cmd/provenanced/cmd/cmd_test.go
```diff
@@ -24,3 +25,60 @@ func TestInitCmd(t *testing.T) {
...
+			args = append(args, tc.args...)
+
+			rootCmd, _ := cmd.NewRootCmd(false)
+			rootCmd.SetArgs(args)
+			err := cmd.Execute(rootCmd)
+			if len(tc.err) > 0 {
+				assertions.AssertErrorValue(t, err, tc.err, "should have correct error")
+			} else {
+				require.NoError(t, err, "should not output error on command execution")
+			}
```
SpicyLemon, line 81:
```suggestion
			assertions.AssertErrorValue(t, err, tc.err, "enable-cli-autocomplete error")
```

The `if len(tc.err) > 0` and `else` part is built into `assertions.AssertErrorValue`.

The `msgAndArgs` part should describe how the variable being checked was made; in this case it's the `err` variable that is created by running the `enable-cli-autocomplete` command. That way, when looking at the failure message, you know more about what triggered the error (or lack of error) that you're looking at in the message.

Whether an error was or was not expected is redundant in the `msgAndArgs`. When any of `EqualError`, `NoError`, or `AssertErrorValue` fails, the failure message tells you whether either a) an error was expected and not received, b) an error was received that was different from expected, or c) an error was not expected, but was received.

Many tests have several `NoError` requirements that happen during setup, so it's really helpful for the `msgAndArgs` to indicate what was done to create the error (even if the test only has one assertion). It removes ambiguity, so it's more likely that you can identify what's going wrong by just looking at the test failure message. Usually, it's good enough to include the function name, but if that same function is being used multiple times, it's also good to include extra information to distinguish one call/check from the other. For commands, though, it's good to use the command name instead of the function being run (`Execute` or whatever) since that function isn't very descriptive.

I know the failure messages will include `TestGenAutoCompleteCmd` in the output, but when looking at a test failure message, you can't assume that `GenAutoCompleteCmd` is what was called to create the thing causing the failure. That's another reason why it's good to provide a `msgAndArgs` that describes the thing being checked.

SpicyLemon (approved):

Issue #1761, opened 2023-11-22
Refactor upgrades.go


Taztingo:
## Summary

Several chains such as Osmosis, Regen, Stride, Gaia, and Secret use a package approach for managing their upgrades. We may want to refactor our code to follow in the footsteps of this pattern.

## Problem Definition

The current approach that we have prevents us from separating upgrade functionality and dependencies. Additionally, the upgrade cannot be worked on without naming the upgrade. This creates a roadblock when work is being done and an upgrade name has not been chosen yet.

## Proposal

- Separate the logic of each upgrade.
- Differentiate an upgrade and a fork and consider if it should be factored into the design. For example, Osmosis handles them separately but Stride does not.
- Decouple the logic from the name of the upgrade.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Taztingo:
Closing since this is not needed.

PR #1779, opened 2023-12-08
Refactor upgrades

Taztingo:
## Description

Refactors upgrades.go to be a separate module structure. This allows us to easily define and manage new upgrades.

closes: #1761 

---

write a little note why.


app/upgrades/v1.17.0/upgrades.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if err = PerformUpgrade(ctx, app.Keepers()); err != nil {
+		return nil, err
+	}
+
+	return newVM, nil
+}
+
+func PerformUpgrade(ctx sdk.Context, k *keepers.AppKeepers) (err error) {
+	if err = rc1.PerformUpgrade(ctx, k); err != nil {
```
iramiller, line 33: These three RC upgrades are all independent... does it make sense to have them all part of this same "perform upgrade" step?

app/keepers/keys.go
```diff
@@ -0,0 +1,110 @@
...
+	msgfeestypes "github.com/provenance-io/provenance/x/msgfees/types"
+	nametypes "github.com/provenance-io/provenance/x/name/types"
+	oracletypes "github.com/provenance-io/provenance/x/oracle/types"
+	rewardtypes "github.com/provenance-io/provenance/x/reward/types"
+	triggertypes "github.com/provenance-io/provenance/x/trigger/types"
+
+	_ "github.com/provenance-io/provenance/client/docs/statik" // registers swagger-ui files with statik
+)
+
+func (appKeepers *AppKeepers) GenerateKeys() {
```
SpicyLemon, line 45: There's some changes to key handling in v0.50: https://github.com/cosmos/cosmos-sdk/blob/v0.50.1/UPGRADING.md#module-wiring

We should probably wait to see what that looks like before overhauling this stuff.

app/export.go
```diff
@@ -181,7 +181,7 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []str
 
 	// Iterate through validators by power descending, reset bond heights, and
 	// update bond intra-tx counters.
-	store := ctx.KVStore(app.keys[stakingtypes.StoreKey])
+	store := ctx.KVStore(app.GetKey(stakingtypes.StoreKey))
```
SpicyLemon, line 184: The keys are not meant to be exposed anywhere. The GetKey function even has a warning saying it only exists for testing purposes.

SpicyLemon:
I really don't like this. It adds a whole lot of work just to make it easier to delete an upgrade.

It's late on a Friday, though, so I'll list out more specific complaints later.

app/upgrades/v1.17.0/upgrades.go
Taztingo: So my initial thought process was that a release should be the aggregation of all release candidates and any additional logic. I tried to extract reusable logic out of the UpgradeStrategy into PerformUpgrade so that the release could use a group of rc functions without having to call each of them independently. Would it make more sense if we renamed PerformUpgrade to something else, or if there was a separate package to share functions between release candidates and releases?

app/keepers/keys.go
Taztingo: Yeah, I'll have to see how that plays out and integrate it into this.

app/export.go
Taztingo: I'm not sure how accurate that comment is, and the two lines are identical. Gaia is using it in the same fashion.
https://github.com/cosmos/gaia/blob/93016bd616d721910387a7b6e114db7c889e71c4/app/export.go#L176

Taztingo:
The goal of this was never to make it "easier to delete an upgrade", but instead make it easier to maintain. I agree that deletion is one benefit that arises from this.

Each of your comments in the hackmd should be addressed and have a follow up comment. This proposed solution includes all the benefits of the previous solution, and has the following additional benefits:

- Handlers are separated from each other allowing the module naturally encapsulate logic.
- The name is decoupled from the handler to allow implementation to continue without a name.
- We reduce the complexity of our chain by having our upgrade management follow other popular chains.
- Resolving the circular dependency allows us to simplify our app.go. App.go is already bloated and will continue to bloat as we add more keepers and network stacks.
- We can manage multiple handlers without the fear of increasing the size of upgrades.go
- The boilerplate naturally creates a template for developers to follow.
- This solution may lead to faster compilation time since we have smaller compilation units with less dependencies.

app/upgrades/v1.17.0/upgrades.go
iramiller: My point was more about how the RCs lay down on testnet… they are additive but they do not run in aggregate. We also fix things in RCs that are not needed in the final upgrade… an example would be if some piece of an upgrade in one RC is found to be incorrect and it is fixed in a subsequent one. The final handler can avoid using the incorrect approach. 

From this point of view the RCs are distinct and unique updates. As a related idea to think about, when we have many networks they are likely to need their own custom logic during an update. This is why we use color names and not version numbers. The sequence of changes (thus upgrades) is independent of version numbers.

Taztingo: Okay, should a release be able to use any of the functions defined by a RC such as RemoveInactiveValidatorDelegations? This would allow the release to mix and match steps between release candidates and use its own. That would be similar to how handlers within upgrades.go work now.

I didn't take multiple networks into consideration so I'll have to think about that some more. How are we planning to upgrade them? I see two approaches for upgrading them. One where they are in all sync and all chains upgrade whether or not they have logic. They all progress from 1.18.0 -> 1.18.1 -> 1.18.2. The other approach is applying upgrades to chains as they are needed. I.e. one chain is on 1.18.0 and another chain is on 1.18.x.

The UpgradeName still exists and uses the color just like we currently have. I know in github we still associate a minor version with a color so 1.18 is tourmaline. I thought ordering by version number would be easier to understand than color especially if we exceed 26 colors. "Which came first gold or green".

iramiller: It probably makes sense to have the upgrades grouped by network -- I don't see a good fit for version number based grouping like outlined above in any case ... it might also be useful to review the chain-registry repositories definition of releases and upgrades ... they list a set of releases with a list of compatible versions for each.

Something like `mainnet-tourmaline` would make it clear which network (in this case pair as testnet/mainnet are intended to be managed together).  This would also support something along the lines of `fes-epsilion` (for example).  

A longer term strategy will need to allow for other networks to use the `provenance` code base as a dependency which provides almost all of the code/functionality but allows an easily created binary release with their own compiled functionality and releases.

The SDK moved to module specific versioning where each module tracks the upgrades needed internally based on the current revision of it running on the network.  This allows modules to perform upgrades in a more self contained/known way and will likely be the best approach for a diverse set of networks.

SpicyLemon:
At the very least, this needs to wait until we're deeper into the `v0.50` changes since many related aspects are changing.

---

> The goal of this was never to make it "easier to delete an upgrade", but instead make it easier to maintain. I agree that deletion is one benefit that arises from this.

I know that the goal here is maintenence, but the **only** part of maintenence that gets easier with this is deleting old upgrades. A couple aspects of maintenence are mostly a wash, but most aspects become more difficult.

The maintenance needed for upgrades have three parts:
1. Creating new handlers.
2. Adding things to the next handlers.
3. Deleting old handlers.

We also often need to answer these question:
1. What all happened (will happen) with upgrade `<color>`?
2. How does the non-rc upgrade compare to each of the same-color rc upgrades?
3. What upgrade(s) was this function part of?

Creating handlers is significantly simpler with the way we've got it vs this PR. With our current setup, all you have to do is add the hard-coded entry to the map (right along-side the other entries). The trickiest part is remembering what the `appUpgrade` fields are named and the `Handler` function signature; but it's all right there in front of you. The addition is 1 to 15 new lines of code per new upgrade. With it split out (like this PR), what's needed is less obvious. You have to dig up other files/folders and copy/paste/tweak them. But when copy/paste/tweak is relied on for stuff over about 20 lines, problems start to show up where needed tweaks or previous pattersn get missed. And with the poorer signal-to-noise ratio, its more likely that even in PR, no one notices the problems. But the bottom line is that adding an entry to `handlers` is significantly easier than creating new folders and files.

If the next handler is being created in its own PR, it takes about 5 minutes to go from `main` to an ready-for-review PR. And that PR is about 1-15 lines per new handler, which is very easy to review. If it's being created as part of another PR, the new upgrade might have 20ish lines (plus a new function that it calls, which you'd have both ways). Creating all these folders/files is significantly more work to do and review. And when PRing the new handlers, currently, you can view the previous entries right in the PR since they're in the same file (directly above). With them split out into files, in order to make that comparison, you have to find and open the relevant files elsewhere. I'd estimate that the new model would take about an hour for the new PR once you get done making sure you've made all the necessary additions/deletes/tweaks to all the new files. And properly reviewing such a PR takes longer since the handy comparisons take more effort to look at.

Adding new things to the next handlers is pretty much the same either way. About the only difference is that currently, each piece has access to private fields in the App. The App stuff is modeled after the SDK's simapp which is their example of how to wire things up. Deviating from that too much might not be wise since the SDK sometimes makes poor assumptions about how everyone uses their stuff. There's definitely some discussion to be had about refactoring the App though, specially with considerations for the `depinject` stuff and other changes that come with v0.50.

Deleting old handlers is definitelly easier with this PR since you can just delete the directory. While it's nice to delete everything that was only needed for an upgrade, leftover stuff rarely causes any problems other than later confusion over why a function exists. If the leftover func is a private one in upgrades.go, though, it's pretty obvious. I'm sure there are some leftover upgrade-related funcs in some of our modules, though, that require added scruteny either way (i.e. "Is this unused keeper func for the module's API, or is it just here because a past upgrade needed it?"). With your proposed model, deleting is as easy as deleting a directory and removing their entries from the upgrades list. The proposed model increases the chances that module-specific deletable stuff gets missed though, since you don't have to look at the contents of what's being deleted in order to delete it. I'd estimate that a PR to delete a color currently takes about 30 minutes, most of which is double checking that each function in the upgrades is no longer needed and finding+deleting the related unit tests. I'd guess that the new model takes about 5 minutes for that PR (with the slightly increased chance that something gets leftover). Reviewing either PR probably takes about the same amount time, maybe slightly more for the new model since there's more files involved.

The most common question I need to answer about an upgrade is "What all happened (will happen) with upgrade <color>?". This is where our current setup really shines. The entry in the `upgrades` map tells you everything. E.g. in `rust`, we 1. ran module migrations, 2. removed inactive delegations, 3. set the account-data name record. 4. Remove the p8e memorialize contract msg fee. 5. fixes some name index entries. With this PR, that all gets obfuscated. I have to first figure out which directory is for the `rust` upgrade. That will lead me to a `constants.go` file. From there, I have to open another file that has the applicable `UpgradeStrategy` in it. From there, it starts forking with the different steps being defined across multiple files. E.g. it runs module migrations then the `PerformUpgrade` func. So I look at `PerformUpgrade` func to find that it runs the rc1 upgrade. After following that, I finally get to the rest of the information that I was looking for. In other words, this proposed model makes it significantly more difficult to answer this question. With our current setup, it takes about a minute. With the new model, that's probably closer to 10 minutes and requires taking notes externally.

Answering "How does the non-rc upgrade compare to each of the rc upgrades?" is different. With the current model, the upgrades are close in the same file, and each action is listed. So its fairly simple to compare each of the actions that the upgrades take. With the proposed model, the easy-view of an upgrade gives you which rcs were included and maybe some extra actions. So if all you're concerned about is which of the rcs are part of the main one, this is sufficient. But if you need to compare the actual actions, you then have to dig further into each individual rc. If you stop at knowing that each of the RCs is being called, its easy to miss that part of one of the RCs shouldn't be applied to the full one, though, since you're not looking at what's actually happening.

"What upgrade(s) was this func part of?" is a question that doesn't come up much. Most of the time, it comes up when looking to clean up some stuff and is more like, "Can this func be deleted?" There have been some other times that its come up, though. There's two locations that funcs can be in that require different processes when trying to answer this question: in `upgrades.go` and elsewhere. If the func-in-question is in `upgrades.go` you can just do a find-in-file for that text since its called directly from the upgrade handler as defined in the `upgrades` map with the color clearly associated. If the func-in-question is somewhere else (e.g. somewhere in a module), you have to find all uses of the function, and work backwards. Sometimes its called directly in something in `upgrades.go`, but sometimes there's a step or two before getting to `upgrades.go`. With the new model, the file containing the function tells you a little, but doesn't fully answer the question. You still need to find out a) what color is associated with that `provenanced` version? and b) was it called from anywhere else (e.g. a 2nd rc or the full upgrade)?

One last point on maintenence is that copy/paste/tweak is significantly more error prone than adding fresh stuff next to existing stuff, specially when what's being copied as whole files or directories. For example, currently, its much less likely that something is added that's missing logging. If there's another function already in the file that has logging, chances are good that the author of the new stuff will see it and remember to add it, or that one of the PR reviewers will see it and notice that it's missing from the new code. If the only examples are in other files, though, no one has that reminder, and it's up to someone to remember that its needed before three people forget and the PR is merged. And even if they do remember, and add logging, they might forget that the upgrade log messages start with capital letters and usually has ending punctuation.

So, bringing it all back together:
1. Adding upgrades is significantly easier with our current setup.
2. Adding things to upgrades is the same unless there's some private aspect of the app that hasn't been needed before, but is now needed.
3. Deleting things is harder with the current setup, but slightly more likely to be more accurate.
4. "What all happened (will happen) with upgrade <color>?" is significantly easier to answer with our current setup.
5. "How does the non-rc upgrade compare to each of the rc upgrades?" is easier with our current setup unless all you're wondering is which rcs are included in the full upgrade, then its a wash.
6. "What upgrade(s) was this func part of?" is easier in our current setup.

Adding new upgrades and removing old ones is only done once every three months or so. The amount of time spent on those is so small that it's almost not worth considering. We spend a little more time creating the upgrades, but still, not really that much, and both setups will require about the same amount of time. So, what it really comes down to is how easy it is to read and understand an upgrade. Most of my time in `upgrades.go` is spent looking at and comparing upgrades to ensure that everything that needs to be done is going to be done. That becomes significantly more difficult when all the pieces are scattered around. Plus, when its scattered around, its much more likely that standard patterns are forgotten.


Basically, our current setup is just easier, safer, and better.

---

Furthermore, an upgrade and a new version of `provenanced` are not the same thing. We can have new versions that don't have upgrades. We can have upgrades that aren't related to a new version. We can have multiple upgrades for the same version (not just rc-level). This new setup assumes that they're always the same. This new setup also obfuscates a lot of what's happening for an upgrade; in one file, the color is associated with a strategy; the strategy is defined in another file; and the actions of the strategy are defined in a separate function which can be in yet another file. You have to combine all three of those to get a clear picture of what an upgrade does.

Our current setup also makes it obvious when we have an upgrade that doesn't do anything other than force nodes to update to a new version. This situation arizes when there's a crutial bug fix, e.g. a `tx` endpoint that's incorrectly failing, or a security problem (although with security stuff, we often just cut a new version and tell nodes to switch when they can ASAP). With our current setup, that just looks like `"quicksilver-rc2": {}, // upgrade for v1.15.0-rc3`. Technically, that causes an upgrade to be registered with a handler that just logs that there's nothing to upgrade (see InstallCustomUpgradeHandlers). If we required the upgrade to be defined, even with a standard `NoOpUpgrade` that does the required logging (that we'd have to create), it's still not nearly as obvious that the upgrade does nothing.

A non-rc upgrade is rarely the combination of all the rc upgrades. There are often things that need to be done to mainnet, but not testnet, and vice versa. And sometimes, in testnet, we have to redo something after tweaking it. The only way to know if an upgrade does what it needs to do is with outside knowledge. There's just no getting away from that. With the current setup, it's all right in front of you though, so even without that knowledge, you can identify, "Hey, this func is being called in an rc, but not in the full. I should ask if that's the behavior we want." With the actions of an upgrade spread out all over the place, it takes significantly more work to get to the point where you can have that question, enough more work that it's significantly less likely to happen and definitely won't happen by accident.

I also want to point out that currently, `app/upgrades.go` is 1 file with 351 lines, and all this new `app/upgrades/...` stuff is 17 files with 667 lines.

---

> This proposed solution [...] has the following additional benefits:
> 
> * Handlers are separated from each other allowing the module naturally encapsulate logic.

When modules have upgrade stuff, that logic is usually contained in the module's code with just a func in `upgrades.go` to add context logging. Alternatively, the module might go with the standard upgrade route by bumping the consensus version and defining the module migration appropriatly, in which case it'd be run with the call to `ModuleManager().RunMigrations(...)` and nothing special gets added to `upgrades.go` or a handler.

If you're talking about encapsulating **all** the logic for an upgrade, I don't see any benefit, but I do see some detriments. Encapsulation usually makes it harder to interact with a thing, and too often, impossible to do what you later need to do (i.e. the original author incorrectly assumed no one would ever need access to a thing, and made it private).

> * The name is decoupled from the handler to allow implementation to continue without a name.

The name **is** the handler. The handler **is** the name. If you've got something that you know needs to be added to an upgrade, and that upgrade doesn't yet exist, you just create the new upgrade. It has to be done at some point anyway, and its better to add it before or with the first needed thing. We're usually pretty quick to pick a color, and even if we weren't, a placeholder is just fine. There's never going to be a case where we are in so much of a rush to get upgrade code merged that we can't stop to pick the color first. It's also easier and safer to add things to an upgrade as the need arises. If creating the upgrade is put off for later, then the person putting it together later needs to know what already exists and needs to be included. When creating it with the first need, they just need to know what they already know.

> * We reduce the complexity of our chain by having our upgrade management follow other popular chains.

I don't really see this as a benefit. Upgrades aren't portable across chains. Even if it's a `provenanced` chain (e.g. in a private network), the upgrade might not be portable.

> * Resolving the circular dependency allows us to simplify our app.go. App.go is already bloated and will continue to bloat as we add more keepers and network stacks.

There's a bunch of stuff in v0.50 that helps with this bloat. It's probably better to diverge as little as possible from what the sdk suggests, though, which is where we're at now.

> * We can manage multiple handlers without the fear of increasing the size of upgrades.go

It's rare that upgrades.go gets over 1000 lines. We don't want to (and sometimes can't) keep old upgrade code around. As upgrades are added, others are removed. I don't think this is anything to be concerned about.

> * The boilerplate naturally creates a template for developers to follow.

We already naturally have a template with a smaller footprint and which is located in a handier place than this new model.
There's two ways upgrades will be added in the new model: 1) An existing upgrade dir is copied and tweaked. 2) The dev manually creates the files and creates each element modeled after either a previous upgrade, or what they assume its supposed to be like. Both ways, there's a lot of room for errors to be introduced, that aren't nearly as much of a concern with the current model; there's far less currently being modeled. Furthermore, unit tests are almost never copied with stuff like this; they're almost always added later based on what the engineer is thinking about. There's some tricky things to set up for these tests, though, and getting the needed info (e.g. what was logged) is tricky too. Since you can't import test code from other packages, we'd either need a package (that gets built and included with `build`) or there'd be a lot of reinventing. But the existing `UpgradeTestSuite` is the result of many iterations and it promotes correct testing patterns for the upgrades. If each upgrade does their own reinventing, we'll just fall back into bad (or problematic) patterns.

> * This solution may lead to faster compilation time since we have smaller compilation units with less dependencies.

I'd be surprised if this were the case, even for unit tests (which limit the build to just whats needed). Even with mocks, the unit tests will have to pull in the `app/app` package, which will pull in pretty much everything anyway.

> * As you mentioned easier to add and delete upgrades.

I said it was easier to delete upgrades. I very much did **not** say it was easier to add them.

---

I have a whole slew of other issues with this PR, but will hold off on bringing them up until I'm convinced that this change is worthwhile.
At the very least, this needs to wait until we're deeper into the `v0.50` changes since many related aspects are changing.

app/upgrades/v1.17.0/upgrades.go
SpicyLemon: > So my initial thought process was that a release should be the aggregation of all release candidates and any additional logic.

That is very much not the case.

> I know in github we still associate a minor version with a color so 1.18 is tourmaline.

Nope. The `tourmaline` upgrade will do a bunch of stuff to make `mainnet` compatible with v1.18 of `provenanced`. As part of that, nodes will need to update their version of `provenanced` to `v1.18`. But `tourmaline` and `v1.18` are not the same thing. There can be upgrades without a version. There can be versions without an upgrade. An upgrade can do more than just bump the version. An upgrade doesn't have to bump the version. While we usually can get away with associating a 1-1 color and version, that isn't always the case.

---

The `RemoveInactiveValidatorDelegations` function should be defined alongside `RunModuleMigrations` since we want to remove inactive validator delegations regularly.

---

> If we aren't expecting overlap or aren't concerned with that then I can restructure it

That structure (with the colors instead of versions) is a significant improvement.

I feel like its best to not have one color reference a func in another. If we find ourselves wanting this, its probably better to copy the func to either the new color (if we don't think we'll want it again) or the common area (if we expect to want it in another future upgrade too). If one color references another, it'll almost always be the new one referencing an old function, which means that it'll break once we delete the old version.

---

> Would it make more sense if we renamed `PerformUpgrade` to something else, or if there was a separate package to share functions between release candidates and releases?

Since the `mainnet` upgrade isn't just the sum of rc upgrades, I don't find the `PerformUpgrade` function at all useful. The **only** thing useful in there would be each of the things that are done.

What would probably be even better is to focus on the `mainnet` upgrade. Put all of the sub-functions in the same file as that upgrade so that when we read it, we see exactly what actions it takes, and those actions are defined in functions below. Then, with the rc upgrades, call those functions directly too.

Basically, our first thought should usually be, "This needs to be in an upgrade, so I'll put it with the production-level upgrade." with a folloup of, "Okay, this also needs to be in our next rc." For something that shouldn't be applied to the production-level upgrade, that can go in the same file as the related rc strategy.

Either way, we still want comments explaining why something is in an rc but not full or vice versa.

app/export.go
Taztingo: > Our keys have never been used this way.
We have never extracted the keys out of app.go. That is also a function on SimApp.

Taztingo:
I'm going to hold off on this PR until we understand what we want to do with multiple networks. I'm not convinced the current approach is the best solution. I have addressed the remainder of the comments so that we understand both points of view if we ever decide to come back to this.

> Creating handlers is significantly simpler with the way we've got it vs this PR. With our current setup, all you have to do is add the hard-coded entry to the map (right along-side the other entries). The trickiest part is remembering what the appUpgrade fields are named and the Handler function signature; but it's all right there in front of you. The addition is 1 to 15 new lines of code per new upgrade. With it split out (like this PR), what's needed is less obvious. You have to dig up other files/folders and copy/paste/tweak them. But when copy/paste/tweak is relied on for stuff over about 20 lines, problems start to show up where needed tweaks or previous pattersn get missed. And with the poorer signal-to-noise ratio, its more likely that even in PR, no one notices the problems. But the bottom line is that adding an entry to handlers is significantly easier than creating new folders and files.

I agree that there is a little bit more involved to add a handler. This is typical when you break down a file into many.

>  the next handler is being created in its own PR, it takes about 5 minutes to go from main to an ready-for-review PR. And that PR is about 1-15 lines per new handler, which is very easy to review. If it's being created as part of another PR, the new upgrade might have 20ish lines (plus a new function that it calls, which you'd have both ways). Creating all these folders/files is significantly more work to do and review. And when PRing the new handlers, currently, you can view the previous entries right in the PR since they're in the same file (directly above). With them split out into files, in order to make that comparison, you have to find and open the relevant files elsewhere. I'd estimate that the new model would take about an hour for the new PR once you get done making sure you've made all the necessary additions/deletes/tweaks to all the new files. And properly reviewing such a PR takes longer since the handy comparisons take more effort to look at.

I don't agree that this is always true especially with a pattern in place.

> Deleting old handlers is definitelly easier with this PR since you can just delete the directory. While it's nice to delete everything that was only needed for an upgrade, leftover stuff rarely causes any problems other than later confusion over why a function exists. If the leftover func is a private one in upgrades.go, though, it's pretty obvious. I'm sure there are some leftover upgrade-related funcs in some of our modules, though, that require added scruteny either way (i.e. "Is this unused keeper func for the module's API, or is it just here because a past upgrade needed it?"). With your proposed model, deleting is as easy as deleting a directory and removing their entries from the upgrades list. The proposed model increases the chances that module-specific deletable stuff gets missed though, since you don't have to look at the contents of what's being deleted in order to delete it. I'd estimate that a PR to delete a color currently takes about 30 minutes, most of which is double checking that each function in the upgrades is no longer needed and finding+deleting the related unit tests. I'd guess that the new model takes about 5 minutes for that PR (with the slightly increased chance that something gets leftover). Reviewing either PR probably takes about the same amount time, maybe slightly more for the new model since there's more files involved.

Agreed it's easier.

> The most common question I need to answer about an upgrade is "What all happened (will happen) with upgrade ?". This is where our current setup really shines. The entry in the upgrades map tells you everything. E.g. in rust, we 1. ran module migrations, 2. removed inactive delegations, 3. set the account-data name record. 4. Remove the p8e memorialize contract msg fee. 5. fixes some name index entries. With this PR, that all gets obfuscated. I have to first figure out which directory is for the rust upgrade. That will lead me to a constants.go file. From there, I have to open another file that has the applicable UpgradeStrategy in it. From there, it starts forking with the different steps being defined across multiple files. E.g. it runs module migrations then the PerformUpgrade func. So I look at PerformUpgrade func to find that it runs the rc1 upgrade. After following that, I finally get to the rest of the information that I was looking for. In other words, this proposed model makes it significantly more difficult to answer this question. With our current setup, it takes about a minute. With the new model, that's probably closer to 10 minutes and requires taking notes externally.

I don't see this as an issue, and I think it's up to preference. You can easily compare two upgrades by putting their files side by side, and it depends on workspace. You could argue that it's easier to even see 3 files side by side.

> "What upgrade(s) was this func part of?" is a question that doesn't come up much. Most of the time, it comes up when looking to clean up some stuff and is more like, "Can this func be deleted?" There have been some other times that its come up, though. There's two locations that funcs can be in that require different processes when trying to answer this question: in upgrades.go and elsewhere. If the func-in-question is in upgrades.go you can just do a find-in-file for that text since its called directly from the upgrade handler as defined in the upgrades map with the color clearly associated. If the func-in-question is somewhere else (e.g. somewhere in a module), you have to find all uses of the function, and work backwards. Sometimes its called directly in something in upgrades.go, but sometimes there's a step or two before getting to upgrades.go. With the new model, the file containing the function tells you a little, but doesn't fully answer the question. You still need to find out a) what color is associated with that provenanced version? and b) was it called from anywhere else (e.g. a 2nd rc or the full upgrade)?

This question isn't a concern in this pattern because they are housed in their respective rc.

> One last point on maintenence is that copy/paste/tweak is significantly more error prone than adding fresh stuff next to existing stuff, specially when what's being copied as whole files or directories. For example, currently, its much less likely that something is added that's missing logging. If there's another function already in the file that has logging, chances are good that the author of the new stuff will see it and remember to add it, or that one of the PR reviewers will see it and notice that it's missing from the new code. If the only examples are in other files, though, no one has that reminder, and it's up to someone to remember that its needed before three people forget and the PR is merged. And even if they do remember, and add logging, they might forget that the upgrade log messages start with capital letters and usually has ending punctuation.

I agree that it is more error prone. I don't agree about the pattern because it's applicable to both designs, and depends on the user and the reviewers. As a counter point we have a function defined in upgrades.go that does not follow this pattern.

> Adding upgrades is significantly easier with our current setup.
Agreed 

> Adding things to upgrades is the same unless there's some private aspect of the app that hasn't been needed before, but is now needed.
Agreed

> Deleting things is harder with the current setup, but slightly more likely to be more accurate.
Agreed

> "What all happened (will happen) with upgrade ?" is significantly easier to answer with our current setup.
This is mainly subjective and up to user preference.

> "How does the non-rc upgrade compare to each of the rc upgrades?" is easier with our current setup unless all you're wondering is which rcs are included in the full upgrade, then its a wash.
This also depends on user preference.

> "What upgrade(s) was this func part of?" is easier in our current setup.
This is not applicable to this design...

> Adding new upgrades and removing old ones is only done once every three months or so. The amount of time spent on those is so small that it's almost not worth considering. We spend a little more time creating the upgrades, but still, not really that much, and both setups will require about the same amount of time. So, what it really comes down to is how easy it is to read and understand an upgrade. Most of my time in upgrades.go is spent looking at and comparing upgrades to ensure that everything that needs to be done is going to be done. That becomes significantly more difficult when all the pieces are scattered around. Plus, when its scattered around, its much more likely that standard patterns are forgotten.

This really depends on preference. It could potentially be easier with it split up because you can upgrades next to each other.

> Basically, our current setup is just easier, safer, and better.
This is a **very big jump** in conclusion.

> Our current setup also makes it obvious when we have an upgrade that doesn't do anything other than force nodes to update to a new version. This situation arizes when there's a crutial bug fix, e.g. a tx endpoint that's incorrectly failing, or a security problem (although with security stuff, we often just cut a new version and tell nodes to switch when they can ASAP). With our current setup, that just looks like "quicksilver-rc2": {}, // upgrade for v1.15.0-rc3. Technically, that causes an upgrade to be registered with a handler that just logs that there's nothing to upgrade (see InstallCustomUpgradeHandlers). If we required the upgrade to be defined, even with a standard NoOpUpgrade that does the required logging (that we'd have to create), it's still not nearly as obvious that the upgrade does nothing.

I'm not sure if I agree this and believe it takes the same amount of effort. Navigating to a file and viewing the contents of the upgrade. May actually be more effort to find it in upgrades.go.

> I also want to point out that currently, app/upgrades.go is 1 file with 351 lines, and all this new app/upgrades/... stuff is 17 files with 667 lines.

That's true, but this is also true in other cases where you break down a file into other modules. Imports need to be redefined... The purpose is to have readable bite size chunks.

> When modules have upgrade stuff, that logic is usually contained in the module's code with just a func in upgrades.go to add context logging. Alternatively, the module might go with the standard upgrade route by bumping the consensus version and defining the module migration appropriatly, in which case it'd be run with the call to ModuleManager().RunMigrations(...) and nothing special gets added to upgrades.go or a handler.
If you're talking about encapsulating all the logic for an upgrade, I don't see any benefit, but I do see some detriments. Encapsulation usually makes it harder to interact with a thing, and too often, impossible to do what you later need to do (i.e. the original author incorrectly assumed no one would ever need access to a thing, and made it private).

Yes, I am referring to the upgrade. The benefit that I am referring to is that we are grouping logic together without inappropriately mixing it. A color will define exactly what it needs. No more and no less.

> The name is the handler. The handler is the name. If you've got something that you know needs to be added to an upgrade, and that upgrade doesn't yet exist, you just create the new upgrade. It has to be done at some point anyway, and its better to add it before or with the first needed thing. We're usually pretty quick to pick a color, and even if we weren't, a placeholder is just fine. There's never going to be a case where we are in so much of a rush to get upgrade code merged that we can't stop to pick the color first. It's also easier and safer to add things to an upgrade as the need arises. If creating the upgrade is put off for later, then the person putting it together later needs to know what already exists and needs to be included. When creating it with the first need, they just need to know what they already know.

I disagree, the name is not the handler. The benefit of this approach was users could contribute to the handler without deciding upon a name. The name could be decided upon at a later time.

> I don't really see this as a benefit. Upgrades aren't portable across chains. Even if it's a provenanced chain (e.g. in a private network), the upgrade might not be portable.

The benefit is to encourage development and make it easier for other developers to contribute.

> We already naturally have a template with a smaller footprint and which is located in a handier place than this new model. There's two ways upgrades will be added in the new model: 1) An existing upgrade dir is copied and tweaked. 2) The dev manually creates the files and creates each element modeled after either a previous upgrade, or what they assume its supposed to be like. Both ways, there's a lot of room for errors to be introduced, that aren't nearly as much of a concern with the current model; there's far less currently being modeled. Furthermore, unit tests are almost never copied with stuff like this; they're almost always added later based on what the engineer is thinking about. There's some tricky things to set up for these tests, though, and getting the needed info (e.g. what was logged) is tricky too. Since you can't import test code from other packages, we'd either need a package (that gets built and included with build) or there'd be a lot of reinventing. But the existing UpgradeTestSuite is the result of many iterations and it promotes correct testing patterns for the upgrades. If each upgrade does their own reinventing, we'll just fall back into bad (or problematic) patterns.

Agreed both approaches have a template, and some thought will need to be put into testing.

> I'd be surprised if this were the case, even for unit tests (which limit the build to just whats needed). Even with mocks, the unit tests will have to pull in the app/app package, which will pull in pretty much everything anyway.

This point is probably off topic and would probably need some tests to verify. I brought this up from prior experience with c++, and it may not be true with golang's compiler. Even if it were true it may depend on configurations.

Issue #1772, opened 2023-11-30
Bump sdk to v0.50.1 in main-v0.50 


SpicyLemon:
Create a PR to `main-v0.50` that bumps the SDK to `v0.50.x` (not our fork) and make it compile and pass tests. Only make fixes for simple things, and favor commenting things out and skipping tests (adding `TODO[1760]`s liberally).

SpicyLemon:
This was done via #1787

Issue #1760, opened 2023-11-22
Epic: SDK v0.50


SpicyLemon:
## Summary

This epic is for identifying and coordinating the work for bumping Cosmos SDK to `v0.50.x`.

## Overview

We will need to make changes across at least two repos, this one and our fork of the SDK. We'll need to do a few key things that then allow us to do some work in parallel. In general, we will be commenting a bunch of stuff out and re-adding it all back in.

As we disable stuff, mark things with todo comments that reference this issue, e.g. `// TODO[1760]: Add the marker module back in.` These comments should give us a nice list of work to do if we do something like `grep -rF 'TODO[1760]' *` 

When you pick up a checklist item, make an issue from it and assign it to yourself (so that we know someone's working on it). Use the "convert to issue" thing (right side when you hover over a checklist item) for this. It's okay to have a PR that checks off multiple things.

## The Plan

1. In this repo, use a `main-v0.50` branch that we will create PRs to. We'll start it off by switching the SDK to `v0.50.x` and commenting out a lot of stuff, then we'll work on fixing everything and putting it all back together.
2. In our SDK fork, use a `release-pio/v0.50.x` branch that starts with a `v0.50.x` tag and re-applies our customizations.
3. We can still make changes to `main`, but those will eventually need to be backported to `main-v0.50`.
4. Once or SDK fork is ready, tag our version and switch us over to that in the `main-v0.50` branch.
5. Once `main-v0.50` is done, merge it back into `main`.

## Provenance Repo Plan

These should be done in this order.

- [x] 1. Identify and Finish any key changes needed that aren't related to `v0.50.x`.
- [x] #1771
- [x] #1772
- [x] 4. From the resulting `TODO[1760]`s compile a general list of things to work on and add entries to the TODOs checklist.
- [x] 5. Finish some TODOs in the checklist by making PRs to `main-v0.50`.
- [x] 6. Backport any PRs in `main` that aren't yet in `main-v0.50`.
- [x] 7. Merge `main-v0.50` back into `main`.
- [x] 8. Finish all TODOs in the checklist by making PRs to `main`.

## Provenance Repo Pre-Work

These are the key changes that need to be done before we start the `v0.50` work in this repo.

- [x] #1498 

## Provenance Repo TODOs

These can (mostly) be done in any order and can be grouped together for PRs as needed. An entry that starts with a tag has related TODOs in the repo.

Example command for finding TODO entries for a tag:
```
[...]

PR #1787, opened 2023-12-16
Initial v0.50 bump

SpicyLemon:
## Description

This PR bumps the SDK to `v0.50.1` (from a `v0.46` based version from our fork).

As part of this, go is bumped to `v1.21` (from `v1.20`) too.

Lots and lots of stuff is just commented out for us to work on later.

closes: #1772

---

write a little note why.


app/app.go
```diff
@@ -436,79 +435,95 @@ func New(
...
 	// set the BaseApp's parameter store
-	bApp.SetParamStore(app.ParamsKeeper.Subspace(baseapp.Paramspace).WithKeyTable(paramstypes.ConsensusParamsKeyTable()))
+	// bApp.SetParamStore(app.ParamsKeeper.Subspace(baseapp.Paramspace).WithKeyTable(paramstypes.ConsensusParamsKeyTable())) // TODO[1760]: params
 
 	// add capability keeper and ScopeToModule for ibc module
 	app.CapabilityKeeper = capabilitykeeper.NewKeeper(appCodec, keys[capabilitytypes.StoreKey], memKeys[capabilitytypes.MemStoreKey])
 
 	// grant capabilities for the ibc and ibc-transfer modules
-	scopedIBCKeeper := app.CapabilityKeeper.ScopeToModule(ibchost.ModuleName)
+	scopedIBCKeeper := app.CapabilityKeeper.ScopeToModule("ibc") // TODO[1760]: ibc-host: was ibchost.ModuleName
```
Taztingo, line 444: This should be `ibcexported.ModuleName`

app/app.go
```diff
@@ -1122,17 +1169,17 @@ func (app *App) GetTxConfig() client.TxConfig {
 func (app *App) Name() string { return app.BaseApp.Name() }
```
Taztingo, line 1170: Pretty sure we need the PreBlocker. "When using (legacy) application wiring, the following must be added to app.go"

```
func (app *App) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlock) (*sdk.ResponsePreBlock, error) {
  return app.mm.PreBlock(ctx, req)
}
```

app/app.go
```diff
@@ -1017,34 +1061,37 @@ func New(
 	app.SetBeginBlocker(app.BeginBlocker)
```
Taztingo, line 1061: Probably need SetPreBlocker as well "When using (legacy) application wiring, the following must be added to app.go".

`app.SetPreBlocker(app.PreBlocker)`

app/app.go
```diff
@@ -797,7 +842,7 @@ func New(
 		slashingtypes.ModuleName,
 		evidencetypes.ModuleName,
 		stakingtypes.ModuleName,
-		ibchost.ModuleName,
```
Taztingo, line 800: Pretty sure we need to remove upgradetypes.ModuleName from BeginBlockers and add it to PreBlockers. "When using (legacy) application wiring, the following must be added to app.go"

```
app.mm.SetOrderPreBlockers(
  upgradetypes.ModuleName,
)
```

cmd/provenanced/cmd/root.go
Taztingo, line 1: Do we need to add textual signing? It's the default for app v2.

iramiller: I feel like we should add it... being able to support signing over messages that convey what's in them seems important.

helpers/utils.go
Taztingo, line 1: I strongly suggest moving these types of functions into internal if we can, is there a reason they can't be? We should also probably break this utils down now before it becomes bigger. The name `utils` typically indicates a `code smell` since it becomes a dumping ground for random functions and breaks SOLID. I am suggesting this now rather than later because this may be heavily touched when we start splitting up the work.

app/app.go
```diff
@@ -194,13 +193,8 @@ var (
 		mint.AppModuleBasic{},
 		distr.AppModuleBasic{},
 		gov.NewAppModuleBasic(append(
-			wasmclient.ProposalHandlers,
+			[]govclient.ProposalHandler{},
```
Taztingo, line 196: Is ModuleBasics still needed?

> Previously, the ModuleBasics was a global variable that was used to register all modules' AppModuleBasic implementation. The global variable has been removed and the basic module manager can be now created from the module manager.

app/app.go
SpicyLemon: It's in the list of TODOs. We can do that later with the rest of the ibc-host stuff.

app/app.go
```diff
@@ -436,79 +435,95 @@ func New(
...
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
-	)
+	// TODO[1760]: account: Put back this call to NewAccountKeeper with fixed arguments.
+	// app.AccountKeeper = authkeeper.NewAccountKeeper(
+	// 	appCodec, keys[authtypes.StoreKey], app.GetSubspace(authtypes.ModuleName),
+	// 	authtypes.ProtoBaseAccount, maccPerms, AccountAddressPrefix,
+	// )
+
+	// TODO[1760]: bank: Put back this call to NewBaseKeeper with fixed arguments.
+	// app.BankKeeper = bankkeeper.NewBaseKeeper(
```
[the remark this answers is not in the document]
SpicyLemon: I've got it in the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

app/app.go
SpicyLemon: I've got it in the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

app/app.go
SpicyLemon: I've got it in the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

app/app.go
SpicyLemon: I've got it in the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

cmd/provenanced/cmd/root.go
SpicyLemon: I've got it in the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

app/app.go
SpicyLemon: I'll add that to the list of TODOs in the [epic](https://github.com/provenance-io/provenance/issues/1760).

helpers/utils.go
SpicyLemon: No, no reason they can't be in internal. I'm just not a fan of internal, so I never think to put things there. But I guess, since our stuff isn't designed at all to be imported elsewhere, they can go in there just fine.

But sometimes there's just stuff that doesn't have a better place. Or rather, the better place is somewhere that can't be modified.

I'll move the `ExitCodeError` into the command since that's the only thing that cares about it.

`MustValAddressFromBech32`, `GetOperatorAddr`, `MustGetOperatorAddr`, and `ValidateBasic` are best put in the SDK. But since that's a pain and we're trying to reduce our modifications on our fork, they need to live somewhere. I could be more specific an put the first in a file called `addr_utils.go` but then `ValidateBasic` would just be alone in a file named something else. But having them all together in a file doesn't make that file at all unmanageable, so I don't see any reason to split them up. I'll just put them all in an `sdk.go` file inside `internal/helpers`.

Not sure what to do about `Keys` though. It's in go's experimental `"golang.org/x/exp/maps"` library, but not in the official `"maps"` one, and we'd rather not allow the experimental stuff to be imported. So it'll need a home too. I guess maybe in a `maps.go` file. I'm not really fan of a single file for a 7 lines of code (and 3 lines of comments) though. But I guess it has to be done.

Taztingo (approved):

iramiller (approved):
This looks like it is on the right track to me 💪

Issue #1749, opened 2023-11-15
Support setting a NAV attribute on metadata scopes 


iramiller:
## Summary

Add the ability to set a NAV on a given scope to indicate its value for TVL

## Problem Definition

With the release of 1.17 markers now have a new TVL component that allows users to set information about the prices/value of a given token.  Expanding on this capability we need to add a similar value to the metadata/scopes so that it is easy to declare a given value of an asset when it is created (USD).

## Proposal

Add an attribute to the scope concept (doesn't have to be directly in the scope proto) which allows a user to assign a USD NAV value similar to the one set for markers.  Unlike the markers there will not be an exchange function that updates these values until perhaps at a future date the exchange system is expanded to allow trading of scopes directly (instead of requiring them to be tied to a marker).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

nullpointer0x00:
Are the scope NAVs only supposed to be in USD?  Marker module supports multiple NAVs of different denoms.  If we want to do that here, I would suggest not having a proto directly on the scope.

iramiller:
To start with I don't know that we will need the alternate pricing ... but that is certainly coming, if not in the first revision then in a subsequent one.

PR #1791, opened 2024-01-02
Support setting a nav attribute on metadata scopes

nullpointer0x00:
## Description

closes: #1749

---

write a little note why.


Taztingo (dismissed):

proto/provenance/metadata/v1/scope.proto
```diff
@@ -252,3 +253,13 @@ message AuditFields {
   // an optional message associated with the creation/update event
   string message = 6 [(gogoproto.moretags) = "yaml:\"message,omitempty\""];
 }
+
+// NetAssetValue defines a scope's net asset value
+message NetAssetValue {
+  // price is the complete value of the asset's volume
+  cosmos.base.v1beta1.Coin price = 1 [(gogoproto.nullable) = false];
+  // volume is the number of tokens of that the scope was purchased for the price
+  uint64 volume = 2;
```
iramiller, line 262: Won't this always be 1? I suppose having this is good for consistency of format.

proto/provenance/metadata/v1/tx.proto
```diff
@@ -107,6 +110,11 @@ message MsgWriteScopeRequest {
   // If there is a value in scope.specification_id that is different from the one created from this uuid, an error is
   // returned.
   string spec_uuid = 4 [(gogoproto.moretags) = "yaml:\"spec_uuid\""];
+
+  // usd_mills value of scope in usd mills (1234 = $1.234) used for net asset value
+  uint64 usd_mills = 12;
+  // volume value of scope used for net asset value
+  uint64 volume = 13;
```
iramiller, line 117: Not sure we need this parameter because it should always be one as a non-fungible token.

proto/provenance/metadata/v1/scope.proto
nullpointer0x00: Yes, you are right.  I'm on the fence about removing it now.  Leaving it will keep it identical to the marker's NAV structure, but it isn't serving any purpose and taking up space if it is always 1.

proto/provenance/metadata/v1/tx.proto
nullpointer0x00: Good point.  I will remove this and all associated code along with it.

proto/provenance/metadata/v1/scope.proto
nullpointer0x00: I slept on it and I think we should just remove it.  Leaving it just adds something to be managed in the code that should always be 1.  Can cause confusion down the road as well.

x/metadata/client/cli/cli_test.go
```diff
@@ -3919,11 +3880,11 @@ func (s *IntegrationCLITestSuite) TestGetCmdAddNetAssetValues() {
 		{
 			name:   "invalid net asset string",
 			args:   argsWStdFlags(scopeID, "invalid"),
-			expErr: "invalid net asset value, expected coin,volume",
+			expErr: "invalid coin invalid",
```
Taztingo, line 3883: This test is a little bit hard to decipher since the supplied name is actually in the error. Can we adjust either the error message or the denom?

x/metadata/client/cli/tx.go
```diff
@@ -1352,27 +1339,20 @@ func validateAccAddress(addr, argName string) (string, error) {
...
-		coin, err := sdk.ParseCoinNormalized(parts[0])
-		if err != nil {
-			return []types.NetAssetValue{}, fmt.Errorf("invalid coin %s", parts[0])
-		}
-		volume, err := strconv.ParseUint(parts[1], 10, 64)
+
+		coin, err := sdk.ParseCoinNormalized(nav)
 		if err != nil {
-			return []types.NetAssetValue{}, fmt.Errorf("invalid volume %s", parts[1])
+			return []types.NetAssetValue{}, fmt.Errorf("invalid coin %s", nav)
```
Taztingo, line 1353: One way to help clear this up is by including what the coin is invalid for such as:

"invalid net asset value coin %s".

It doesn't have to be that, but just improves readability.

x/metadata/client/cli/cli_test.go
Taztingo: The comment below should help resolve this if you still want to use the name "invalid".

iramiller:
> I do not have an event that gets emitted when a NAV is removed (add does). Removal of a NAV only happens when a scope is removed. Scope removal does emitted an event...I would think that should be enough. Thoughts? If we do add this, I will have to add it to the marker module too.

I don't see the need for a NAV removal event ... the only thing that matters is when a new instance of the data is available IMO.  In general lately I am anti-event given all of the negative aspects of data bloat.

Taztingo (approved):

nullpointer0x00:
>  In general lately I am anti-event given all of the negative aspects of data bloat.

@iramiller I'm feeling the same way.  A removal of the scope event should be enough to indicate that everything related was removed.

Taztingo (approved):

iramiller (approved):

Issue #1569, opened 2023-05-31
Investigate synchronized missed blocks.


SpicyLemon:
## Summary of Bug

According to [ping.pub](https://ping.pub/provenance/uptime) (the click on "blocks"), there are blocks that several validators will simultaneously miss.

<img width="914" alt="Screenshot 2023-05-31 at 12 56 13" src="https://github.com/provenance-io/provenance/assets/6164208/34068192-9b2a-4d21-a9cf-e9d0d6451fa0">

In that image, you can see that several validators have red squares for the same blocks. The right-most synchronized miss is 11282380 and the image also indicates an issue with 11282370, and 11282360.

The block info query (e.g. `provenanced q block 11282381`) has a `.block.last_commit.signatures` section that indicates which validators signed off on the previous block.

We need to get a better picture of these synchronized missed blocks and try to identify why they might be happening.

1. Put something together that will identify synchronized missed blocks and the validators involved with each.
2. Identify validators often involved, sometimes involved, and never involved in these misses.
3. Identify the proposers of these misses to see if there's a pattern.
4. Investigate some of the identified blocks to see if there are any similarities.
5. Identify next steps for either further investigation or fixes.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Here are some notes I have so far.

* `provenanced q block n` -> `.block.header.proposer_address` gives the block proposer.
* `provenanced q block n+1` -> `.block.last_commit.signatures` gives the validator signature info for that block. The addresses here seem to be hex strings of the address bytes.
* `provenanced q tendermint-validator-set` contains bech32 address strings for the validators with HRP `pbvalcons` (e.g. `pbvalcons1ryz97ullxwphrtdvkfm7m88havn6scvumdf227`).
* `provenanced q staking validators` contains more info on the validators (e.g. descriptions), and has bech32 address strings for the validators with HRP `pbvaloper` (e.g. `pbvaloper1hxd75vuud3gddkfpglkzzsvpjwrz9zf05xu4ss`).

I wrote https://github.com/SpicyLemon/SpicyLemon/tree/master/go_fun/bech32 to be to convert hex/base64/bech32 strings and will hopefully be useful for associating the `q staking validators` info with `q block` last block signatures.

[...]

PR #1792, opened 2024-01-04
Adds validation checks related to the missed blocks

Taztingo:
## Description

Adds additional checks that are ran before the StoreLoader.

closes: #1569

---

write a little note why.


app/store_loader.go
```diff
@@ -0,0 +1,54 @@
...
+func PruningWrapper(logger log.Logger, appOpts servertypes.AppOptions, storeLoader baseapp.StoreLoader) baseapp.StoreLoader {
+	return WrapStoreLoader(func(ms sdk.CommitMultiStore, sl baseapp.StoreLoader) error {
+		const MAX_PRUNING_INTERVAL = 100
+
+		// No error checking is needed because we leave that up to the sdk.
+		pruningString, _ := appOpts.Get("pruning-interval").(string)
+		interval, _ := strconv.ParseUint(pruningString, 10, 64)
+
+		if interval > MAX_PRUNING_INTERVAL {
+			logger.Error(fmt.Sprintf("pruning-interval %s IS INVALID AND CANNOT EXCEED %d", pruningString, MAX_PRUNING_INTERVAL))
```
iramiller, line 49: I would say that exceeding the level is not recommended because it can lead to missed blocks (but only for validators)

app/store_loader.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// PruningWrapper creates a new StoreLoader that first validates the pruning settings before calling the provided StoreLoader.
+func PruningWrapper(logger log.Logger, appOpts servertypes.AppOptions, storeLoader baseapp.StoreLoader) baseapp.StoreLoader {
+	return WrapStoreLoader(func(ms sdk.CommitMultiStore, sl baseapp.StoreLoader) error {
+		const MaxPruningInterval = 13
+
+		// No error checking is needed because we leave that up to the sdk.
+		pruningString, _ := appOpts.Get("pruning-interval").(string)
+		interval, _ := strconv.ParseUint(pruningString, 10, 64)
```
SpicyLemon, line 41: Usually we use the `"github.com/spf13/cast"` library when getting things from `appOpts`:

```suggestion
		interval := cast.ToUint64(appOpts.Get("pruning-interval"))
```

`cast.ToUint` should be just fine too if that makes use of `interval` easier. I just have `ToUint64` there because that's what the SDK uses to get the pruning interval from `appOpts` (it's in `server/pruning.go` in the SDK).

app/.db/000001.log
SpicyLemon, line 1: This `app/.db` directory should not exist. I'm guessing it got created from a unit test that didn't correctly use a temporary directory instead. Haven't gotten down there yet to see if that's been corrected yet.

app/store_loader_test.go
```diff
@@ -0,0 +1,129 @@
...
+			storeLoader: createMockStoreLoader(),
+			wrapper:     createMockFlipWrapper(&flag),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			storeLoader := WrapStoreLoader(tc.wrapper, tc.storeLoader)
+			db := dbm.GoLevelDB{}
+			ms := rootmulti.NewStore(&db, nil)
```
SpicyLemon, line 45: When creating these test dbs/stores, you need to use a temporary directory provided by `t.TempDir()`. If you intend to re-use the same db/store across multiple tests you'd use the `t` outside this `t.Run(...)`. Alternatively, maybe use a `MemDB`. I'm pretty sure that doesn't create any dirs/files on the system.

The problem is that it ends up using some default location which is sometimes even the default `--home` used by `provenanced` which can then mess other things up on the system running the tests. It can also cause tests to fail because that already exists but is not what the test is expecting. Judging by the `app/.db` dir in this PR, I'm guessing it's using a default of the repo's root directory since that

app/store_loader.go
iramiller: Is there a way to determine if the current node is a validator and only show this message if that is the case?  Also a limit of 13 is very low ... I would say we want something more along the lines of "x < 1000" here as there are reasonable setups that might use 100 or so.

SpicyLemon (approved):

iramiller (approved):

Issue #1834, opened 2024-02-07
Allow transfer agents to do exchange settlements with restricted coins.


SpicyLemon:
## Summary

Allow for the transfer agent to be set in the context (before initiating a send) such that the marker's send restrictions can utilize that information in the send restrictions.

## Problem Definition

In the exchange module, lots of settlement actions require an admin to initiate them. If such an admin is also a transfer authority on a marker, transfer of that marker through the exchange module should be allowed for them.

Currently, in the marker send restrictions, only the `to` and `from` are available though, so that admin's transfer authority cannot be conveyed from other endpoints like the exchange module's settlement stuff.

## Proposal

* Create helpers for setting and getting a transfer agent address in the context.
* Update the marker's send restriction to check for a transfer agent and utilize that information (and the new `ValidateCanSend`).
* Update the exchange settlement endpoints to set the transfer agent in the context before initiating transfers.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1835, opened 2024-02-08
Allow transfer agents to do exchange settlements with restricted coins.

SpicyLemon:
## Description

closes: #1834 

This PR:
1. Makes it possible to provide a transfer agent to the marker module's `SendRestrictionFn`.
2. Provide the `admin` as the transfer agent for applicable transfers being orchestrated by the exchange module.
3. Prevent funds from going to or from a marker without the transfer agent having deposit or withdraw access (respectively). This applies to both the marker module's `SendRestrictionFn` and its `Transfer` endpoint.
4. Prevent funds from being sent to a blocked address by the exchange module.
5. Adds an empty `tourmaline-rc2` upgrade handler.

---

write a little note why.


x/exchange/keeper/market.go
```diff
@@ -1537,8 +1539,19 @@ func (k Keeper) GetMarketBrief(ctx sdk.Context, marketID uint32) *exchange.Marke
...
+	if err != nil {
+		return fmt.Errorf("invalid admin %q: %w", withdrawnBy, err)
+	}
+	if k.bankKeeper.BlockedAddr(toAddr) {
+		return fmt.Errorf("%s is not allowed to receive funds", toAddr)
+	}
 	marketAddr := exchange.GetMarketAddress(marketID)
-	err := k.bankKeeper.SendCoins(ctx, marketAddr, toAddr, amount)
+	xferCtx := markertypes.WithTransferAgent(ctx, admin)
+	if toAddr.Equals(admin) {
```
iramiller, line 1551: This feels like logic that we shouldn't have to understand in the exchange module ... for the sdk50 version we might consider a special ante handler that takes care of this type of setup instead of having to ensure the restrictions are implemented at every use in every module.

x/exchange/keeper/fulfillment.go
```diff
@@ -222,35 +222,40 @@ func (k Keeper) FillAsks(ctx sdk.Context, msg *exchange.MsgFillAsksRequest) erro
 }
 
 // SettleOrders attempts to settle all the provided orders.
-func (k Keeper) SettleOrders(ctx sdk.Context, marketID uint32, askOrderIDs, bidOrderIds []uint64, expectPartial bool) error {
+func (k Keeper) SettleOrders(ctx sdk.Context, req *exchange.MsgMarketSettleRequest) error {
```
Taztingo, line 225: I like how the params were grouped into a single object. However, I recommend using an interface over a message because it is more intuitive for a keeper method, and it will also help down the road with dependency injection. Your proto messages were generated with getters so you should be able to leverage those instead of directly accessing a field.

x/marker/keeper/keeper.go
```diff
@@ -267,22 +268,30 @@ func (k Keeper) GetSendDenyList(ctx sdk.Context, markerAddr sdk.AccAddress) []sd
...
+			errs = append(errs, fmt.Errorf("net asset value denom cannot match marker denom %q", marker.GetDenom()))
+			continue
 		}
+
 		if nav.Price.Denom != types.UsdDenom {
 			_, err := k.GetMarkerByDenom(ctx, nav.Price.Denom)
 			if err != nil {
-				return fmt.Errorf("net asset value denom does not exist: %v", err.Error())
+				if err2 := nav.Validate(); err2 == nil {
+					navEvent := types.NewEventSetNetAssetValue(marker.GetDenom(), nav.Price, nav.Volume, source)
```
Taztingo, line 282: Why are we emitting an event for a net asset value that is being skipped?

Taztingo (approved):

x/marker/keeper/keeper.go
iramiller: less events are better based on recent understanding...

x/exchange/keeper/fulfillment.go
SpicyLemon: In this instance, I disagree that an interface is more intuitive for a keeper method. First off, though, I got the object for free. It was already defined, already being created and passed in, it has all the data needed, and little (if any) extra. If I made an interface, this function is the only thing that would make use of it, and this `MsgMarketSettleRequest` is the only thing that would implement it. Now think about someone trying to call this method elsewhere. Since it takes in the object, they know, "I need to make one of those objects." But if it took in an interface, now they'd need to either track down which object they can use, or else they have to write their own object.

x/marker/keeper/keeper.go
SpicyLemon: We want a record of the NAV data somewhere, even if it doesn't get recorded in state. This is a reflection of `SetNetAssetValue` which always emits the event as long as the NAV is valid. It too might emit this event without the NAV being written to state.

This `AddSetNetAssetValues` method was designed for the `AddNetAssetValues` endpoint. I don't know enough about the business requirements on that to properly try to remove some of the failure points, though. So, even though I feel like we should record NAVS even if there's no marker for the price denom, I wasn't comfortable removing this failure point. But at least this way, the data still goes somewhere.

In the exchange module, where `AddSetNetAssetValues` is also called, there's a bunch of other stuff that also helps ensures this event is emitted, even if it can't be stored to state.

iramiller: Emitting events isn't really important with the new state streaming direction that the Cosmos SDK is going ... eventually we will may remove all events from the block results.

x/exchange/keeper/market.go
SpicyLemon: I don't think we can get away with taking this logic out of the exchange module. At some level, something would have to identify that, for this endpoint (and a couple others), the context should have a transfer agent. There isn't anything that automatically identifies a field as being a transfer agent, it's something that needs to be explicitly identified. Since that identification needs to be somewhere, I feel its best to put it right where that functionality is needed.

The thing I like about having it here is that it's clear that there's a transfer agent for this send, and its only this send that has a transfer agent (in this flow at least). Another thing I like is that I can have unit tests that specifically ensure that a transfer agent is being used. If we rely on an antehandler for this, not only would it be difficult to test that a transfer agent is being used when it should, but we'd also have to add the transfer agent to the context for some unit tests; either way requires understanding of transfer agents in the exchange module.

Ultimately, I feel that, since transfer agents are so rare, it'd be better to explicitly identify them where they're needed, rather than obfuscate their use.

Taztingo (approved):

iramiller (approved):

PR #1844, opened 2024-02-20
sdk-50 TODOs

nullpointer0x00:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -60,6 +60,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/version"
 	"github.com/cosmos/cosmos-sdk/x/auth"
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	authcodec "github.com/cosmos/cosmos-sdk/x/auth/codec"
 	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	authtx "github.com/cosmos/cosmos-sdk/x/auth/tx"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
@@ -452,11 +453,7 @@ func New(
 	app.CapabilityKeeper.Seal()
 
 	// add keepers
-	// TODO[1760]: account: Put back this call to NewAccountKeeper with fixed arguments.
-	// app.AccountKeeper = authkeeper.NewAccountKeeper(
-	// 	appCodec, keys[authtypes.StoreKey], app.GetSubspace(authtypes.ModuleName),
-	// 	authtypes.ProtoBaseAccount, maccPerms, AccountAddressPrefix,
-	// )
+	app.AccountKeeper = authkeeper.NewAccountKeeper(appCodec, runtime.NewKVStoreService(keys[authtypes.StoreKey]), authtypes.ProtoBaseAccount, maccPerms, authcodec.NewBech32Codec(AccountAddressPrefix), AccountAddressPrefix, authtypes.NewModuleAddress(govtypes.ModuleName).String())
 
 	// TODO[1760]: bank: Put back this call to NewBaseKeeper with fixed arguments.
 	// app.BankKeeper = bankkeeper.NewBaseKeeper(


internal/handlers/msg_service_router.go
```diff
@@ -5,40 +5,45 @@ import (
...
 	"github.com/provenance-io/provenance/internal/helpers"
+	"github.com/provenance-io/provenance/internal/protocompat"
 	msgfeeskeeper "github.com/provenance-io/provenance/x/msgfees/keeper"
 )
 
 // PioMsgServiceRouter routes fully-qualified Msg service methods to their handler with additional fee processing of msgs.
 type PioMsgServiceRouter struct {
 	interfaceRegistry codectypes.InterfaceRegistry
 	routes            map[string]MsgServiceHandler
+	hybridHandlers    map[string]func(ctx context.Context, req, resp protoiface.MessageV1) error
```
SpicyLemon, line 31: Nit: It'd be nice to use a named type here instead of the full `func` signature (same with in `NewPioMsgServiceRouter`).
```suggestion
	hybridHandlers    map[string]protocompat.Handler
```

scripts/proto-update-deps.sh
```diff
@@ -70,16 +65,20 @@ rm -rf 'proto/cosmos'
...
-} > "$CONFIO_FILE"
-rm "$CONFIO_FILE.orig"
+# {
+#   head -n3 "$ICS23_FILE.orig"
+#   printf 'option go_package = "github.com/confio/ics23/go";\n'
+#   printf 'option java_package = "tech.confio.ics23";\n'
+#   printf 'option java_multiple_files = true;\n'
+#   tail -n+4 "$ICS23_FILE.orig"
+# } > "$ICS23_FILE"
+# rm "$ICS23_FILE.orig"
```
SpicyLemon, line 84: The new `proofs.proto` file has the a `go_package` line, so we probably don't need that customization still. It doesn't have the `java_package` or `java_multiple_files` options though. The stuff in the `/protoBindings` directory does some java stuff with all these protos (supposedly), and might be affected.

I'm pretty sure the new `buf` stuff we're switching too allows that to be defined at a higher level (e.g. in a `.buf` file somewhere). If we still need those java options, it'd be great to do it that way (instead of modifying the file like we've been doing).

scripts/protocgen.sh
```diff
@@ -7,36 +7,27 @@
...
 done
 
-# command to generate docs using protoc-gen-doc
-buf protoc \
-  -I "proto" \
-  -I "third_party/proto" \
-  --doc_out=./docs \
-  --doc_opt=./docs/protodoc-markdown.tmpl,proto-docs.md \
-  $(find "$(pwd)/proto" -maxdepth 5 -name '*.proto')
-
```
SpicyLemon, line 39: We should probably figure out what (if anything) consumes that `proto-docs.md` file so we can know if its safe to not generate anymore. In my past explorations with the updated `buf` stuff, I couldn't find a straight-forward replacement for this, though. So if anything still needs that content, we might have to come up with a different solution for it.

third_party/proto/cosmos/auth/v1beta1/query.proto
```diff
@@ -7,45 +7,55 @@ import "google/protobuf/any.proto";
...
-  // Accounts returns all the existing accounts
+  // Accounts returns all the existing accounts.
+  //
+  // When called from another module, this query might consume a high amount of
+  // gas if the pagination field is incorrectly set.
   //
   // Since: cosmos-sdk 0.43
   rpc Accounts(QueryAccountsRequest) returns (QueryAccountsResponse) {
-    option (google.api.http).get = "/cosmos/auth/v1beta1/accounts";
+    option (cosmos.query.v1.module_query_safe) = true;
```
SpicyLemon, line 23: Note to self: investigate what this `cosmos.query.v1.module_query_safe` means/does.

third_party/proto/cosmos/authz/v1beta1/authz.proto
```diff
@@ -13,28 +14,17 @@ option (gogoproto.goproto_getters_all) = false;
 // GenericAuthorization gives the grantee unrestricted permissions to execute
 // the provided method on behalf of the granter's account.
 message GenericAuthorization {
-  option (cosmos_proto.implements_interface) = "Authorization";
+  option (amino.name)                        = "cosmos-sdk/GenericAuthorization";
```
SpicyLemon, line 17: Note to self: Investigate what this `amino.name` means/does (and if we need to add it to some of our stuff).

scripts/protocgen.sh
SpicyLemon: I pinged AJ on this. He said he doesn't know of anything that uses it.

So we're probably okay not generating it anymore, and should probably also delete the `docs/proto-docs.md` file (the thing this generated).

internal/handlers/msg_service_router.go
nullpointer0x00: You got it.  Done!

scripts/proto-update-deps.sh
nullpointer0x00: Ok, I will remove this.  Thanks for the input on that.

SpicyLemon (approved):

Taztingo (approved):

PR #1873, opened 2024-03-15
TODOs finalize-block, event-history, signing,add cosmos.msg.v1.signer to protos txs

nullpointer0x00:
## Description

This PR finished up the finalize-block, event-history, and signing TODOS.

The `cosmos.msg.v1.signer` was needed for the protos to work.  However, I will add another TODO to our list for removing GetSigners() from the message types.

While doing the finalize-block TODOs, the rewards vote tally rules broke due to events changing.  I have fixed that in this PR.

closes: #XXXX

---

write a little note why.


app/sim_bench_test.go
```diff
@@ -41,8 +40,7 @@ func BenchmarkFullAppSimulation(b *testing.B) {
 	app := New(logger, db, nil, true, map[int64]bool{}, b.TempDir(), simcli.FlagPeriodValue, MakeEncodingConfig(), simtestutil.EmptyAppOptions{}, interBlockCacheOpt())
 
 	// run randomized simulation
-	// TODO[1760]: event-history: Add _ return arg back in.
-	_, simParams, simErr := simulation.SimulateFromSeed(
+	_, _, simParams, simErr := simulation.SimulateFromSeed(
```
SpicyLemon, line 43: The `SimulateFromSeed` function still returns only three arguments. I made `SimulateFromSeedProv` that also returns the last block time of the simulation. So for the tests where we don't need it (like this one) we can still use `SimulateFromSeed`.

app/test_helpers.go
```diff
@@ -75,15 +76,15 @@ type SetupOptions struct {
...
+func setup(t *testing.T, withGenesis bool, invCheckPeriod uint, chainID string) (*App, GenesisState) {
 	db := dbm.NewMemDB()
 	encCdc := MakeEncodingConfig()
 	// set default config if not set by the flow
 	if len(pioconfig.GetProvenanceConfig().FeeDenom) == 0 {
 		pioconfig.SetProvenanceConfig("", 0)
 	}
 
-	app := New(loggerMaker(), db, nil, true, map[int64]bool{}, t.TempDir(), invCheckPeriod, encCdc, simtestutil.EmptyAppOptions{})
+	app := New(loggerMaker(), db, nil, true, map[int64]bool{}, t.TempDir(), invCheckPeriod, encCdc, simtestutil.EmptyAppOptions{}, baseapp.SetChainID(chainID))
```
SpicyLemon, line 87: Does this work when `chainID` is `""`? Should we have it use a default instead?

internal/handlers/bank_send_restriction_test.go
```diff
@@ -174,15 +175,18 @@ func ConstructAndSendTx(tt *testing.T, app piosimapp.App, ctx sdk.Context, acct
...
-	*/
+	res, err := app.FinalizeBlock(&abci.RequestFinalizeBlock{
+		Height: ctx.BlockHeight() + 1,
+		Time:   time.Now().UTC(),
+		Txs:    [][]byte{txBytes},
+	},
+	)
+	require.NoError(tt, err, "FinalizeBlock expected no error")
+	require.Len(tt, res.TxResults, 1, "TxResults expected len not met")
+	require.Equal(tt, expectedCode, res.TxResults[0].Code, "res=%+v", res)
```
SpicyLemon, line 188:
```suggestion
	require.Equal(tt, int(expectedCode), int(res.TxResults[0].Code), "res=%+v", res)
```

When an assert.Equal fails on unsigned numbers, the failure message outputs them as hex. As ints, the numbers are shown as decimals, which makes it easier to figure out what's going on.

x/msgfees/keeper/query_server.go
```diff
@@ -52,38 +54,36 @@ func (k Keeper) QueryAllMsgFees(c context.Context, req *types.QueryAllMsgFeesReq
...
-		gasUsed := int64(float64(gasInfo.GasUsed) * float64(gasAdjustment))
-		totalFees := gasMeter.FeeConsumed().Add(sdk.NewCoin(baseDenom, minGasPrice.Amount.MulRaw(gasUsed)))
-
-		return &types.CalculateTxFeesResponse{
-			AdditionalFees: gasMeter.FeeConsumed(),
-			TotalFees:      totalFees,
-			EstimatedGas:   uint64(gasUsed),
-		}, nil
-	*/
 	return nil, errors.New("not yet updated")
```
SpicyLemon, line 88: This line should be deleted.

app/sim_bench_test.go
nullpointer0x00: Weird, I edited the vendor simulate file to return the block time for `SimulateFromSeed` so everything was compiling locally (not realizing it was in the comos-sdk).  It used to not allow that and wouldn't compile.  Not sure what changed... Anyway, makes sense and is fixed now.

app/test_helpers.go
nullpointer0x00: It defaults to that when not set.  However, we have some tests that are setting the chain-id  in one area but not the app level.  In v50, on init chain it checks to see if the chainid match.

Test: https://github.com/provenance-io/provenance/blob/fa05e85e513a05dce5c6015152296d96657d0e75/internal/handlers/msg_service_router_test.go#L234

Check: https://github.com/provenance-io/cosmos-sdk/blob/6e4cf880fcdffb7ec38b3f8704880e7bad01da32/baseapp/abci.go#L40

Perhaps, we could just remove the naming of the chain from the tests and revert my other changes?  I was just trying not to change too much on how the tests were setup.

Taztingo (approved):

Taztingo (approved):

PR #1879, opened 2024-03-18
Adds IBC and BasicModuleManager

Taztingo:
## Description

Adds more IBC functionality back to the app, and replaces ModuleBasics with BasicModuleManager.

---

write a little note why.


app/upgrades.go
```diff
@@ -205,6 +208,17 @@ var upgrades = map[string]appUpgrade{
...
+			err = migrateParams(ctx, app)
+			if err != nil {
+				return nil, err
+			}
+
+			err = upgradeIBC(ctx, app)
+			if err != nil {
+				return nil, err
+			}
+
```
SpicyLemon, line 221: Usually the custom migration stuff needs to happen after `runModuleMigrations`. Otherwise, you're using code/functions designed for the new state against data formatted using the old state. Are we sure these should be done first?

app/upgrades.go
```diff
@@ -533,3 +558,48 @@ func updateIbcMarkerDenomMetadata(ctx sdk.Context, app *App) {
...
+	params := app.IBCKeeper.ClientKeeper.GetParams(ctx)
+	params.AllowedClients = append(params.AllowedClients, exported.Localhost)
+	app.IBCKeeper.ClientKeeper.SetParams(ctx, params)
+	ctx.Logger().Info("Done upgrading to IBCv8")
+	return nil
+}
+
+// Migrate to new ConsensusParamsKeeper
+// TODO: Remove with the umber handlers.
+func migrateParams(ctx sdk.Context, app *App) error {
```
SpicyLemon, line 595: Can we call this `migrateBaseappParams`?

There's a lot of params that we'll need to migrate on top of these ones. I'm not yet sure if we'll want to do those using module migrations or directly in the upgrade. I do know the upgrade will need to make a call to update the bank params, though.

app/upgrades.go
```diff
@@ -533,3 +558,48 @@ func updateIbcMarkerDenomMetadata(ctx sdk.Context, app *App) {
...
+func upgradeIBC(ctx sdk.Context, app *App) error {
+	if err := pruneIBCExpiredConsensusStates(ctx, app); err != nil {
+		return err
+	}
+	return upgradeToIBCv8(ctx, app)
+}
+
+// pruneIBCExpiredConsensusStates upgrades IBC from v6 to v8.
+func pruneIBCExpiredConsensusStates(ctx sdk.Context, app *App) error {
+	ctx.Logger().Info("Pruning expired consensus states for IBC")
```
SpicyLemon, line 572: For log messages, we should include ending punctuation.

app/upgrades.go
Taztingo: I can rename this to migrateIBCParams since this is IBC specific.

app/upgrades.go
Taztingo: This is the order that was recommended by IBC. I also noticed that our chain runs module migrations first, but other chains and the sdk documentation recommend doing it last. I'm digging into the documentation for this now, but I don't think our chain does anything exceptional that should differentiate it.

SpicyLemon: Gotcha. I think that, in general, you want stuff to happen after the module migrations, but there are definitely cases where you want things done before them.

What can happen is a discrepancy between how the data is stored and how the data is expected. We had an upgrade a while back where this was an issue, but I can't remember too many details; I believe it was bank module related, though.

Basically, we needed to do some lookup/processing of stuff in a module that had an upgrade. The migration was moving a bunch of module's data in order to use different key and/or value formats. E.g. all the state keys that started with `0x01` were being moved to `0x02`. All the code (except for inside the migration) expected the data to be at `0x02`. So if we did our lookup/processing first, it just looked like there wasn't any data. Worse problems can happen if the key or value formats change without being moved; the new code tries to read the entry using the new format, but can't, resulting in anything from a panic to a silent failure.

A more concrete example is `removeInactiveValidatorDelegations`. If there's a module upgrade in the staking keeper that changes how that stuff is recorded, then we need to do that **after** the module migrations.

cmd/provenanced/cmd/root.go
```diff
@@ -60,7 +63,25 @@ const (
...
+	// tempDir creates a temporary home directory.
+	var tempDir = func() string {
+		dir, err := os.MkdirTemp("", "provenanced")
+		if err != nil {
+			panic("failed to create temp dir: " + err.Error())
+		}
+		defer os.RemoveAll(dir)
+
+		return dir
+	}
```
SpicyLemon, line 75:
```suggestion
	tempDir, err := os.MkdirTemp("", "provenanced")
	if err != nil {
		panic(fmt.Errorf("failed to create temp dir: %w", err))
	}
	os.RemoveAll(tempDir)
```

Then below, when creating `tempApp`, provide `tempDir` instead of `tempDir()`.

go.mod
```diff
@@ -32,6 +33,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/hashicorp/go-metrics v0.5.2
 	github.com/rakyll/statik v0.1.7
+	github.com/regen-network/cosmos-proto v0.3.1 // TODO[1760]: Verify that this is still needed
```
SpicyLemon, line 36: Once we get our proto-gen stuff updated, I don't think we'll need this anymore. Probably good to leave this comment there, but wanted to also answer it.

x/attribute/module.go
```diff
@@ -28,10 +29,11 @@ import (
...
 var (
-	_ module.AppModule           = AppModule{}
-	_ module.AppModuleBasic      = AppModuleBasic{}
-	_ module.AppModuleSimulation = AppModule{}
-	// TODO[1760]: app-module: Add more assertions for the new types and clean up stuff no longer needed.
+	_ module.AppModuleBasic      = (*AppModule)(nil)
+	_ module.AppModuleSimulation = (*AppModule)(nil)
+
+	_ appmodule.AppModule       = (*AppModule)(nil)
+	_ appmodule.HasBeginBlocker = (*AppModule)(nil)
```
SpicyLemon, line 36: Is there an `appmodule.HasEndBlocker` that we should be asserting here too?

SpicyLemon: NM... missed that this was in a single module (and not the app as a whole). All good.

SpicyLemon (approved):

nullpointer0x00 (approved):

nullpointer0x00 (approved):

SpicyLemon (approved):

nullpointer0x00 (approved):

PR #1887, opened 2024-03-21
Update to BaseApp.RegisterStreamingServices and add Textual Signing

Taztingo:
## Description

This PR updates app.go to use the new RegisterStreamingServices method attached to BaseApp. This should provide the same logic that we had before. It also adds Textual Signing.

closes: #1760

---

write a little note why.


cmd/provenanced/cmd/root.go
```diff
@@ -79,10 +84,36 @@ func NewRootCmd(sealConfig bool) (*cobra.Command, params.EncodingConfig) {
...
-				initClientCtx = initClientCtx.WithHomeDir(rootDir)
+			initClientCtx, err := client.ReadPersistentCommandFlags(initClientCtx, cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			initClientCtx, err = clientconfig.ReadFromClientConfig(initClientCtx)
+			if err != nil {
+				return err
+			}
```
SpicyLemon, line 95: This is going to be problematic, and I'm not yet sure how best to fix it.

To begin with, we do our own thing for the `client.config` file. I'm pretty sure that's only because we wanted a different default for one or more of the fields. We can now handle those defaults in the config manager stuff, though, so maybe we can just get rid of our version and switch back to the SDKs.

Additionally, we have the concept of a packed config file, which combines all the configs into a single JSON with just entries for fields with non-default values. This `ReadFromClientConfig` will not find that.

We can probably leave this all here for now, but we should put a `TODO[1760]` comment on it so we can come back to it and straighten it all out later.

SpicyLemon (approved):

SpicyLemon (approved):

nullpointer0x00 (approved):

PR #1888, opened 2024-03-21
v1.18.0 mainnet upgrade handler for net asset values and block height of pricing timestamp

nullpointer0x00:
## Description

This PR updates marker Net Asset Values (NAV) on the blockchain using the latest data from the Figures Pricing API, ensuring values accurately reflect current market conditions. The API provides prices like $1.23, which we represent as an integer 1230 to maintain precision. Data is sourced from `https://figure.tech/service-pricing-engine/external/api/v1/pricing/marker/new?time=2024-03-21T00:00:00.000000Z`. Each update aligns the NAV with the blockchain block height according to the `pricingTimestamp`, enhancing transparency and accuracy in marker valuations.

closes: #XXXX

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
	- Introduced an upgrade handler for setting net asset values and updating block heights for specific network markers.
	- Added functionality to update net asset values with corresponding heights for markers during the upgrade process.

- **Chores**
	- Added a note for future removal of specific functionality in the codebase.

- **Tests**
	- Implemented tests to verify the addition and update of marker net asset values with heights.


x/marker/keeper/keeper.go
```diff
@@ -318,6 +318,29 @@ func (k Keeper) SetNetAssetValue(ctx sdk.Context, marker types.MarkerAccountI, n
...
 }
 
+// SetNetAssetValueWithBlockHeight adds/updates a net asset value to marker with a specific block height
+func (k Keeper) SetNetAssetValueWithBlockHeight(ctx sdk.Context, marker types.MarkerAccountI, netAssetValue types.NetAssetValue, source string, blockHeight uint64) error {
+	netAssetValue.UpdatedBlockHeight = blockHeight
+	if err := netAssetValue.Validate(); err != nil {
+		return err
+	}
+
+	setNetAssetValueEvent := types.NewEventSetNetAssetValue(marker.GetDenom(), netAssetValue.Price, netAssetValue.Volume, source)
```
nullpointer0x00, line 328: @iramiller Any thoughts on if we should still emit an event here?  We can either remove it, leave it as is, or add a new optional property to net asset value typed event of the height.

x/marker/keeper/keeper.go
```diff
@@ -318,6 +318,29 @@ func (k Keeper) SetNetAssetValue(ctx sdk.Context, marker types.MarkerAccountI, n
...
+	key := types.NetAssetValueKey(marker.GetAddress(), netAssetValue.Price.Denom)
+	bz, err := k.cdc.Marshal(&netAssetValue)
+	if err != nil {
+		return err
+	}
+	store := ctx.KVStore(k.storeKey)
+	store.Set(key, bz)
+
+	return nil
+}
```
Taztingo, line 342: Looks good, but it looked really familiar. The function above it has exactly the same logic. Is it possible to update the `SetNetAssetValue` to use this so we don't have to maintain both?

Example...
```
func (k Keeper) SetNetAssetValue(ctx sdk.Context, marker types.MarkerAccountI, netAssetValue types.NetAssetValue, source string) error {
	return k.SetNetAssetValueWithBlockHeight(ctx, marker, netAssetValue, source, uint64(ctx.BlockHeight())
}
```

nullpointer0x00: I thought about that. However, @iramiller and I were talking about events and how they should work with the upgrade handler. I will wait till I hear back from my other conversation I added to the method for his input. Also, didn't know if it is something we might want to remove after upgrade.

Taztingo (approved):

x/marker/keeper/keeper.go
iramiller: I suppose we can leave it in there ... the block of events for changes in the upgrade is still useful even if the data set applies to a range of dates/blockheights.

x/marker/keeper/keeper.go
iramiller: The larger flow for events feels like we would change the system to suppress "non-indexed" events as a configuration item ... if we do that it would make all this go away as these events mostly wouldn't be indexed going forward unless someone explicitly is looking for that in which case we presume they know what they are doing

SpicyLemon: My usual strategy is to have the function emit the events. If the caller doesn't want those events, they can set a new event manager in the context that is provided when the function is called.

SpicyLemon (approved):

PR #1957, opened 2024-04-26
Clean up Msgs and Codecs.

SpicyLemon:
## Description

This PR does the following:
* Adds unit tests ensuring the new `signer` stuff works for all our `Msg`s.
* Fixes the signers of a few `Msg`s.
* Removes all the `GetSigners` methods as well as some other ones previously needed by the `Msg` interface.
* Standardizes and cleans up all the `codec.go` files.
* Removes all the `ModuleCdc` variables.
* Uses the same options on the `Coins` proto fields that the SDK does.
* Switch to auto-generated `String` and `Equal` methods for most proto messages.

Related to:
* #1760

---

write a little note why.


nullpointer0x00 (approved):

proto/cosmos/sanction/v1beta1/sanction.proto
```diff
@@ -13,15 +14,23 @@ message Params {
   // If this is zero, immediate sanctioning is not available.
   // Otherwise, if a sanction governance proposal is issued with a deposit at least this large, a temporary sanction
   // will be immediately issued that will expire when voting ends on the governance proposal.
-  repeated cosmos.base.v1beta1.Coin immediate_sanction_min_deposit = 1
-      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+  repeated cosmos.base.v1beta1.Coin immediate_sanction_min_deposit = 1 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (amino.dont_omitempty)   = true,
+    (amino.encoding)         = "legacy_coins"
```
Taztingo, line 21: Why are these needed, and do we even use amino encoding anymore?

proto/cosmos/sanction/v1beta1/sanction.proto
```diff
@@ -13,15 +14,23 @@ message Params {
...
   // If this is zero, immediate unsanctioning is not available.
   // Otherwise, if an unsanction governance proposal is issued with a deposit at least this large, a temporary
   // unsanction will be immediately issued that will expire when voting ends on the governance proposal.
-  repeated cosmos.base.v1beta1.Coin immediate_unsanction_min_deposit = 2
-      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+  repeated cosmos.base.v1beta1.Coin immediate_unsanction_min_deposit = 2 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (amino.dont_omitempty)   = true,
+    (amino.encoding)         = "legacy_coins"
```
Taztingo, line 32: Same as above...

x/trigger/types/codec.go
```diff
@@ -1,19 +1,16 @@
...
-// ignoring RegisterLegacyAminoCodec registers all the necessary types and interfaces for the
-// double check
+// RegisterInterfaces registers concrete implementations for this module.
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
-	registry.RegisterImplementations(
-		(*sdk.Msg)(nil),
-		&MsgCreateTriggerRequest{},
-		&MsgDestroyTriggerRequest{},
-	)
+	messages := make([]proto.Message, len(AllRequestMsgs))
```
Taztingo, line 11: Do we need an extra copy/heap allocation for this? RegisterImplementations is only reading, and this would also allow us to get rid of the new import.

`registry.RegisterImplementations((*sdk.Msg)(nil), AllRequestMsgs...)`

Taztingo (approved):

proto/cosmos/sanction/v1beta1/sanction.proto
SpicyLemon: I'm not entirely sure. I added them here because the SDK added them to all their `Coins` fields, but I couldn't find anything that indicated why they were added or what exactly they do.  The `amino.proto` documentation makes it sound like it's related to JSON encoding, though.

I took these lines out and regenerated stuff, and the only thing that changed was the `fileDescriptor_...` byte slice which is pretty much just the contents of that file. But that's also the case with options like `(cosmos_proto.scalar) = "cosmos.AddressString"` or `(cosmos.msg.v1.signer)`.

I did some testing on two versions of a proto message, one with these options, one without. I generated JSON from each in three different ways:`encoding/json.Marshal`, `"github.com/cosmos/cosmos-sdk/codec".Codec#MarshalJSON` and `"github.com/cosmos/cosmos-sdk/codec".LegacyAmino#MarshalJSON`. I used a fully set-up encoding config for the latter two. All six combos produced the same json.

So either:
1. We don't need these options.
2. Or we're not wiring things up right, causing these to not have meaning.
3. Or they affect something else.

x/trigger/types/codec.go
SpicyLemon: Yeah.  `RegisterImplementations` takes in a `[]proto.Message`, but the `AllRequestMsgs` variable is a `[]sdk.Msg`. Since `proto.Message` is a subset of `sdk.Msg`, Golang can automatically cast a single entry, but it can't cast slices like that, so we have to copy them.

In previous versions of the SDK, the `sdk.Msg` interface was a `proto.Message` with several other methods. It was important to have some sort of type assertion on them to make sure all required methods were defined. That's why the `AllRequestMsgs` variables are all `[]sdk.Msg`; having them in that list gave us the desired type assertion for free.

Over the various versions, though, the SDK has removed methods from the `sdk.Msg` interface, and it is now just an alternative name for `proto.Message`. If we changed `AllRequestMsgs` to a `[]proto.Message`, we wouldn't have to do this copying. I'm not quite sure what other ramifications that might have, though. But I'd rather leave them as `[]sdk.Msg` since that's the type we're registering them as in these codecs.

Taztingo: I actually got this mixed up with `attribute/types/codec.go` where `allRequestMsgs` already has the []proto.Message type.

PR #2127, opened 2024-08-14
Ignore this for now.

SpicyLemon:
## Description

This PR will add caching to the go stuff in the tests workflow, hopefully speeding things up overall.

---

write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting)).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,14 @@ jobs:
             go.mod
             go.sum
             .github/workflows/test.yml
+      # This setup-go action will cache/restore using a key with a hash of go.sum.
+      - uses: actions/setup-go@v5
+        id: setup-go
+        with:
+          go-version-file: 'go.mod'
+      - name: Populate Mod Cache
+        if: steps.setup-go.outputs.cache-hit != 'true'
+        run: go mod download
       - name: Define Variables
         id: def-vars
         run: |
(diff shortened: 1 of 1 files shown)


SpicyLemon:
Eh. There really isn't any improvement with this.

The `setup-go` action already restores from a previous cache when there aren't any changes to `go.sum`. So this change would only really affect PRs that have a change to `go.sum`, and tests haven't been run yet with it. If there's no changes to `go.sum`, or tests for the PR have already run using the same `go.sum`, the jobs restore go stuff from the cache.

Without this PR, when there's no cache to restore from, each of the test runners will set up go, download the needed modules, and at the end, attempt to save the new cache. There will be 10 runners trying to do this, but only one save will go through. Any subsequent runs (in the same PR or even in a different PR with the same `go.sum`), will then be able to restore from that cache.

This PR added some steps to `init` that would set up go and, if not restoring from cache, download all the modules we need so that at the end of the `init` job, that cache would be written. That'd then allow the test jobs to always be able to restore from cache.

So, in total, there'd be less processing needed because what used to be done 10 times, is now only being done once. But those 10 times were all being done in parallel anyway, so it really only took about the same amount of time as doing it just once. And adding the steps to `init` add a tiny bit of extra overhead in all cases.

That might matter if we start regularly running out of runners, but a) it only takes maybe 15 or 30 seconds to set up go and download the modules, and b) that setup part is regularly skipped because the cache already exists. So as is, we just sometimes tie up 10 runners for 15-30 seconds longer than we might by setting stuff up in `init`.

Ultimately, there just isn't any real improvement by doing this.

PR #2136, opened 2024-08-23
Remove some unused stuff from the devtools makefile

SpicyLemon:
## Description

This PR removes a bunch of unused stuff from `contrib/devtools/Makefile`, and also moves some stuff from there into the primary `Makefile`.  There's a little more cleanup done in this too.

---

write a little note why.

## Summary by CodeRabbit

- **Improvements**
	- Streamlined the build process by removing obsolete configurations from Makefiles, enhancing maintainability and clarity.
	- Improved logic for detecting the Go executable and validating the Go version, leading to a more robust build experience.
	- Simplified developer tools Makefile by eliminating unnecessary environment detection and setup logic.
	- Updated error handling for missing Go executable to provide clearer user feedback.
  
These changes contribute to a more efficient and user-friendly development environment.


SpicyLemon:
When working on this I found out some interesting things about makefile syntax that I'd like to share.

When you assign a value using `=`, the RHS is NOT executed. The variable contains all the references still. So if you do `FOO = $(shell echo foofy >&2 && echo foo)`, that shell command is not executed (yet). It'll be executed any/every time you're trying to use the value of the variable though. So if you then do `BAR := $(FOO) $(FOO)` then you'll see "foofy" printed to stderr twice.

The `:=` assignment operator does evaluate/execute the RHS before setting the variable though. So, building on the above example, if you then do `BAZ := $(BAR) $(BAR)`, BAZ will have the value `foo foo` and nothing new is printed to stderr since BAR's value had already been resolved too.

Another example:
1. `FOO = $(shell echo foofy >&2 && echo foo)`
2. `BAR2 = $(FOO) $(FOO)`
3. `BAZ2 := $(BAR2) $(BAR2)`

Nothing is printed to stderr when setting `BAR2` now, since it uses `=` there. But from line three, you'll end up with four `foofy`s in stderr because each `$(BAR2)` is evaluated to `$(FOO) $(FOO)` which then causes the shell command to be run twice; two `$(BAR2)`s = four `$(FOO)`s = four `foofy`s in stderr and `BAZ2` = `foo foo foo foo`.

That means that, if we were to put a line between 2 and 3 that changes `FOO` to something else, e.g. `FOO = $(shell echo foofun >&2 && echo fOO)`, then when setting `BAZ2`, it'll end up as `fOO fOO fOO fOO` and `foofun` will be printed to stderr four times. This happens because on line 2, we used `=`, so the RHS wasn't evaluated, and `BAR2` just contained references.

You can force a variable to resolve itself by doing `FOO := $(FOO)`, but I find that a little confusing, so have tried to avoid it.

---

Now, to add to all this, we have the `?=` assignment operator.

From `https://www.gnu.org/software/make/manual/html_node/Setting.html`
> These two settings of the variable ‘FOO’ are identical
> ```
> FOO ?= bar
> ```
> and
> ```
> ifeq ($(origin FOO), undefined)
> FOO = bar
> endif
> ```

The first critical piece in there is that it uses the `=` operator, and not `:=`. So if you do `FOO ?= $(shell echo foofy >&2 && echo foo)`, then if FOO was undefined, it now contains a reference to that shell function, so any uses of `$(FOO)` will cause `foofy` to be printed to stderr. So it's probably best to not use `?=` if you want the RHS to be a shell execution (or something else that might be expensive).

The second critical piece is that makefile distinguishes between a variable being undefined and being empty. If my environment has `export FOO=''`, then when make gets to `FOO ?= whatever`, it sees that the env var `FOO` is defined, so nothing happens, i.e. `$(FOO)` will resolve to an empty string.

---

Lastly, wrapping quotes are not stripped from variable values. E.g. all three of these are different: `FOO := 'foo'`, `FOO := "foo"`, and `FOO := foo`. If you did `FOO := 'foo'`, then the test `ifeq ($(FOO),foo)` is false, but `ifeq ($(FOO),'foo')` is true.

iramiller (approved):

nullpointer0x00 (approved):

PR #2194, opened 2024-10-21
Fix: Add node flag to WASM queries (build-address)

GunanshuJoshi:
## Description  

This PR adds a new `--node` flag to support custom endpoints for node connections while making Wasm queries specifically `build-address`. This update ensures that the `provenanced q wasm build-address` command works correctly with custom node configurations.  

## Issue:
While creating the predictable contract address I encountered the following error:
```bash
unknown flag: --node
````
This occurred when running the following command:

```bash
provenanced query wasm build-address <code-hash> <creator-address> <salt> --node https://rpc.test.provenance.io:443
```
It seems the flag is not being recognized properly. I suspect there might be an issue with either the flag registration in the command definition (root.go) or how the CLI parser processes the command.

The main file updated is:  
- `cmd/provenanced/cmd/root.go`  

closes: `n/a`  
I haven't created any particular issue for this change. However, this was discussed on Discord, where it was agreed that the node flag should function correctly.  
Reference: [Discord conversation](https://discord.com/channels/897185242182484029/933506958873464944/1295915569329078323).  

---

### Checklist  

- [x] **Targeted PR against the correct branch** (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting)).  
- [ ] **Linked to Github issue with discussion and accepted design** OR **link to spec that describes this work.**  
  _(I haven't created any issue for this minor fix.)_  
- [ ] **Wrote unit and integration tests** (see [Testing Guidelines](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)).  
- [ ] **Updated relevant documentation** (`docs/`) or **specification** (`x/<module>/spec/`).  
- [ ] **Added relevant GoDoc comments** ([GoDoc Documentation](https://blog.golang.org/godoc-documenting-go-code)).  
- [ ] **Added relevant changelog entries** under `.changelog/unreleased` (see [Adding Changes](https://github.com/provenance-io/provenance/blob/main/.changelog/README.md#adding-changes)).  
- [x] **Re-reviewed `Files changed` in the Github PR explorer.**  
- [ ] **Reviewed Codecov report** in the comment section below once CI passes.  

## Summary by CodeRabbit

- **New Features**
  - Introduced new command flags for custom denomination and message fee floor price.
  - Added a flag to specify the node connection for enhanced configurability.
  - Enhanced `query wasm build-addr` command with additional query flags for improved usability.

- **Improvements**
  - Enhanced error handling for better user experience, including automatic creation of the snapshot directory.
  - Updated command initialization to disable the `start` command upon error, preventing misleading usage information.
  - Improved telemetry configuration for better flexibility.

- **Miscellaneous Adjustments**
  - Removed shorthand options for conflicting flags and ensured correct command aliasing.


cmd/provenanced/cmd/root.go
```diff
@@ -222,6 +222,7 @@ func queryCommand() *cobra.Command {
 	)
 
 	cmd.PersistentFlags().String(flags.FlagChainID, "", "The network chain ID")
+	cmd.PersistentFlags().String(flags.FlagNode, "","The node to connect to (IP:Port)")
```
iramiller, line 225: Generally speaking each command will have [`flags.AddQueryFlagsToCmd(cmd)`](https://github.com/cosmos/cosmos-sdk/blob/1b517b02b48197794c8162f4ac0f03bd8114db99/client/flags/flags.go#L105-L115) added (as appropriate).  As one example the [`server.QueryBlockCmd()`](https://github.com/cosmos/cosmos-sdk/blob/31ef899f4c3c091b641e68be25214715f4ea78c7/server/cmt_cmds.go#L302) already registers the required flag.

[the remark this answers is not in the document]
iramiller: This is not the correct way to approach this.  Each command will add the appropriate flags using the collection found with `[flags.AddTxFlagsToCmd(cmd)](https://github.com/cosmos/cosmos-sdk/blob/1b517b02b48197794c8162f4ac0f03bd8114db99/client/flags/flags.go#L118-L149) if they are required.  They should not be individually added as indicated here.

SpicyLemon:
The `query wasm build-address` command does not add the standard `AddQueryFlagsToCmd` to it because it doesn't actually make a query of a node; it just runs some local calculations to output the desired result. So there's no need to provide a `--node` flag to it because it doesn't use a node; there's nothing in the command's code (defined in the wasmd library) that looks for the flag.

GunanshuJoshi:
Yes, it doesn't actually have any dependency on the node but it does tries to connect with the node. Thus, giving the following error:
```bash
Error: post failed: Post "http://localhost:26657": dial tcp 127.0.0.1:26657: connect: connection refused
```
We can also consider removing the validation for this.

cmd/provenanced/cmd/root.go
GunanshuJoshi: Hi @iramiller 
Correct, it should have recognized the node flag but it did not. Should I add  [flags.AddQueryFlagsToCmd(cmd)](https://github.com/cosmos/cosmos-sdk/blob/1b517b02b48197794c8162f4ac0f03bd8114db99/client/flags/flags.go#L105-L115) here instead?

iramiller: As @SpicyLemon pointed out it shouldn’t need this flag since this is supposed to be a local only command. 

If the process is trying to make an out going connection then that seems to be the issue that needs to be fixed vs the node connection flag.

SpicyLemon:
> Yes, it doesn't actually have any dependency on the node but it does tries to connect with the node. Thus, giving the following error:
> 
> ```shell
> Error: post failed: Post "http://localhost:26657": dial tcp 127.0.0.1:26657: connect: connection refused
> ```
> 
> We can also consider removing the validation for this.

Sorry. You're right. I was looking at an earlier version of the command, but it looks like the current version does make a request of the node.

cmd/provenanced/cmd/root.go
SpicyLemon: Yes, the `build-address` command should have [flags.AddQueryFlagsToCmd(cmd)](https://github.com/cosmos/cosmos-sdk/blob/1b517b02b48197794c8162f4ac0f03bd8114db99/client/flags/flags.go#L105-L115) applied to it. That looks like an oversight by the wasm team.

To fix it in this repo, you'd have to do something similar to `fixTxWasmInstantiate2Aliases`. After all the sub-commands have been added, find the one in question and apply `AddQueryFlagsToCmd` to it.

GunanshuJoshi: Hey, I tried the snippet below as suggested by @SpicyLemon, but it didn’t quite work. Could you please tell me if I'm missing something here?
```bash
func fixBuildAddressQueryFlags(rootCmd *cobra.Command) {
    // Find the "build-address" command.

    // cmd, _, err := rootCmd.Find([]string{"query", "build-address"})
    cmd, _, err := rootCmd.Find([]string{"query", "wasm",  "build-address"})
    if err != nil || cmd == nil {
        // If the command doesn't exist, there's nothing to do.
        return
    }

    // Apply query flags to the command.
    flags.AddQueryFlagsToCmd(cmd)
}
```
this is how the queryCommand function looked like
```bash
func queryCommand() *cobra.Command {
	cmd := &cobra.Command{
		Use:                        "query",
		Aliases:                    []string{"q"},
		Short:                      "Querying subcommands",
		DisableFlagParsing:         true,
		SuggestionsMinimumDistance: 2,
		RunE:                       client.ValidateCmd,
	}

	cmd.AddCommand(
		rpc.ValidatorCommand(),
		rpc.QueryEventForTxCmd(),
		server.QueryBlockCmd(),
		authcmd.QueryTxsByEventsCmd(),
		server.QueryBlocksCmd(),
		authcmd.QueryTxCmd(),
		server.QueryBlockResultsCmd(),
	)
	
	cmd.PersistentFlags().String(flags.FlagChainID, "", "The network chain ID")
	fixBuildAddressQueryFlags(cmd)
	// flags.AddQueryFlagsToCmd(cmd)

	return cmd
}

```

SpicyLemon: At the point where you call `fixBuildAddressQueryFlags`, the query commands for each module have not yet been added. In `NewRootCmd`, there's a call to `fixTxWasmInstantiate2Aliases` that fixes one of the wasm tx commands. That command is added with the other module commands. So right after that would be a good place to call your fixer.

Also, can we call it `fixQueryWasmBuildAddressFlags`? That way it's named after the actual command and uses the same naming scheme as the other fixers in there.

GunanshuJoshi: Hi @SpicyLemon 
Thank you very much for the input. That has resolved the issue.

GunanshuJoshi:
Hi @SpicyLemon @iramiller 
I have made the required changes. Please review and let me know if anything is left.

SpicyLemon (approved):

Taztingo (approved):

SpicyLemon:
> Hi SpicyLemon iramiller I have made the required changes. Please review and let me know if anything is left.

Looks good! Thanks for the fix!

nullpointer0x00 (approved):

SpicyLemon:
Oh darn. Sorry. We require all commits to be signed, so this can't be merged as it is. I've recreated it as #2199 (with signed commits), and I'll close this one.

Here's more info on signing commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Thanks again for the fix!

Issue #1841, opened 2024-02-12
Mint with Withdraw


iramiller:
## Summary

A significant number of transactions are submitted which mint then distribute the resulting tokens.  In the interest of streamlining the workload on the system it should be possible to add a recipient address to a mint request

## Problem Definition

Streamlining and improving the efficiency of the blockchain is all about identifying areas where things can be done more efficiently.  There are a number of transactions can be consolidated if the mint and withdraw steps were combined

## Proposal

Add an optional recipient address to the mint message/end point.  The caller must possess both the `mint` and `withdraw` permissions if using the target address, otherwise the existing behavior of `mint` and hold in the marker should be used.

NOTE: verify that a withdraw operation is not bypassing the `transfer` restrictions for moving funds between two accounts.  A `withdraw` operation is equivalent to a `bank send` where the signer holding the `withdraw` permission effectively is granted the `send` permission from the marker's escrow account.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Figure also brought up that sometimes, it'd be helpful to also have the newly minted funds committed to a market.

SpicyLemon:
To do this change, you'll do the following:
1. In `proto/provenance/marker/v1/tx.proto`, find the `message MsgMintRequest`. Add a `recipient` field.
2. Regenerate go code from the protos (requires docker) using `make proto-regen`.
3. Update the `MsgMintRequest.ValidateBasic` method (in `x/marker/types/msgs.go`) to verify that, if the `recipient` field is not empty, it is a valid AccAddress.
4. In `x/marker/keeper/msg_server.go`, find the `msgServer.Mint` method. After the coin is minted, if the new `msg.Recipient` field isn't empty, make a call to `k.Keeper.WithdrawCoins` (See the `msgServer.Withdraw` method for an example of its use).
5. Don't forget unit tests.
6. Add a changelog entry using the `.changelog/add-change.sh` script (`--help` available). If you've got `fzf` (Fuzzy Finder) installed, that script is a bit nicer.

PR #2331, opened 2025-04-21
feat(mint-with-withdraw): supporting minting with recipient address for withdrawal(#1841)

nagarajdivine:
## Description

* Add support for minting with a recipient address [#1841](https://github.com/provenance-io/provenance/issues/1841).
* Added  `recipient` param in proto file.
* Update `mint` function in `message_server.go` file with withdraw `recepient`.
* Added args `tx mint` command with `recepient`.
* Fixed all test coverage for new recipient-based logic.

closes: #1841

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Minting now supports specifying a recipient address, allowing minted coins to be sent directly to another account.
  - The command-line interface for minting accepts an optional recipient argument.
- **Documentation**
  - Updated documentation to reflect the new recipient field in minting operations.
- **Tests**
  - Enhanced test coverage for minting with recipient addresses and related permissions.

### DIFF
--- a/x/marker/keeper/msg_server.go
+++ b/x/marker/keeper/msg_server.go
@@ -275,9 +275,6 @@ func (k msgServer) Mint(goCtx context.Context, msg *types.MsgMintRequest) (*type
 		if err != nil {
 			return nil, sdkerrors.ErrInvalidAddress.Wrapf("invalid recipient: %s", msg.Recipient)
 		}
-		if recipient.Empty() {
-			return nil, sdkerrors.ErrInvalidAddress.Wrap("recipient address is zero or empty")
-		}
 		if err := k.Keeper.WithdrawCoins(ctx, admin, recipient, msg.Amount.Denom, sdk.NewCoins(msg.Amount)); err != nil {
 			ctx.Logger().Error("unable to withdraw coins", "err", err)
 			return nil, sdkerrors.ErrInvalidRequest.Wrap(err.Error())


.changelog/unreleased/features/1841-mint-with-recipien.md
```diff
@@ -1,5 +1,5 @@
 * Add support for minting with a recipient address [#1841](https://github.com/provenance-io/provenance/issues/1841).
-* Added made in recipient in proto file.
```
arnabmitra: could we change the file name while we are at it.

docs/proto-docs.md
```diff
@@ -7210,6 +7210,7 @@ MsgMintRequest defines the Msg/Mint request type
 | ----- | ---- | ----- | ----------- |
 | `amount` | [cosmos.base.v1beta1.Coin](#cosmos-base-v1beta1-Coin) |  |  |
 | `administrator` | [string](#string) |  |  |
+| `recipient` | [string](#string) |  |  |
```
arnabmitra, line 7213: imo, sometimes mint can occur without a recipient, so maybe make this field optional, but i will defer to @dcshock/@jdfigure/ @SpicyLemon , to me it seems like you are mandating that a recipient address is always present, which will also break existing client calls (i.e a breaking change i think)

arnabmitra (changes requested):

docs/proto-docs.md
nagarajdivine: As I understand, the current process has two steps: first, we mint the tokens, and then we send them to `recipient`. This implementation allowing us to add a `recipient’s address` when minting the tokens, so both steps can be combined.Please correct me if I’m wrong, but if the recipient is optional, it seems we might still need to use both steps, depending on whether a recipient is provided.

I’m open to suggestions, and I’m happy to make any updates.

nagarajdivine: I agree with you, if you would like me change, please do let me know, will do the updates :)

.changelog/unreleased/features/1841-mint-withdraw-to-recipient.md
```diff
@@ -0,0 +1,5 @@
+* Added recipient address support with withdrawal when minting [#1841](https://github.com/provenance-io/provenance/issues/1841).
+* Added recipient field in proto file.
+* Updated `mint` function in `message_server.go` file with withdraw `recipient` in standard minting flow.
+* Added args `tx mint` command with `recepient`.
+* Fixed all test coverage for new recipient-based logic.
```
SpicyLemon, line 5: These are a bit too detailed and can be removed.

docs/proto-docs.md
SpicyLemon: The new recipient address should be optional. If omitted, only the mint happens (no withdrawal, same as current behavior); but if it's provided, then after the mint, the newly minted funds should be withdrawn to that address.

x/marker/client/cli/tx.go
```diff
@@ -177,14 +177,14 @@ with the given supply amount and denomination provided in the coin argument
 // GetCmdMint implements the mint additional supply for marker command.
 func GetCmdMint() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "mint [coin]",
+		Use:     "mint [coin] [recipient]",
 		Aliases: []string{"m"},
-		Args:    cobra.ExactArgs(1),
+		Args:    cobra.ExactArgs(2),
```
SpicyLemon, line 182: The recipient needs to be optional.

```suggestion
		Args:    cobra.RangeArgs(1, 2),
```

Then, below, only parse and set the recipient if `len(args) > 1`.

x/marker/keeper/msg_server.go
```diff
@@ -264,12 +264,18 @@ func (k msgServer) Mint(goCtx context.Context, msg *types.MsgMintRequest) (*type
 	}
 
 	admin := sdk.MustAccAddressFromBech32(msg.Administrator)
+	recipient := sdk.MustAccAddressFromBech32(msg.Recipient)
```
SpicyLemon, line 267: Only do this if `len(msg.Recipient) > 0` or else it'll panic when no recipient is provided. Similarly, only do the `WithdrawCoins` below if a recipient is provided.

x/marker/types/msgs.go
```diff
@@ -92,17 +92,21 @@ func (msg MsgDeleteRequest) ValidateBasic() error {
...
 	}
 }
 
 func (msg MsgMintRequest) ValidateBasic() error {
 	if _, err := sdk.AccAddressFromBech32(msg.Administrator); err != nil {
 		return err
 	}
+	if _, err := sdk.AccAddressFromBech32(msg.Recipient); err != nil {
+		return err
+	}
```
SpicyLemon, line 109: Need to allow `msg.Recipient` to be empty.

SpicyLemon:
The first commit on this branch isn't signed, which will prevent it from merging. You'll have to do a force-push after fixing that.

x/marker/client/cli/tx.go
```diff
@@ -177,26 +177,32 @@ with the given supply amount and denomination provided in the coin argument
 // GetCmdMint implements the mint additional supply for marker command.
 func GetCmdMint() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "mint [coin]",
+		Use:     "mint [coin] [recipient]",
 		Aliases: []string{"m"},
 		Args:    cobra.ExactArgs(1),
```
SpicyLemon, line 182: Because of `Args:    cobra.ExactArgs(1)`, you cannot provide the recipient. Providing a second argument will cause cobra to return an error without even attempting to call `RunE`. You'll want `cobra.RangeArgs(1, 2)`.

Also, the `Use` line should be updated to indicate that `coin` is required, but `recipient` is optional: `"mint <coin> [recipient]"`.

---

Extra info:

The `Use` field in the `cobra.Command` has a comment on it with syntax guidance.

```
	// Use is the one-line usage message.
	// Recommended syntax is as follows:
	//   [ ] identifies an optional argument. Arguments that are not enclosed in brackets are required.
	//   ... indicates that you can specify multiple values for the previous argument.
	//   |   indicates mutually exclusive information. You can use the argument to the left of the separator or the
	//       argument to the right of the separator. You cannot use both arguments in a single use of the command.
	//   { } delimits a set of mutually exclusive arguments when one of the arguments is required. If the arguments are
	//       optional, they are enclosed in brackets ([ ]).
	// Example: add [-F file | -D dir]... [-f format] profile
	Use string
```

We don't really follow the `Arguments that are not enclosed in brackets are required.` part in order to distinguish them from part of the command. So we usually use `<>` to identify a required argument.

proto/provenance/marker/v1/tx.proto
```diff
@@ -192,6 +192,7 @@ message MsgMintRequest {
 
   cosmos.base.v1beta1.Coin amount        = 1 [(gogoproto.nullable) = false];
   string                   administrator = 2;
+  optional string          recipient     = 3;
```
SpicyLemon, line 195: You don't need or want the `optional` here. Removing it and regenerating the code should make it name the field `Recipient` again (instead of the weird `XRecipient` thing it's got now). To check if it was provided, you'd just do `if len(msg.Recipient) > 0 {...}`.

It'd probably be good to put a comment on the field in order to communicate that it's optional, something like `// recipient is the optional address of the account to receive the newly minted funds. If omitted, the minted funds are placed in the marker's account. Can only be provided if the marker is active and the administrator has withdraw permissions.`

x/marker/client/cli/cli_test.go
```diff
@@ -721,22 +721,23 @@ func (s *IntegrationTestSuite) TestMarkerTxCommands() {
 			false, &sdk.TxResponse{}, 0,
 		},
 		{
-			"finalize",
-			markercli.GetCmdFinalize(),
+			"mint supply",
```
SpicyLemon, line 724: It'd be good to add a second `mint` test case here that has a recipient (one with a recipient, and one without).

x/marker/client/cli/tx.go
```diff
@@ -177,26 +177,32 @@ with the given supply amount and denomination provided in the coin argument
...
 		Short:   "Mint coins against the marker",
 		Long: strings.TrimSpace(`Mints coins of the marker's denomination and places them
 in the marker's account under escrow.  Caller must possess the mint permission and 
 marker must be in the active status.`),
-		Example: fmt.Sprintf(`$ %s tx marker mint 1000hotdogcoin --from mykey`, version.AppName),
+		Example: fmt.Sprintf(`$ %s tx marker mint 1000hotdogcoin recipient_address --from mykey`, version.AppName),
```
SpicyLemon, line 187:
```suggestion
		Example: fmt.Sprintf(`$ %s tx marker mint 1000hotdogcoin pb1gghjut3ccd8ay0zduzj64hwre2fxs9ldmqhffj --from mykey`, version.AppName),
```

Since this is an example, it should have real-looking values instead of field names. That address is used in several other examples, so it should work good here too.

x/marker/keeper/msg_server.go
```diff
@@ -270,6 +270,20 @@ func (k msgServer) Mint(goCtx context.Context, msg *types.MsgMintRequest) (*type
...
 	}
 
+	if len(msg.Recipient) > 0 {
+		recipient, err := sdk.AccAddressFromBech32(msg.Recipient)
+		if err != nil {
+			return nil, sdkerrors.ErrInvalidAddress.Wrapf("invalid recipient: %s", msg.Recipient)
+		}
+		if recipient.Empty() {
+			return nil, sdkerrors.ErrInvalidAddress.Wrap("recipient address is zero or empty")
+		}
```
SpicyLemon, line 280: This check can be removed. `AccAddressFromBech32` will either return a non-empty address, or an error.

SpicyLemon (approved):

docs/proto-docs.md
SpicyLemon: Adding a comment, then will resolve again in the hopes that it clears the "1 requested change."

Issue #2252, opened 2024-12-31
Query nft coin commitments in the exchange module


rgipson-figure:
## Summary

Requesting a way to query if an nft coin is committed to a market in the exchange module.

## Problem Definition

There currently isn't a way to query if an nft coin has been committed to a market. 

The closest query options available are [getaccountcommitments](https://developer.provenance.io/docs/sdk/exchange/queries#getaccountcommitments) and [getholds](https://developer.provenance.io/docs/sdk/hold/queries#getholds), but it will return all the coins for the account.

Our use case is our loan onboarding process will commit a loan nft to a market and confirm it is in the market before proceeding with the next onboarding steps. Our other services will also need the ability to check the commitment status of a loan for warehousing the loan.

## Proposal

Options:

1. Add an optional query param of `denom: string` in [getaccountcommitments](https://developer.provenance.io/docs/sdk/exchange/queries#getaccountcommitments)
2. Add an optional query param of `denom: string` in [getholds](https://developer.provenance.io/docs/sdk/hold/queries#getholds)

Up for discussing other options or best practices for this use case.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

nagarajdivine:
Hi @jdfigure  @rgipson-figure, imo  
1. Adding `optional(denom)` `query params` to [getaccountcommitments](https://developer.provenance.io/docs/sdk/exchange/queries#getaccountcommitments)  is good options, bcoz it directly queries market commitments, but we might need to updated at client end. 
2. [getholds](https://developer.provenance.io/docs/sdk/hold/queries#getholds) which is not directly tied up to market commitments not a good option. 
3. Adding new dedicated end-point `getnftcommitment` is straightforward, but it query same as above `getaccountcommitments` So multiple API for querying same thing is not good solution. 

open for any suggestion :)

PR #2334, opened 2025-04-30
feat(nft-commitment-query-by-denom): Enhanced getaccountcommitments  to support querying coin commitments with an optional asset filter.

nagarajdivine:
## Description

Enhanced [getaccountcommitments](https://developer.provenance.io/docs/sdk/exchange/queries#getaccountcommitments)   to support querying coin commitments with an optional asset filter.

closes: #2252 

---

write a little note why.


x/exchange/client/cli/query_setup.go
```diff
@@ -341,6 +342,9 @@ func MakeQueryGetAccountCommitments(_ client.Context, flagSet *pflag.FlagSet, ar
 
 	var err error
 	rv.Account, err = ReadStringFlagOrArg(flagSet, args, FlagAccount, "account")
+	if asset, err := flagSet.GetString("asset"); err == nil && asset != "" {
+		rv.XAsset = &exchange.QueryGetAccountCommitmentsRequest_Asset{Asset: asset}
+	}
 
 	return rv, err
```
SpicyLemon, line 349: One thing to watch for in go is that errors are being properly checked. With your change, the `err` from reading the account (line 344) is now being ignored. Usually, we'd handle that with an `if err != nil` check right after, but in these makers, we read all the fields and return multiple errors if needed. So instead of all those `if err != nil` checks, we create a `[]error` and put each field's error in a specific index and, at the end, combine them all using `errors.Join(...)` (which ignores nil entries). See `MakeQueryGetMarketCommitments` for an example.

Once you remove the `optional` keyword from the proto field, this'll be easier too.

```suggestion
	errs := make([]error, 2)
	rv.Account, errs[0] = ReadStringFlagOrArg(flagSet, args, FlagAccount, "account")
	rv.Denom, errs[1] = flagSet.GetString(FlagDenom)

	return rv, errors.Join(errs...)
```

x/exchange/client/cli/query_setup_test.go
```diff
@@ -793,6 +793,7 @@ func TestSetupCmdQueryGetAccountCommitments(t *testing.T) {
 		expExamples: []string{
 			exampleStart + " " + cli.ExampleAddr,
 			exampleStart + " --account " + cli.ExampleAddr,
+			exampleStart + " --asset nash",
```
SpicyLemon, line 796: After fixing the `AddQueryExample`, you'll need to fix it here too:
```suggestion
			exampleStart + " " + cli.ExampleAddr + " --asset nash",
```

x/exchange/spec/05_queries.md
```diff
@@ -178,6 +178,7 @@ To find out how much an account has committed to a market, use the `GetCommitmen
 ## GetAccountCommitments
 
 To look up the amounts an account has committed to any market, use the `GetAccountCommitments` query.
+You can optionally filter the results by a specific asset denomination using the `asset` query parameter.
```
SpicyLemon, line 181:
```suggestion
You can optionally filter the results for a specific denomination using the `denom` query parameter.
```

proto/provenance/exchange/v1/query.proto
```diff
@@ -68,8 +68,9 @@ service Query {
   }
 
   // GetAccountCommitments gets all the funds in an account that are committed to any market.
+  // Optionally, you can filter the results by a specific asset denomination using the `asset` query parameter.
   rpc GetAccountCommitments(QueryGetAccountCommitmentsRequest) returns (QueryGetAccountCommitmentsResponse) {
-    option (google.api.http).get = "/provenance/exchange/v1/commitments/account/{account}";
+    option (google.api.http).get = "/provenance/exchange/v1/commitments/account/{account}?asset={asset}";
```
SpicyLemon, line 73: You don't want the `?asset={asset}` part here. The client/server stuff automatically adds query parameters for stuff that doesn't have a placeholder in the string.

In other words, if the path is `"/provenance/exchange/v1/commitments/account/{account}"`, it'll still work just fine. If there isn't an `asset` value limiting the query, the client won't include a query parameter for it; then, in the server, it'll leave the value as an empty string in `QueryGetAccountCommitmentsRequest `. If there **is* an asset, the client will notice that there isn't a placeholder for it, and add it to the URL as a query parameter which the server then sees and sets in the request.

proto/provenance/exchange/v1/query.proto
```diff
@@ -68,8 +68,9 @@ service Query {
   }
 
   // GetAccountCommitments gets all the funds in an account that are committed to any market.
+  // Optionally, you can filter the results by a specific asset denomination using the `asset` query parameter.
```
SpicyLemon, line 71:
```suggestion
  // Optionally, you can filter the results for a specific denomination using the `denom` query parameter.
```

.changelog/unreleased/features/2252-nft-commitment-query-by-denom.md
```diff
@@ -0,0 +1,5 @@
+* Enhanced `GetAccountCommitments` to support querying NFT coin commitments with optional asset filter in the exchange module [#2252](https://github.com/provenance-io/provenance/issues/2252).
+* The `GetAccountCommitments` query now lets you filter commitments by a specific asset, thanks to an optional asset parameter.
+* The CLI command `SetupCmdQueryGetAccountCommitments` was updated to include a --asset flag, making it easier to filter commitments by asset directly from the command * line.
```
SpicyLemon, line 3: Instead of `SetupCmdQueryGetAccountCommitments`, use the actual command that people would be typing, `query exchange account-commitments`. You'll also want to tweak the wording to not use the word "asset". And there's an extra `*` toward the end of the CLI entry.

SpicyLemon (changes requested):
Even though there's a bunch of comments here, it looks pretty good. Most of the comments are about changing the name of the new field to `denom`, but there's a couple other things too.

x/exchange/client/cli/query_setup.go
```diff
@@ -321,7 +321,7 @@ func MakeQueryGetCommitment(_ client.Context, flagSet *pflag.FlagSet, _ []string
 // SetupCmdQueryGetAccountCommitments adds all the flags needed for MakeQueryGetAccountCommitments.
 func SetupCmdQueryGetAccountCommitments(cmd *cobra.Command) {
 	cmd.Flags().String(FlagAccount, "", "The account's address")
-
+	cmd.Flags().String("asset", "", "Optional asset denom to filter commitments")
 	AddUseArgs(cmd,
 		fmt.Sprintf("{<account>|--%s <account>}", FlagAccount),
 	)
```
SpicyLemon, line 327: You should add a new entry for the newly added flag.

```suggestion
	AddUseArgs(cmd,
		fmt.Sprintf("{<account>|--%s <account>}", FlagAccount),
		fmt.Sprintf("[--%s <denom>]", FlagDenom),
	)
```

SpicyLemon (dismissed):
Looks good!

SpicyLemon (approved):

SpicyLemon (approved):

Issue #2322, opened 2025-04-09
Exchange: add a new Market Transfer Commitment message


jtalis:
## Summary

Exchange: add a new Market Transfer Commitment message for moving committed funds between exchanges 

## Problem Definition

Currently to move funds between exchanges requires the funds to be uncommitted from exchange 1 by a signer with exchange permissions and then the account signer to commit the funds to exchange 2. The use case we are trying to solve for is to perform this as a single transaction so we can ensure the funds are always encumbered by an exchange. To build a simple user experience around this currently requires requesting the user to sign an authz grant to commit the funds on their behalf ahead of time, so that the exchange can then initiate the transfer in a single provenance transaction that releases the commitment from exchange 1 and commits to exchange 2. 

## Proposal

Add a new Market Transfer Commitment message to allow moving committed funds between exchanges

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
Pretty straight-forward I think.

Create a new `MarketTransferCommitment` endpoint, with these fields:
* `admin` (the signer, must have `cancel` permission on the market that currently has the funds).
* `current_market_id`
* `new_market_id`
* `account`
* `amount`
* `creation_fee` (?)
* `event_tag`

It'll remove the amount from the current commitment record (deleting it if it's now zero), and add the amount to the account's commitment record with the new market (creating it if needed). Since no new funds are being committed, we can skip releasing and holding the amount.

The only question to iron out is about the `creation_fee`. When committing funds, that creation fee is paid by the user. But since the user isn't signing this new msg, we shouldn't be taking funds out of their account. And there's no guarantee that the `amount` has funds in the denom needed for the fee, so we can't depend on paying it out of committed funds. That means that the market would then be responsible for paying them, which feels like something they wouldn't be okay with. That being said, Figure doesn't have or want commitment creation fees, and they're the only market(s) out there right now. So maybe, for starters, we leave out the `creation_fee` field here until it's something that we know we want, and have figured out how it should be paid.

iramiller:
What’s the feeling on user funds control for this movement between markets? Seems like it could be an unexpected change if the new market had a substantially different fee structure for example… is there any log of user intention associated with this?

SpicyLemon:
[...]

PR #2338, opened 2025-05-06
feat/market-transfer-commitments: (Exchange) Added a new message to transfer committed funds between exchanges.

nagarajdivine:
## Description

* Added a new message to transfer committed funds between exchanges.

closes: #2322 

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "Market Transfer Commitment" feature, allowing administrators to transfer committed funds directly between markets via a new endpoint and CLI command.
- **Documentation**
  - Updated user and developer documentation to describe the new market transfer commitment functionality and permissions.
  - Improved and clarified descriptions for several API parameters and message fields.
- **Bug Fixes**
  - Corrected minor wording and formatting issues in documentation and API descriptions.
- **Tests**
  - Added comprehensive tests for the new market transfer commitment functionality, including CLI, message validation, and keeper logic.


x/exchange/client/cli/flags.go
```diff
@@ -68,6 +68,8 @@ const (
 	FlagIcon                 = "icon"
 	FlagInputs               = "inputs"
 	FlagMarket               = "market"
+	FlagCurrentMarket        = "current-market"
+	FlagNewMarket            = "new-market"
```
SpicyLemon, line 72: Let's keep this list ordered alphabetically by the variable name.

x/exchange/keeper/commitments.go
```diff
@@ -417,6 +417,37 @@ func (k Keeper) SettleCommitments(ctx sdk.Context, req *exchange.MsgMarketCommit
...
+			req.Amount, currentAmount, account, req.CurrentMarketId)
+	}
+	// subtract requested amount and store the new balance for the current market
+	setCommitmentAmount(store, req.CurrentMarketId, sdk.AccAddress(req.Account), newAmt)
+
+	// update the commitment to new market for given account
+	addCommitmentAmount(store, req.NewMarketId, account, req.Amount)
+
+	k.emitEvent(ctx, exchange.NewEventCommitmentTransferred(req.Account, req.Amount, req.CurrentMarketId, req.NewMarketId, req.EventTag))
+	return nil
```
SpicyLemon, line 448: For consistency, we should still emit the `EventCommitmentReleased` and `EventFundsCommitted` events, and we can probably get rid of the `EventCommitmentTransferred`.

Reasoning:
Events are consumed by external services in order react to what's happening on chain. Currently, if they need to monitor commitments, they'd be looking for `EventCommitmentReleased` and `EventFundsCommitted`. If this uses those, then they don't need to change anything in how they look for commitment-related stuff. With the addition of the `EventCommitmentTransferred`, now they'd have to look for a third type to handle specially. And due to the nature of that type of stuff, there's a good chance they won't notice that they're not handling everything until after it starts causing a problem. So, even though emitting those two events is more verbose than this new one, it's probably best to just stick with those two.

x/exchange/keeper/commitments.go
```diff
@@ -426,3 +457,43 @@ func (k Keeper) consumeCommitmentSettlementFee(ctx sdk.Context, req *exchange.Ms
...
+func (k Keeper) validateTransferCommitments(
+	ctx sdk.Context,
+	store storetypes.KVStore,
+	req *exchange.MsgMarketTransferCommitmentsRequest,
+	account sdk.AccAddress,
+) error {
+	// Blocked address check
+	if k.bankKeeper.BlockedAddr(account) {
+		return fmt.Errorf("%s is not allowed to receive funds", account)
+	}
```
SpicyLemon, line 471: We don't need to check this; no funds are actually being moved, so we don't need to worry about a blocked account receiving any funds.

x/exchange/keeper/commitments.go
```diff
@@ -426,3 +457,43 @@ func (k Keeper) consumeCommitmentSettlementFee(ctx sdk.Context, req *exchange.Ms
...
+	// Zero amount
+	if req.Amount.IsZero() {
+		return fmt.Errorf("account %s does not have any funds committed to market %d", account, req.CurrentMarketId)
+	}
```
SpicyLemon, line 475:
```suggestion
	if req.Amount.IsZero() {
		return fmt.Errorf("cannot transfer zero for %s in market %d", account, req.CurrentMarketId)
	}
```

The `req.Amount` is how much they're asking to transfer, not how much they actually have committed.

x/exchange/keeper/commitments.go
```diff
@@ -426,3 +457,43 @@ func (k Keeper) consumeCommitmentSettlementFee(ctx sdk.Context, req *exchange.Ms
...
+
+	// Negative amount
+	if req.Amount.IsAnyNegative() {
+		return fmt.Errorf("cannot transfer negative commitment amount %q for %s in market %d", req.Amount, req.Account, req.CurrentMarketId)
+	}
+
+	// Market validity
+	if err := validateMarketIsAcceptingCommitments(store, req.CurrentMarketId); err != nil {
+		return fmt.Errorf("current market %d is invalid: %w", req.CurrentMarketId, err)
+	}
```
SpicyLemon, line 485: The current market doesn't need to be accepting commitments for this.

Use case: A market used to allow commitments, but has since disabled them and now they want to transfer the existing commitments to another market.

x/exchange/keeper/commitments.go
```diff
@@ -426,3 +457,43 @@ func (k Keeper) consumeCommitmentSettlementFee(ctx sdk.Context, req *exchange.Ms
...
+	if err := validateMarketIsAcceptingCommitments(store, req.NewMarketId); err != nil {
+		return fmt.Errorf("new market %d is invalid: %w", req.NewMarketId, err)
+	}
+
+	// User permission
+	if err := k.validateUserCanCreateCommitment(ctx, req.CurrentMarketId, account); err != nil {
+		return fmt.Errorf("account %s does not have permission to create commitments in market %d: %w", account, req.CurrentMarketId, err)
+	}
```
SpicyLemon, line 493: We're not creating a commitment in the current market, so we don't need to check this.

Use case: A user used to be able to commit funds to a market, but now, no longer has the required attributes, which is why they want their commitment transferred to a new market (that they're allowed to use).

proto/provenance/exchange/v1/tx.proto
```diff
@@ -48,6 +48,9 @@ service Msg {
   // MarketReleaseCommitments is a market endpoint return control of funds back to the account owner(s).
   rpc MarketReleaseCommitments(MsgMarketReleaseCommitmentsRequest) returns (MsgMarketReleaseCommitmentsResponse);
 
+  // MarketTransferCommitments is a market endpoint to transfers committed funds from one market to another.
+  rpc MarketTransferCommitments(MsgMarketTransferCommitmentsRequest) returns (MsgMarketTransferCommitmentsResponse);
```
SpicyLemon, line 52: Let's just call this `MarketTransferCommitment` (singular) since it only transfers a single commitment. Same with the related messages and methods.

SpicyLemon: Alternatively, we could replace the `account` and `amount` fields with a `repeated AccountAmount transfers` (in the request `Msg`). That'd let a market transfer commitments for multiple accounts in a single message. It's okay that it'd only allow one market to one other. There should be some cli stuff already created for getting `AccountAmount` info via CLI.

This adds a bit of complexity though, so it's up to you.

x/exchange/keeper/msg_server.go
```diff
@@ -139,6 +139,19 @@ func (k MsgServer) MarketReleaseCommitments(goCtx context.Context, msg *exchange
 	return &exchange.MsgMarketReleaseCommitmentsResponse{}, nil
 }
 
+// MarketTransferCommitments is a market endpoint transfer funds from one market to another market back to the account owner(s).
```
SpicyLemon, line 142:
```suggestion
// MarketTransferCommitments is a market endpoint to transfers committed funds from one market to another.
```

x/exchange/spec/03_messages.md
```diff
@@ -299,6 +300,28 @@
...
+
+It is expected to fail if:
+* The currentmarket does not exist.
+* The newmarket does not exist.
+* The `admin` does not have `PERMISSION_CANCEL` in the market, and is not the `authority`.
+* One or more of the amounts is more than what is currently committed by the associated `account`.
+
+#### MsgMarketTransferCommitmentsRequest
+
++++ https://github.com/provenance-io/provenance/blob/v1.20.0/proto/provenance/exchange/v1/tx.proto#L324-L344
```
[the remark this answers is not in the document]
SpicyLemon: You are wrong here. First, `MsgMarketTransferCommitmentsRequest` is new, so it does exist with this PR. Second, stop complaining about these links. They are correct for us.

x/exchange/spec/03_messages.md
```diff
@@ -299,6 +300,26 @@
...
+
+#### MsgMarketTransferCommitmentsResponse
+
++++ https://github.com/provenance-io/provenance/blob/v1.20.0/proto/provenance/exchange/v1/tx.proto#L346-L347
```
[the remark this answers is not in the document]
SpicyLemon: This is correct, and you don't need to complain about links like this in spec docs.

x/exchange/spec/04_events.md
```diff
@@ -147,6 +148,21 @@ Event Type: `provenance.exchange.v1.EventCommitmentReleased`
...
 
 
+## EventCommitmentTransferred
+
+When funds are transferred from one market to another, an `EventCommitmentTransferred` is emitted.
+
+Event Type: `provenance.exchange.v1.EventCommitmentTransferred`
+
+| Attribute Key     | Attribute Value                                                          |
+|---------------    |--------------------------------------------------------------------------|
```
SpicyLemon, line 158:
```suggestion
|-------------------|--------------------------------------------------------------------------|
```

proto/provenance/exchange/v1/tx.proto
nagarajdivine: Here we are moving fund between the markets, since this is just single account transfer, using above `account` and `amount`, which simple and straightforward. IMO using `AccountAmount` it might introduce extra complexity.

SpicyLemon (approved):

Issue #2311, opened 2025-03-28
Error Querying TXs with Old WASM Messages


jdfigure:
## Summary of Bug

When using TX queries where messages containing a `/cosmwasm.wasm.v1beta1.MsgExecuteContract` would be returned is causing a client crash.

## Version

1.21

## Steps to Reproduce

```
provenanced q txs --query "message.sender='pb1ktgsj6ap4cfq45g5dyw5vj47zhcrd6cl9p6z43'" --node tcp://10.10.32.16:26657 --output json | jq .
Error: unable to resolve type URL /cosmwasm.wasm.v1beta1.MsgExecuteContract: tx parse error
```

Or via a grpc client (i.e. Kotlin): 
```
io.grpc.StatusRuntimeException: INTERNAL: unable to resolve type URL /cosmwasm.wasm.v1beta1.MsgExecuteContract: tx parse error
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:268)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:249)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:167)
	at cosmos.tx.v1beta1.ServiceGrpc$ServiceBlockingStub.getTxsEvent(ServiceGrpc.java:640)
```

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

SpicyLemon:
The issue here is that the msg used to execute the contract is the old `/cosmwasm.wasm.v1beta1.MsgExecuteContract` msg (the new one is `/cosmwasm.wasm.v1.MsgExecuteContract`, `v1` instead of `v1beta1`). So the codec doesn't know how to unpack the msg, causing this bug.

The solution will be to register the old v1beta1 types with the codec. However, the wasm library doesn't generate the golang types for the `v1beta1` stuff anymore. So this won't be an easy fix.

PR #2361, opened 2025-05-28
fix:(wasm-v1beta1): Added compatibility for deprecated WASM v1beta1 messages

nagarajdivine:
## Description

* Added compatibility for deprecated WASM v1beta1 messages.
closes: #2311 

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Added support for the legacy MsgExecuteContract message type to ensure backward compatibility when decoding and processing older transactions.
  - Introduced documentation for the MsgExecuteContract message, including detailed field descriptions.
- **Bug Fixes**
  - Registered the legacy MsgExecuteContract type to maintain proper decoding of transactions using the older format.
- **Tests**
  - Added comprehensive tests to verify correct registration, packing, and unpacking of MsgExecuteContract messages, including scenarios involving governance proposals and unknown message types.


.changelog/unreleased/bug-fixes/2311-register-wasm-v1beta1-types.md
```diff
@@ -0,0 +1 @@
+* register legacy CosmWasm v1beta1 MsgExecuteContract to support older tx decoding [#2311](https://github.com/provenance-io/provenance/issues/2311).
```
SpicyLemon: Please capitalize the first character since changelog entries are supposed to be sentences.

x/wasm/types/interface_registry.go
```diff
@@ -1,14 +1,20 @@
...
 	"github.com/cosmos/gogoproto/proto"
+
+	v1beta1 "github.com/provenance-io/provenance/x/wasm"
 )
 
 var _ codectypes.InterfaceRegistry = &WasmInterfaceRegistry{}
 
 // NewWasmInterfaceRegistry returns a new WasmInterfaceRegistry instance
 func NewWasmInterfaceRegistry(registry codectypes.InterfaceRegistry) WasmInterfaceRegistry {
+	registry.RegisterImplementations((*sdk.Msg)(nil), &v1beta1.MsgExecuteContract{}, &wasmv1.MsgExecuteContract{})
```
SpicyLemon: You only need to register the v1beta1 version here. The other one is registered by the wasm module itself.

x/wasm/types/interface_registry_test.go
```diff
@@ -0,0 +1,153 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			// Pack into Any
+			anyMsg, err := types.NewAnyWithValue(tt.msg)
+			require.NoError(t, err)
+			require.Equal(t, tt.typeURL, anyMsg.TypeUrl)
+			var unpacked sdk.Msg
+			err = cdc.UnpackAny(anyMsg, &unpacked)
+			require.NoError(t, err)
+			require.NotNil(t, unpacked)
+			require.IsType(t, tt.msg, unpacked)
```
SpicyLemon: It's best to always include a message in test assertions, specially when you've got multiple assertions. It's even more important when you have multiple assertions of the same type, e.g two `NoError` assertions. Doing so makes it significantly easier to address failures.

There's two main things these messages do: 1. Provide context in the failure message, and 2. Makes it easier to identify exactly which assertion failed.

A lot of times, we're seeing these failures in a Github action runner, or somewhere else outside of our IDE. With a little context in those failure messages, you can often jump directly to the code that needs fixing. Without any extra context, you usually have no choice but to start by finding the unit test code, and it also usually requires more effort to identify exactly which assertion is failing so that you can finally go to the code that needs fixing.

It is crucial to ALWAYS include a message with a `NotNil` assertion. When that assertion fails, all it says is "Expected value not to be nil." That is not helpful at all. Unless you're already familiar with that test, there is no way at all that you'll know what the problem is (or where to fix it) without reading the test.

It's also very important to always include a message when there are multiple assertions of the same type, e.g. two `NoError` assertions. Without a message, it's sometimes possible to identify the failing assertion based on its type. E.g. If the failure says an error wasn't expected, but it got one, you know it's probably a `NoError` assertion (or maybe some other assertion involving an error). The failure message is what we're most likely to remember as we start to address the failure. If we can't tell what's actually wrong from the message, we then head to the test code. If there are multiple assertions, though, we need some way to figure out which one is the offending one. Without extra context in the failure, all we can do is go back to the failure output (often in another window/app) to identify the line number, then go back to the test code. But with a message in the assertions, there's a string on the line we need that's probably something in our memory since it's part of the failure message, so we don't have to switch windows/apps.

If you make it a habit to always include a message when writing an assertion, it's an unnoticeable amount of extra work, but it makes life easier in the future. In some cases, you'll be able to skip reading the test code; but for the times where you do still need to start with the test code, having the extra context makes it easier to find which part of the test is failing.

```suggestion
			// Pack into Any
			anyMsg, err := types.NewAnyWithValue(tt.msg)
			require.NoError(t, err, "NewAnyWithValue")
			require.Equal(t, tt.typeURL, anyMsg.TypeUrl, "anyMsg.TypeUrl")
			
			// Unpack it.
			var unpacked sdk.Msg
			err = cdc.UnpackAny(anyMsg, &unpacked)
			require.NoError(t, err, "UnpackAny")
			require.NotNil(t, unpacked, "UnpackAny result")
			require.IsType(t, tt.msg, unpacked, "UnpackAny result")
```

x/wasm/types/interface_registry_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// Create an Any with a completely unknown type URL
+	unknownAny := &types.Any{
+		TypeUrl: "/some.unknown.MsgType",
+		Value:   []byte("garbage"),
+	}
+
+	var msg sdk.Msg
+	err := cdc.UnpackAny(unknownAny, &msg)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "no registered implementations")
```
SpicyLemon: There's a `ErrorContains` assertion that would be better to use here.

Also, `Error` is another assertion that should always have a message on it. All the standard failure message gives you is "An error is expected but got nil." which isn't very helpful.

That being said, usually there's a better assertion to use than `Error`, e.g. `EqualError` or `ErrorContains`. There are a few rare spots where `Error` is still handy, but it's usually not the right thing to use.

x/wasm/types/interface_registry_test.go
```diff
@@ -0,0 +1,153 @@
...
+			if v1msg, ok := tt.msg.(*wasmv1.MsgExecuteContract); ok {
+				unpackedV1, _ := unpacked.(*wasmv1.MsgExecuteContract)
+				require.Equal(t, v1msg.Sender, unpackedV1.Sender)
+				require.Equal(t, v1msg.Contract, unpackedV1.Contract)
+			}
+			if beta1msg, ok := tt.msg.(*v1beta1.MsgExecuteContract); ok {
+				unpackedBeta1, _ := unpacked.(*v1beta1.MsgExecuteContract)
+				require.Equal(t, beta1msg.Sender, unpackedBeta1.Sender)
+				require.Equal(t, beta1msg.Contract, unpackedBeta1.Contract)
+			}
```
SpicyLemon: It's okay to mix `assert` and `require` stuff in tests.

* `require` is used when there is nothing else in the test worth doing if the assertion fails. E.g. `NotNil` is often good to have as `require` because there's often follow-ups that can only be done if the thing isn't nil.
* `assert` is used when there's other things that can be checked. E.g. `Equal` is often best as an assertion because you don't usually need two things to be equal in order to test other aspects of things.

By mixing them, you get a bigger picture of what's wrong. This allows you to fix it all in one pass. But if you only use `require`, the test will halt at the first failure, so that's all you know is wrong. You fix that then run the test again only to find out that the next thing was also wrong and needed fixing. So you end up with this annoying back and forth where you're making a change, running the test, making another change, etc. sometimes without a clear picture of how many more iterations it's going to take.

Also, the `assert` version of the assertions all return a `bool` that is `true` if the assertion passed. E.g. with `if assert.Equal(t, a, b) { ... }`, the code block will execute if `a` and `b` are equal. This pattern is useful in tests where you're checking multiple things, but a subset of the assertions should only be checked if possible. E.g. `if assert.NotNil(t, a) { assert.Equal(t, expX, a.FieldX); assert.Equal(t, expY, a.FieldY); }`

With this test, the top stuff (through `require.IsType`) is good as `require` since, at each one, there really isn't anything more to test if the assertion fails. However, the bottom stuff would be better as `assert`. That way, if both the sender and contract are different, you know it.

Lastly, here's another example of where including a message in the assertions would greatly help you in the future. If any of the `Equal` assertions fail the failure message won't tell you what's being compared. So you'll have to start by looking at the test code. You know two things aren't equal, but you get in here and see that there are five  equals assertions. So you have to then go back to the test run to identify the line number from the output and come back to the test code to finally find the problem assertion.

x/wasm/types/interface_registry_test.go
nagarajdivine: totally agree :)

Issue #2346, opened 2025-05-20
Bump golangci-lint and golangci/golangci-lint-action


SpicyLemon:
## Summary

Bump the version of `golangci-lint` to a more recent version and also `golangci/golangci-lint-action` too.

## Problem Definition

Dependabot created this PR to bump `golangci/golangci-lint-action` to 8 (from 6).
* #2335 

However, that action only works on newer versions of `golangci-lint`. So we'll need to bump both together.

## Proposal

Bump both `golangci-lint` and `golangci/golangci-lint-action` to newer versions.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2363, opened 2025-05-29
Bump golangci-lint to v2.1.6.

nagarajdivine:
## Description

* Bump golangci/golangci-lint-action from 6 to 8 and golangci-lint to v2.1.6.
closes: #2346 

---

write a little note why.

## Summary by CodeRabbit

* **New Features**
  * Added a calculate-tx-fees CLI command (aliases: simulate, calculate-fees) and a --do-not-convert flag.
  * Added new TX command to update conversion factors for fee handling.

* **Bug Fixes**
  * Strengthened validation for flatfees parameters, conversion factors, and message-fee updates.

* **Chores**
  * Upgraded golangci-lint and consolidated linting configuration; numerous lint annotations and docs updated across the codebase.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/x/marker/client/cli/tx.go
+++ b/x/marker/client/cli/tx.go
@@ -1801,11 +1801,11 @@ From stdin:
 				}
 
 			case strings.HasPrefix(authInput, "@"):
-				filePath := filepath.Clean(strings.TrimPrefix(authInput, "@"))
+				filePath := strings.TrimPrefix(authInput, "@")
 				if filePath == "" {
 					return fmt.Errorf("missing file path after '@'")
 				}
-
+				filePath = filepath.Clean(filePath)
 				f, err := os.Open(filePath) // Use filePath directly
 				if err != nil {
 					if os.IsNotExist(err) {


.changelog/unreleased/dependencies/2346-upgrade-golangci-lint.md
```diff
@@ -0,0 +1 @@
+* golangci/golangci-lint-action` bumped to 8 (from 6) and golangci-lint v2.1.6 [#2346](https://github.com/provenance-io/provenance/issues/2346).
```
SpicyLemon: These should be split into multiple entries. The dependencies section is a little special in our changlog, and works best if there's only one thing per line.

```suggestion
* `golangci/golangci-lint-action` bumped to 8 (from 6) [#2346](https://github.com/provenance-io/provenance/issues/2346).
* `golangci-lint` bumped to v2.1.6 (from v1.60) [#2346](https://github.com/provenance-io/provenance/issues/2346).
```

.github/workflows/lint.yml
```diff
@@ -31,13 +31,14 @@ jobs:
...
+      - uses: golangci/golangci-lint-action@v8
         if: env.GIT_DIFF
         with:
           # If you change this version, be sure to also change it in contrib/devtools/Makefile.
-          version: v1.60
-          args: --timeout 10m --out-${NO_FUTURE}format colored-line-number
+          version: v2.1.6
+          args: --timeout 10m --output.text.path=stdout --output.text.colors=true
           github-token: ${{ secrets.github_token }}
+          only-new-issues: true
```
SpicyLemon, line 41: We shouldn't do this. From our terminals, we should be able to run `make lint` without it complaining about anything. With this `only-new-issues: true`, any existing problems won't be identified by the github action, but `make lint` will fail locally.

Usually, when we bump `golangcli-lint`, we get a new slew of failures that all need to be fixed (or get a `nolint` comment). Some of them can be fixed automatically using the `--fix` flag when invoking `golangcli-lint`. The `make lint-fix` target will do this too. It probably won't be able to fix everything, but it'll get some of it.

.golangci.yml
```diff
@@ -39,15 +35,19 @@ linters:
...
     - tparallel
-    - typecheck
     - unconvert
     - unparam
     - unused
     - whitespace
-
+    # - stylecheck
+    # - tenv
+    # - gosimple
```
SpicyLemon: Are these not available anymore? They look available, but maybe they don't work on this version?

* https://pkg.go.dev/honnef.co/go/tools/stylecheck
* https://github.com/sivchari/tenv
* https://pkg.go.dev/github.com/surullabs/lint/gosimple

It'd be good to keep those if possible.

nagarajdivine: `stylecheck`, `tenv` & `gosimple` does not work in new version.

.github/workflows/lint.yml
nagarajdivine: `only-new-issues: true` if we disable this, linter reports a large number of existing issues, including many from external dependencies like the Cosmos SDK

SpicyLemon: I just copied this branch and removed this `only-new-issues: true` flag to see what the errors are. There aren't any for code in another repo, it's all about problems in our own code.

From what I can tell, the github action is not properly identifying and/or using our `.golangci.yml` file. For example, it shows a `depguard` failure, claiming that `cosmossdk.io/log` isn't allowed, but it's in our allow list. So something isn't happening the way it should.

Once that's ironed out, I expect that there are some new failures because of either a new linter or changes to existing ones. Those need to be addressed by either fixing them or adding an appropriate nolint directive comment.

But right now, it appears that this `only-new-issues: true` flag is hiding a large problem, that it's not actually linting like it should.

.golangci.yml
```diff
@@ -140,15 +131,114 @@ linters-settings:
...
+          - github.com/provenance-io/provenance/x/trigger/client/cli
+          - github.com/provenance-io/provenance/x/trigger/simulation
+          - github.com/provenance-io/provenance/testutil
+          - github.com/provenance-io/provenance/testutil/assertions
+          - github.com/provenance-io/provenance/testutil/contracts
+          - github.com/provenance-io/provenance/testutil/queries
+          - github.com/provenance-io/provenance/testutil
+          - github.com/provenance-io/provenance/x/attribute
+          - github.com/provenance-io/provenance/x/attribute/keeper
+          - github.com/provenance-io/provenance/x/attribute/types
```
SpicyLemon: We shouldn't need all of these. The `github.com/provenance-io/provenance` entry should cover all of these.

app/app.go
```diff
@@ -125,7 +120,7 @@ import (
 	ibckeeper "github.com/cosmos/ibc-go/v8/modules/core/keeper"
 	ibctm "github.com/cosmos/ibc-go/v8/modules/light-clients/07-tendermint"
 	ibctestingtypes "github.com/cosmos/ibc-go/v8/testing/types"
-
+	"github.com/gorilla/mux"
 	simappparams "github.com/provenance-io/provenance/app/params"
```
SpicyLemon, line 135: If this passes the linter, the `gci` linter isn't set up correctly. It should require a space between the non-provenance imports and the provenance imports.

nagarajdivine: @SpicyLemon I will go through it again, linter was giving many `gci` issues whenever I run `golangci-lint run -v` in v2.1.6, after fixing the space, issue was fixed.

app/export.go
```diff
@@ -69,14 +66,14 @@ func (app *App) ExportAppStateAndValidators(forZeroHeight bool, jailAllowedAddrs
...
-		ConsensusParams: app.BaseApp.GetConsensusParams(ctx),
+		ConsensusParams: app.GetConsensusParams(ctx),
 	}, err
 }
 
 // prepare for fresh start at zero height
 // NOTE: zero height genesis is a temporary feature which will be deprecated in favor of export at a block height
 func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []string) {
-	applyAllowedAddrs := false
+	applyAllowedAddrs := false //nolint:staticcheck // QF1007
```
SpicyLemon: **issue (blocking)**: **All** nolint directives need to answer the question, "Why are we ignoring the linter here?"

The linters exist to make our lives easier. If we're ignoring them, we need to state why.

app/params/encoding.go
```diff
@@ -1,3 +1,4 @@
+// Package params handles app encoding.
```
SpicyLemon: **question**: Which linter is requiring these comments?

I'm not a fan of them since they don't really add anything useful.

**suggestion**: If a single linter is responsible, lets disable that linter. If it's part of another linter, let's see if we can turn off just that check.

app/prefix.go
```diff
@@ -5,21 +5,28 @@ import (
 )
 
 const (
+	// AccountAddressPrefixMainNet is the mainnet address prefix.
```
SpicyLemon: **question**: We don't need to require comments on all constants. Please disable that check, and maybe undo some of the comment additions.

E.g. This comment doesn't provide any information that I can't get directly from the name of the variable.

Plus, without the comments, `gofmt` will align the `=` and you get a nice comparison of values that becomes harder when there's a comment between them and the values are no longer vertically lined up either.

app/test_helpers.go
```diff
@@ -512,15 +510,13 @@ func NewPubKeyFromHex(pk string) (res cryptotypes.PubKey) {
...
+	tempDir := t.TempDir()
+	tempApp := New(
+		log.NewNopLogger(),
+		dbm.NewMemDB(),
+		nil,
+		true,
+		simtestutil.NewAppOptionsWithFlagHome(tempDir),
+	)
 	return tempApp.GetEncodingConfig()
 }
```
SpicyLemon, line 526: **issue (blocking)**: This still needs to be able to handle when `t` is `nil`.

What was the linter complaining about?

client/docs/embed.go
```diff
@@ -1,6 +1,9 @@
+// Package docs provides embedded Swagger UI files for API documentation.
 package docs
 
 import "embed"
 
+// SwaggerUI holds the embedded Swagger UI files.
+//
```
SpicyLemon: **issue**: This is another example of a comment that should not be required.

app/export.go
SpicyLemon: This applies to all new nolint directives, not just this one. All of the ones that I saw were insufficient and need to be expanded upon.

app/app.go
SpicyLemon: Then the `gci` linter either isn't set up correctly, or the config for it isn't getting used. The `gci` linter needs to require a space here between the `default` section and the `prefix(github.com/provenance-io)` section. 

The way the imports are right now makes it look like the linter is ignoring the `prefix(github.com/provenance-io)` entry and just lumping all the provenance imports in with the `default` ones. This is also what the default behavior of that linter would look like. That's not what we want. We want our stuff separated out and everyone else's lumped together.

---

Once you change the linter settings, you can use `make lint-fix` to make the linter automatically update the orders of all the imports (and fix some other stuff). If running `golangci-lint` directly, add the `--fix` flag to do the same thing.

app/test_helpers.go
nagarajdivine: @SpicyLemon this is linter error : 
```
os.MkdirTemp() could be replaced by t.TempDir() in MakeTestEncodingConfig (usetesting)
tempDir, err := os.MkdirTemp("", "tempprovapp")
```

app/params/encoding.go
nagarajdivine: linter was complaining about (revive)
```
package-comments: should have a package comment (revive)
package params
```

app/export.go
nagarajdivine: here also same linter error :(review), most of comments, which I have added all related to linter (revive)
```
app/prefix.go:8:2: exported: exported const AccountAddressPrefixMainNet should have comment (or a comment on this block) or be unexported (revive)
	AccountAddressPrefixMainNet = "pb"
	^
app/prefix.go:21:2: exported: exported var AccountAddressPrefix should have comment or be unexported (revive)
	AccountAddressPrefix   = AccountAddressPrefixMainNet

```

app/test_helpers.go
SpicyLemon: We should keep this the way it was and put a nolint directive on it. Something like `//nolint:usetesting // Not using t.TempDir() here to allow t to be nil and so the dir is cleaned up sooner.`

app/prefix.go
nagarajdivine: here also same linter error :(revive), most of comments, which I have added all related to linter (revive)
```
app/prefix.go:8:2: exported: exported const AccountAddressPrefixMainNet should have comment (or a comment on this block) or be unexported (revive)
	AccountAddressPrefixMainNet = "pb"
	^
app/prefix.go:21:2: exported: exported var AccountAddressPrefix should have comment or be unexported (revive)
	AccountAddressPrefix   = AccountAddressPrefixMainNet
```

app/params/encoding.go
SpicyLemon: We'll need to configure `revive` a bit then: https://golangci-lint.run/docs/linters/configuration/#revive and https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md

At the very least, we'll want to disable `package-comments` since none of our stuff uses them. I'm sure there are others that can be disabled, but I don't know which ones they'd be.

app/prefix.go
SpicyLemon: I kind of like the `exported` linter, so I'd rather not disable it. When I look at its description, it sounds like you should be able to put a single comment above the `const` line that satisfies that linter, though. That'd be better than a comment on each field like this.

If the single comment above `const` doesn't satisfy the linter, consider disabling just the `disableChecksOnConstants` sub-sub-check on the `exported` sub-checker in `revive`.

app/export.go
SpicyLemon: I'm confused by this one. The thing `QF1007` is checking for seems to be what we've got.

Maybe try moving this line down between lines 78 and 79 to see if that fixes the lint complaint.

.golangci.yml
```diff
@@ -39,15 +35,20 @@ linters:
...
     - tparallel
-    - typecheck
     - unconvert
     - unparam
     - unused
     - whitespace
-
+    - usetesting
+    # - stylecheck
+    # - tenv
```
SpicyLemon, line 45: Let's delete these instead of having them commented out.

.golangci.yml
```diff
@@ -39,15 +35,20 @@ linters:
...
+    # - gosimple
```
SpicyLemon: Let's delete these instead of having them commented out.

.golangci.yml
```diff
@@ -71,84 +72,49 @@ issues:
...
+          # Provenance
+          - github.com/provenance-io/provenance
+          # CosmWasm
+          - github.com/CosmWasm
+          # Cosmossdk.io
+          - cosmossdk.io
+          # Cosmos
+          - github.com/cosmos
+          # CometBFT
+          - github.com/cometbft
```
SpicyLemon: Let's just put these entries under a `# Primary` comment.

Each of these comments doesn't add anything but noise. I.e. The comment doesn't contain any information that the line below it doesn't already provide.

.golangci.yml
```diff
@@ -162,9 +128,6 @@ linters-settings:
     sections:
       - standard
       - default
-      - prefix(github.com/cometbft)
-      - prefix(cosmossdk.io)
-      - prefix(github.com/cosmos)
       - prefix(github.com/provenance-io)
       - blank
       - dot
```
SpicyLemon: Does this maybe need to go under `formatter.settings`? That might be why the linter doesn't seem to be applying these customizations.

client/docs/embed.go
SpicyLemon: Nevermind. This one is fine.

app/test_helpers.go
```diff
@@ -512,14 +514,18 @@ func NewPubKeyFromHex(pk string) (res cryptotypes.PubKey) {
...
 	switch {
 	case t != nil:
 		require.NoError(t, err, "failed to create temp dir %q", tempDir)
 	case err != nil:
 		panic(fmt.Errorf("failed to create temp dir %q: %w", tempDir, err))
 	}
-	defer os.RemoveAll(tempDir)
+	defer func() {
+		if err := os.RemoveAll(tempDir); err != nil {
+			fmt.Printf("failed to remove temp dir %s: %v\n", tempDir, err)
```
SpicyLemon: We don't actually care about this error. It'd be better to go back to `defer os.RemoveAll(tempDir) //nolint:errcheck`

Additionally, I think this is one of the rare cases where you don't really need an explanation with the nolint directive. Specifically, I'm saying that any of the `defer` lines that the `errcheck` linter is complaining about can be nolinted without also listing a reason. Any `errcheck` problems on non-defer lines should still have an explanation, though.

x/attribute/keeper/keeper.go
```diff
@@ -117,7 +117,11 @@ func (k Keeper) IterateRecords(ctx sdk.Context, prefix []byte, handle Handler) e
 	// Init an attribute record iterator
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, prefix)
-	defer iterator.Close()
+	defer func() {
+		if err := iterator.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateRecords", "error", err)
+		}
+	}()
```
SpicyLemon: Let's `nolint` this one instead. We never really care about these errors as they don't affect anything.

x/attribute/keeper/keeper.go
```diff
@@ -537,7 +541,7 @@ func (k Keeper) DeleteExpiredAttributes(ctx sdk.Context, limit int) int {
 	for ; iterator.Valid(); iterator.Next() {
 		expirationKeys = append(expirationKeys, iterator.Key())
 	}
-	iterator.Close()
+	iterator.Close() //nolint:errcheck,gosec
```
SpicyLemon: What is `gosec` complaining about here? This line should include the reason we can ignore that linter here.

x/attribute/types/keys.go
```diff
@@ -25,8 +25,8 @@ const (
 	AccountDataName = "accountdata"
 )
 
+// Legacy amino encoded objects use this key prefix.
 var (
-	// Legacy amino encoded objects use this key prefix
 	AttributeKeyPrefixAmino      = []byte{0x00}
```
SpicyLemon: **issue**: This comment only applies to the  `AttributeKeyPrefixAmino` variable. It doesn't apply to any of the other variables, though, so it doesn't make sense to put it on the `var` line. 

**suggestion**: Move the comment back to the `AttributeKeyPrefixAmino` variable (and fix it there), then add a new comment above `var` that's something like `// Data store key prefixes.`

x/marker/client/cli/tx.go
```diff
@@ -1390,6 +1394,7 @@ func ParseAccessGrantFromString(addressPermissionString string) []types.AccessGr
 		}
 		partsPerAddress := strings.Split(p, ",")
 		// if it has an address has to have at least one access associated with it
+		//nolint:staticcheck // QF1001: intentional logic for clarity
 		if !(len(partsPerAddress) > 1) {
```
SpicyLemon: This would be a good fix to apply.

```suggestion
		if len(partsPerAddress) <= 1 {
```

x/marker/keeper/keeper.go
```diff
@@ -373,7 +373,11 @@ func (k Keeper) GetNetAssetValue(ctx sdk.Context, markerDenom, priceDenom string
 func (k Keeper) IterateNetAssetValues(ctx sdk.Context, markerAddr sdk.AccAddress, handler func(state types.NetAssetValue) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.NetAssetValueKeyPrefix(markerAddr))
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateNetAssetValues", "error", err)
+		}
+	}()
```
SpicyLemon: This one's better as a nolint directive too.

x/marker/keeper/keeper.go
```diff
@@ -390,7 +394,11 @@ func (k Keeper) IterateNetAssetValues(ctx sdk.Context, markerAddr sdk.AccAddress
 func (k Keeper) IterateAllNetAssetValues(ctx sdk.Context, handler func(sdk.AccAddress, types.NetAssetValue) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.NetAssetValuePrefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateAllNetAssetValues", "error", err)
+		}
+	}()
```
SpicyLemon: Same here, let's just nolint this one.

x/metadata/keeper/objectstore.go
```diff
@@ -67,7 +67,11 @@ func (k Keeper) IterateOSLocators(ctx sdk.Context, cb func(account types.ObjectS
 	prefix := types.OSLocatorAddressKeyPrefix
 	it := storetypes.KVStorePrefixIterator(store, prefix)
 
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateRecords", "error", err)
+		}
+	}()
```
SpicyLemon: This one is okay as a nolint:errcheck directive.

x/metadata/keeper/record.go
```diff
@@ -99,7 +99,11 @@ func (k Keeper) IterateRecords(ctx sdk.Context, scopeID types.MetadataAddress, h
 		return err
 	}
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateRecords", "error", err)
+		}
+	}()
```
SpicyLemon: Also, this one is okay as a nolint directive.

x/metadata/keeper/scope.go
```diff
@@ -18,7 +18,11 @@ import (
 func (k Keeper) IterateScopes(ctx sdk.Context, handler func(types.Scope) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.ScopeKeyPrefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateScopes", "error", err)
+		}
+	}()
```
SpicyLemon: Here.

x/metadata/keeper/scope.go
```diff
@@ -34,7 +38,11 @@ func (k Keeper) IterateScopesForAddress(ctx sdk.Context, address sdk.AccAddress,
 	store := ctx.KVStore(k.storeKey)
 	prefix := types.GetAddressScopeCacheIteratorPrefix(address)
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateScopesForAddress", "error", err)
+		}
+	}()
```
SpicyLemon: here

x/metadata/keeper/scope.go
```diff
@@ -54,7 +62,11 @@ func (k Keeper) IterateScopesForScopeSpec(ctx sdk.Context, scopeSpecID types.Met
 	store := ctx.KVStore(k.storeKey)
 	prefix := types.GetScopeSpecScopeCacheIteratorPrefix(scopeSpecID)
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateScopesForScopeSpec", "error", err)
+		}
+	}()
```
SpicyLemon: here

x/metadata/keeper/scope.go
```diff
@@ -186,7 +198,11 @@ func (k Keeper) RemoveScope(ctx sdk.Context, id types.MetadataAddress) error {
 	store := ctx.KVStore(k.storeKey)
 	prefix, _ := id.ScopeRecordIteratorPrefix() // Can't return an error because we know it's a valid scope id.
 	iter := storetypes.KVStorePrefixIterator(store, prefix)
-	defer iter.Close()
+	defer func() {
+		if err := iter.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close RemoveScope", "error", err)
+		}
+	}()
```
SpicyLemon, line 195: and here.

SpicyLemon (commented):
These are all okay as `nolint:errcheck` directives.

x/metadata/keeper/scope.go
```diff
@@ -897,7 +913,11 @@ func (k Keeper) SetNetAssetValue(ctx sdk.Context, scopeID types.MetadataAddress,
 func (k Keeper) IterateNetAssetValues(ctx sdk.Context, scopeID types.MetadataAddress, handler func(state types.NetAssetValue) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.NetAssetValueKeyPrefix(scopeID))
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateNetAssetValues", "error", err)
+		}
+	}()
```
SpicyLemon: Also okay as a nolint directive.

x/metadata/keeper/scope.go
```diff
@@ -918,7 +938,7 @@ func (k Keeper) RemoveNetAssetValues(ctx sdk.Context, scopeID types.MetadataAddr
 	for ; it.Valid(); it.Next() {
 		keys = append(keys, it.Key())
 	}
-	it.Close()
+	it.Close() //nolint:errcheck,gosec
```
SpicyLemon: Why does `gosec` also complain about this?

x/metadata/keeper/session.go
```diff
@@ -87,7 +87,11 @@ func (k Keeper) IterateSessions(ctx sdk.Context, scopeID types.MetadataAddress,
 		return err
 	}
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateSessions", "error", err)
+		}
+	}()
```
SpicyLemon: Can just be a nolint directive.

x/metadata/keeper/specification.go
```diff
@@ -16,7 +16,11 @@ import (
 func (k Keeper) IterateRecordSpecs(ctx sdk.Context, handler func(specification types.RecordSpecification) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.RecordSpecificationKeyPrefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateRecordSpecs", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -61,7 +65,11 @@ func (k Keeper) IterateRecordSpecsForContractSpec(ctx sdk.Context, contractSpecI
 		return err
 	}
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateRecordSpecsForContractSpec", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -162,7 +170,11 @@ func (k Keeper) isRecordSpecUsed(_ sdk.Context, _ types.MetadataAddress) bool {
 func (k Keeper) IterateContractSpecs(ctx sdk.Context, handler func(specification types.ContractSpecification) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.ContractSpecificationKeyPrefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateContractSpecs", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -181,7 +193,11 @@ func (k Keeper) IterateContractSpecsForOwner(ctx sdk.Context, ownerAddress sdk.A
 	store := ctx.KVStore(k.storeKey)
 	prefix := types.GetAddressContractSpecCacheIteratorPrefix(ownerAddress)
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateContractSpecsForOwner", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -366,7 +382,11 @@ func (k Keeper) ValidateWriteContractSpecification(_ sdk.Context, existing *type
 func (k Keeper) IterateScopeSpecs(ctx sdk.Context, handler func(specification types.ScopeSpecification) (stop bool)) error {
 	store := ctx.KVStore(k.storeKey)
 	it := storetypes.KVStorePrefixIterator(store, types.ScopeSpecificationKeyPrefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateScopeSpecs", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -385,7 +405,11 @@ func (k Keeper) IterateScopeSpecsForOwner(ctx sdk.Context, ownerAddress sdk.AccA
 	store := ctx.KVStore(k.storeKey)
 	prefix := types.GetAddressScopeSpecCacheIteratorPrefix(ownerAddress)
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close IterateScopeSpecsForOwner", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/metadata/keeper/specification.go
```diff
@@ -403,7 +427,11 @@ func (k Keeper) IterateScopeSpecsForContractSpec(ctx sdk.Context, contractSpecID
 	store := ctx.KVStore(k.storeKey)
 	prefix := types.GetContractSpecScopeSpecCacheIteratorPrefix(contractSpecID)
 	it := storetypes.KVStorePrefixIterator(store, prefix)
-	defer it.Close()
+	defer func() {
+		if err := it.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close iterator", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as nolint.

x/name/keeper/keeper.go
```diff
@@ -240,7 +240,12 @@ func (k Keeper) IterateRecords(ctx sdk.Context, prefix []byte, handle func(recor
 	// Init a name record iterator
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, prefix)
-	defer iterator.Close()
+	defer func() {
+		if err := iterator.Close(); err != nil {
+			k.Logger(ctx).Error("failed to close iterator", "error", err)
+		}
+	}()
```
SpicyLemon: Okay as a nolint directive.

x/trigger/keeper/event_listener.go
```diff
@@ -44,7 +44,11 @@ func (k Keeper) IterateEventListeners(ctx sdk.Context, eventName string, handle
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, triggertypes.GetEventListenerPrefix(eventName))
 
-	defer iterator.Close()
+	defer func() {
+		if err := iterator.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close iterator", "error", err)
+		}
+	}()
```
SpicyLemon: This one can be a nolint directive.

x/trigger/keeper/migrations_v2.go
```diff
@@ -31,7 +31,9 @@ func (m Migrator) DeleteGasLimits(ctx sdk.Context, logger log.Logger) {
 	iter := storetypes.KVStorePrefixIterator(store, GasLimitKeyPrefix)
 	closeIter := func() {
 		if iter != nil {
-			iter.Close()
+			if err := iter.Close(); err != nil {
+				m.keeper.Logger(ctx).Error("Failed to close iterator", "error", err)
+			}
```
SpicyLemon: This one can also be a nolint directive.

x/trigger/keeper/queue.go
```diff
@@ -93,7 +93,11 @@ func (k Keeper) iterateQueue(ctx sdk.Context, handle func(trigger types.QueuedTr
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, types.QueueKeyPrefix)
 
-	defer iterator.Close()
+	defer func() {
+		if err := iterator.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close iterator", "error", err)
+		}
+	}()
```
SpicyLemon: Let's just nolint this one.

x/trigger/keeper/trigger.go
```diff
@@ -44,7 +44,11 @@ func (k Keeper) IterateTriggers(ctx sdk.Context, handle func(trigger types.Trigg
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, types.TriggerKeyPrefix)
 
-	defer iterator.Close()
+	defer func() {
+		if err := iterator.Close(); err != nil {
+			k.Logger(ctx).Error("Failed to close iterator", "error", err)
+		}
+	}()
```
SpicyLemon: We can nolint this one.

SpicyLemon (commented):
I've made comments on the several places where we should go back to what we had with a nolint directive.

Most are part of `defer` statements where we just don't care about the error. Logging the error doesn't really help anything since it's not actually a problem, but also because the only people that might see that error are the node operators, and it's just going to add confusion.

There are still many places where we need an explanation as to why we're ignoring the linter in that place; the nolint directive on its own is insufficient. I haven't commented on all of those places though. You'll need to go through all the ones you've added and add the reasons.

x/attribute/keeper/keeper.go
nagarajdivine:
```
iterator.Close() //nolint:errcheck,gosec
G104: Errors unhandled (gosec)
```

x/metadata/keeper/scope.go
nagarajdivine:
```
iterator.Close() //nolint:errcheck,gosec
G104: Errors unhandled (gosec)
```

app/export.go
```diff
@@ -79,9 +79,7 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []str
 	applyAllowedAddrs := false
 
 	// check if there is a allowed address list
-	if len(jailAllowedAddrs) > 0 {
-		applyAllowedAddrs = true
-	}
+	applyAllowedAddrs = len(jailAllowedAddrs) > 0
```
SpicyLemon: No need to set this to false and then to something else immediately after.

testutil/ibc/testchain.go
```diff
@@ -161,7 +161,7 @@ func SignAndDeliver(
 	chainID string, accNums, accSeqs []uint64, _ bool, blockTime time.Time, nextValHash []byte, priv ...cryptotypes.PrivKey,
 ) (*abci.ResponseFinalizeBlock, error) {
 	tx, err := simtestutil.GenSignedMockTx(
-		rand.New(rand.NewSource(1)),
+		rand.New(rand.NewSource(1)), //nolint:gosec // G304
```
SpicyLemon: G304 doesn't make sense for this line, is that the right thing to put here?

G304 is about a tained filepath being used, but this is just a random number generator line.

x/attribute/keeper/keeper.go
```diff
@@ -117,7 +117,9 @@ func (k Keeper) IterateRecords(ctx sdk.Context, prefix []byte, handle Handler) e
 	// Init an attribute record iterator
 	store := ctx.KVStore(k.storeKey)
 	iterator := storetypes.KVStorePrefixIterator(store, prefix)
-	defer iterator.Close()
+
+	iterator.Close() //nolint:errcheck,gosec // close error safe to ignore in this context
+
```
SpicyLemon, line 121: **issue (blocking)**: This still needs to be a defer.

x/attribute/types/keys.go
```diff
@@ -118,7 +118,7 @@ func GetNameKeyBytes(name string) []byte {
 func GetAttributeExpireTimePrefix(expireTime time.Time) []byte {
 	key := AttributeExpirationKeyPrefix
 	expireTimeBz := make([]byte, 8)
-	binary.BigEndian.PutUint64(expireTimeBz, uint64(expireTime.Unix()))
+	binary.BigEndian.PutUint64(expireTimeBz, uint64(expireTime.Unix())) //nolint:gosec // G304: safe fixed size buffer write.
```
SpicyLemon: Here too, I don't think this is a G304, is it?

.golangci.yml
```diff
@@ -37,17 +32,92 @@ linters:
...
+    dogsled:
+      max-blank-identifiers: 3
+    misspell:
+      locale: US
+      mode: default
+      ignore-rules:
+      - cancelled
+    nolintlint:
+      allow-unused: false
+      require-explanation: false
```
SpicyLemon: Let's change this to "true". I'm always going to require one in PRs anyway.

testutil/ibc/testchain.go
```diff
@@ -161,7 +161,7 @@ func SignAndDeliver(
 	chainID string, accNums, accSeqs []uint64, _ bool, blockTime time.Time, nextValHash []byte, priv ...cryptotypes.PrivKey,
 ) (*abci.ResponseFinalizeBlock, error) {
 	tx, err := simtestutil.GenSignedMockTx(
-		rand.New(rand.NewSource(1)),
+		rand.New(rand.NewSource(1)), //nolint:gosec
```
SpicyLemon: Needs an explanation.

x/metadata/keeper/msg_server.go
```diff
@@ -652,13 +652,13 @@ func (k msgServer) BindOSLocator(
...
-	if k.Keeper.OSLocatorExists(ctx, ownerAddress) {
+	if k.OSLocatorExists(ctx, ownerAddress) {
 		ctx.Logger().Error("Address already bound to an URI", "owner", msg.Locator.Owner)
 		return nil, sdkerrors.ErrInvalidRequest.Wrap(types.ErrOSLocatorAlreadyBound.Error())
 	}
 
 	// Bind owner to URI
-	if err := k.Keeper.SetOSLocator(ctx, ownerAddress, encryptionKey, msg.Locator.LocatorUri); err != nil {
+	if err := k.SetOSLocator(ctx, ownerAddress, encryptionKey, msg.Locator.LocatorUri); err != nil {
 		ctx.Logger().Error("unable to bind name", "err", err)
```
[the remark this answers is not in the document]
SpicyLemon: We'll do this later.

x/trigger/keeper/migrations_v2.go
SpicyLemon, line 1: This file was deleted, and must have gotten re-added during the catch-up merge. Please delete it again.

x/marker/client/cli/tx.go
```diff
@@ -1800,7 +1801,7 @@ From stdin:
 				}
 
 			case strings.HasPrefix(authInput, "@"):
-				filePath := strings.TrimPrefix(authInput, "@")
+				filePath := filepath.Clean(strings.TrimPrefix(authInput, "@"))
 				if filePath == "" {
 					return fmt.Errorf("missing file path after '@'")
```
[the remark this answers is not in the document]
SpicyLemon: @nagarajdivine This needs to be addressed.

Issue #1588, opened 2023-06-13
Allow definition of concrete attribute type


SpicyLemon:
## Summary

Allow the definition of the concrete attribute type when the type is proto or json.

## Problem Definition

As a writer of attributes, I want to be able to provide a type identifier (e.g. a type URL) when setting some attribute values that have type `ATTRIBUTE_TYPE_JSON` or `ATTRIBUTE_TYPE_PROTO` (or maybe others too), so that people reading the attributes know more about what the value is.

As a reader of attribute values, I want to be able to know more about the structure being represented in attributes of type `ATTRIBUTE_TYPE_JSON` and `ATTRIBUTE_TYPE_PROTO` (and maybe others too), so that I can more easily write code that uses attribute values.

## Proposal

Add a new optional `concrete_type` string field to an attribute.

Allow the field to be used for any `attribute_type`, and it should never be required.

Limit the length to something reasonable that is unlikely to interfere (100? 150? 250?). Consider adding a param for the limit, but that might be overkill; a hardcoded number is probably fine.

An made up example of what the field is expected to hold: `/figure.provenance.attributes.v1.KnowYourCustomer` (that's 49 characters btw). There might be other, more verbose formats out there that people will want to use to identify the types, though. So maybe look around to see what all is done for that elsewhere to decide on a good length limit.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2369, opened 2025-06-04
feat(attribute): support optional concrete_type field for attributes

nagarajdivine:
## Description

* support optional concrete_type field for attributes
closes: #1588 

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Added support for an optional "concrete_type" field to attributes, allowing explicit specification of attribute value types (up to 200 characters) for PROTO or JSON attributes.
  - The "concrete_type" field is now available in CLI commands, API schemas, and message definitions for adding and updating attributes.

- **Documentation**
  - Updated API documentation, protobuf message documentation, and CLI examples to describe and illustrate the new "concrete_type" field.

- **Bug Fixes**
  - Enhanced validation to ensure the "concrete_type" field does not exceed 200 characters.

- **Tests**
  - Expanded test coverage to verify correct handling, validation, and persistence of the "concrete_type" field in attributes.


client/docs/swagger-ui/swagger.yaml
```diff
@@ -220,6 +220,14 @@ paths:
...
                       format: date-time
                       description: Time that an attribute will expire.
+                    concrete_type:
+                      type: string
+                      description: >-
+                        concrete_type defines the specific type of data for
+                        PROTO or JSON attribute.
+
+
+                        Optional. Max 200 characters.
```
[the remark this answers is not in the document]
SpicyLemon: This `client/docs/swagger-ui/swagger.yaml` file is auto-generated and you should ignore it.

proto/provenance/attribute/v1/attribute.proto
```diff
@@ -28,6 +28,8 @@ message Attribute {
   string address = 4;
   // Time that an attribute will expire.
   google.protobuf.Timestamp expiration_date = 5 [(gogoproto.stdtime) = true, (gogoproto.nullable) = true];
+  // concrete_type defines the specific type of data for PROTO or JSON attribute.
+  string concrete_type = 6; // Optional. Max 200 characters.
```
SpicyLemon: Please don't split comments like this.

A lot of things (IDEs, proto-docs-gen, etc.) parse these files and do stuff with the comments. But that can get messed up if a field has both a comment above it and one after it. It's best to just put them both together above the field.

x/attribute/client/cli/tx.go
```diff
@@ -91,23 +90,29 @@ Refer to %s tx name bind --help for more information on how to do this.`, versio
 				msg.ExpirationDate = &expireTime
 			}
 
+			concreteType, _ := cmd.Flags().GetString("concrete-type")
```
SpicyLemon: Please extract magic value "concrete-type" to a variable for use here and a few lines below when you add the flag to the command.

x/attribute/client/cli/tx.go
```diff
@@ -49,9 +49,9 @@ func NewAddAccountAttributeCmd() *cobra.Command {
 		Short:   "Add an account attribute to the provenance blockchain",
 		Long: fmt.Sprintf(`Note: the attribute name must have already been created through the name module.  
 Refer to %s tx name bind --help for more information on how to do this.`, version.AppName),
-		Args: cobra.RangeArgs(4, 5),
+		Args: cobra.RangeArgs(4, 6),
```
SpicyLemon: The new concrete type is being provided by flag, not positional args, so this should go back to `cobra.RangeArgs(4, 5)`.

x/attribute/client/cli/tx.go
```diff
@@ -91,23 +90,29 @@ Refer to %s tx name bind --help for more information on how to do this.`, versio
 				msg.ExpirationDate = &expireTime
 			}
 
+			concreteType, _ := cmd.Flags().GetString("concrete-type")
+			if len(concreteType) > 200 {
+				return fmt.Errorf("concrete-type length must be less than or equal to 200 characters")
+			}
+			if concreteType != "" {
+				msg.ConcreteType = concreteType
+			}
```
SpicyLemon: The length validation should be part of `ValidateBasic()` which is done automatically in `GenerateOrBroadcastTxCLI`, so you don't need to also do that check here.

```suggestion
			msg.ConcreteType, _ := cmd.Flags().GetString("concrete-type")
```

x/attribute/client/cli/tx.go
```diff
@@ -91,23 +90,29 @@ Refer to %s tx name bind --help for more information on how to do this.`, versio
...
 	flags.AddTxFlagsToCmd(cmd)
-
 	return cmd
 }
 
 // NewUpdateAccountAttributeCmd creates a command for adding an account attributes.
 func NewUpdateAccountAttributeCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "update <name> <address> <original-type> <original-value> <update-type> <update-value>",
+		Use:     "update <name> <address> <original-type> <original-value> <update-type> <update-value> [concrete-type]",
```
SpicyLemon: Let's just add the `--concrete-type` flag to this command too and leave the positional args unchanged.

x/attribute/keeper/keeper_test.go
```diff
@@ -1279,3 +1280,185 @@ func (s *KeeperTestSuite) TestSetAccountData() {
...
+			if tc.expectError {
+				s.Assert().Error(err)
+				return
+			}
+			s.Assert().NoError(err)
+			s.Assert().Equal(tc.attributeName, savedAttr[0].Name)
+			s.Assert().Equal(tc.attributeType, savedAttr[0].AttributeType)
+			s.Assert().Equal(tc.value, savedAttr[0].Value)
+			s.Assert().Equal(s.user1Addr.String(), savedAttr[0].Address)
+			s.Assert().Equal(tc.concreteType, savedAttr[0].ConcreteType)
```
SpicyLemon: Always always always include messages in your test assertions. Make it a habit to always add a message as you're writing the assertion line. Without the extra context that a message provides, identifying what's going wrong requires tracking down the test, identifying (usually by line number) the offending line, then identifying the field being tested. If you included that field as a message in the assertion, you can skip all those steps. This makes it significantly easier to maintain stuff.

x/attribute/keeper/keeper_test.go
```diff
@@ -1279,3 +1280,185 @@ func (s *KeeperTestSuite) TestSetAccountData() {
...
+
```
SpicyLemon: There's some trailing empty lines in here that can be removed.

x/attribute/keeper/keeper_test.go
```diff
@@ -1279,3 +1280,185 @@ func (s *KeeperTestSuite) TestSetAccountData() {
...
+	savedAttr, _ = s.app.AttributeKeeper.GetAttributes(s.ctx, s.user1, noTypeAttr.Name)
+	s.Assert().Equal(noTypeAttr.Value, savedAttr[0].Value)
+
+}
+
+func (s *KeeperTestSuite) TestValidateAttributeConcreteType() {
+	testCases := []struct {
+		name         string
+		concreteType string
+		expectValid  bool
```
SpicyLemon: This is a bad pattern to use.

Instead of a `bool` to indicate if something is expected to be valid (or invalid), you should just have an `expErr string` field. If the string is empty, no error is expected (i.e. it's expected to be valid). If the string has content, the error is expected to equal it (i.e. it's expected to be invalid). A lot of our old tests use this bad pattern, but nothing new should use it.

We have a helper for one-lining the error assertion depending on the content (or lack of content) of the expected string: `assertions.AssertErrorValue(t, err, tc.expErr, "...")`

The difference is that, with a string, you're being explicit about HOW the thing is expected to be invalid. There's usually multiple reasons why something is invalid and it's important to make sure that you're actually testing what  you think you're testing. By using an expected error string, you get that assurance. If validation shifts, and your test case is no longer testing what it was set up to test, it'll fail because the error message changed.

And you don't need both a `bool` and `string` for this. An empty string indicates that no error is expected; a non-empty string indicates an error is expected. Having both leads to problems in test setups where someone is adding tests and doesn't know that you have to use both a bool and string to indicate failure, leading to improper tests setups that look like they're doing what they should and passing, but actually are demonstrating (without telling us) that something is broken.

There's only a couple times where it's appropriate to use a bool for test setup to indicate success/failure:
1. The thing being tested only returns a bool (i.e. doesn't return an error to test against).
2. The error is coming from an external source and is difficult (or impossible) to fully predict. In this case, we often use a `[]string` and make sure the error has all of those as sub-strings.

x/attribute/types/attribute.go
```diff
@@ -66,6 +67,9 @@ func (a Attribute) ValidateBasic() error {
 	if !isValidValueForType(a.AttributeType, a.Value) {
 		return fmt.Errorf("invalid attribute value for assigned type: %s", a.AttributeType)
 	}
+	if len(a.ConcreteType) > 200 {
```
SpicyLemon, line 70: Just leaving a note here.

The longest msg-type URL is 80 characters long. So 200 seem like a reasonable max to me.

x/attribute/types/msgs.go
```diff
@@ -36,7 +36,10 @@ func (msg MsgAddAttributeRequest) ValidateBasic() error {
 	if _, err := sdk.AccAddressFromBech32(msg.Owner); err != nil {
 		return err
 	}
-	a := NewAttribute(msg.Name, msg.Account, msg.AttributeType, msg.Value, msg.ExpirationDate)
+	if len(msg.ConcreteType) > 200 {
+		return fmt.Errorf("concrete_type exceeds maximum length of 200 characters")
+	}
+	a := NewAttribute(msg.Name, msg.Account, msg.AttributeType, msg.Value, msg.ExpirationDate, msg.ConcreteType)
 	return a.ValidateBasic()
```
SpicyLemon: Since you're calling `a.ValidateBasic()` here, and that does the length check on the concrete type, you don't also need to validate it here. Same with below.

It's even more crucial to remove these duplicate validation pieces because that 200 is hard coded in each instead of being a variable.

SpicyLemon (changes requested):

proto/provenance/attribute/v1/tx.proto
```diff
@@ -55,6 +55,8 @@ message MsgAddAttributeRequest {
   string owner = 5;
   // Time that an attribute will expire.
   google.protobuf.Timestamp expiration_date = 6 [(gogoproto.stdtime) = true, (gogoproto.nullable) = true];
+  // concrete_type defines the specific type of data for PROTO or JSON attribute.
+  string concrete_type = 7; // Optional. Max 200 characters.
```
SpicyLemon: Here too with the two different comment placements. Also down on line 85.

x/attribute/client/cli/tx.go
```diff
@@ -51,7 +55,7 @@ func NewAddAccountAttributeCmd() *cobra.Command {
 Refer to %s tx name bind --help for more information on how to do this.`, version.AppName),
 		Args: cobra.RangeArgs(4, 5),
 		Example: fmt.Sprintf(`$ %s tx attribute add "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value"
-		$ %s tx attribute add "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" 2050-01-15T00:00:00Z`, version.AppName, version.AppName),
+		$ %s tx attribute add "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" 2050-01-15T00:00:00Z provenance.attributes.v1.TestJSON`, version.AppName, version.AppName),
```
SpicyLemon:
```suggestion
		$ %s tx attribute add "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" --concrete-type provenance.attributes.v1.TestJSON`, version.AppName, version.AppName),
```

x/attribute/client/cli/tx.go
```diff
@@ -106,7 +109,7 @@ func NewUpdateAccountAttributeCmd() *cobra.Command {
 		Use:     "update <name> <address> <original-type> <original-value> <update-type> <update-value>",
 		Aliases: []string{"u"},
 		Short:   "Update an account attribute on the provenance blockchain",
-		Example: fmt.Sprintf(`$ %s tx attribute update "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" "int" 100`, version.AppName),
+		Example: fmt.Sprintf(`$ %s tx attribute update "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" "int" 100 provenance.attributes.v1.TestJSON`, version.AppName),
```
SpicyLemon:
```suggestion
		Example: fmt.Sprintf(`$ %s tx attribute update "attr1.pb" tp1jypkeck8vywptdltjnwspwzulkqu7jv6ey90dx "string" "test value" "int" 100 --concrete-type provenance.attributes.v1.TestJSON`, version.AppName),
```

x/attribute/types/msgs.go
```diff
@@ -36,7 +36,10 @@ func (msg MsgAddAttributeRequest) ValidateBasic() error {
 	if _, err := sdk.AccAddressFromBech32(msg.Owner); err != nil {
 		return err
 	}
-	a := NewAttribute(msg.Name, msg.Account, msg.AttributeType, msg.Value, msg.ExpirationDate)
+	if len(msg.ConcreteType) > 200 {
+		return fmt.Errorf("concrete_type exceeds maximum length of 200 characters")
+	}
```
SpicyLemon: You don't need this check since the attribute validation (called a couple lines below) will do that.

proto/provenance/attribute/v1/tx.proto
SpicyLemon: Looks like you got the lower comment, but not this one.

SpicyLemon (approved):

Issue #2207, opened 2024-11-04
Create a multi-authorization.


SpicyLemon:
## Summary

Create a `MultiAuthorization` with a `msg` (type url) and a list of sub-authorizations.

## Problem Definition

Currently, we have a few different authz authorization types, but each is narrowly defined and sometimes we want it to have extra restrictions.

## Proposal

Create a `MultiAuthorization` with a `msg` (type url) and a list of `sub_authorizations`.

When accepting a `MultiAuthorization`, call `Accept` on each sub-authorization. If any return a `delete`, the parent also returns that. If any return an `update`, the parent should return an updated version of itself with the updated sub-authorizations. The parent only returns `accept` if all sub-messages also do.

The `MsgTypeUrl` of each `sub_authorization` must equal the parent's.

The `sub_authorizations` should be `repeated Any` unless there's a more specific type or interface that we can use for it.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
I'm torn on whether to allow `sub_authorizations` to have only one entry, or if we should require it to have at least two.

My initial thought was that it should have at least two since otherwise, it'd just be that authorization. But then I realized that the parent counts as a `GenericAuthorization`, so there might be a use where we need one of these with only one sub-authorization. Then again, all authorizations are extensions of a `GenericAuthorization`, so maybe that shouldn't count like this.

iramiller:
Since the name has Multi in it then it would be reasonable to expect a single to not work imo… but ultimately it seems like whichever leads to the simplistic/least code solution should be used.

ethsman:
Hello @SpicyLemon @iramiller , sorry for leaving a comment here without respecting the team rules. As a blockchain developer, I think my tech stack is fit for a team project. So I want to prove my skills by solving some team issues and join the team as a full-time developer. Could you give me a test task?

PR #2379, opened 2025-06-12
feat/(multi-authorization) authz: support multiple sub-authorizations via MultiAuthorization

nagarajdivine:
## Description

* Support multiple sub-authorizations via MultiAuthorization
closes: #2207 

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Introduced MultiAuthorization, enabling multiple sub-authorizations to collectively approve a message.
  - Added CLI commands for granting and revoking MultiAuthorization with support for JSON input from inline, file, or stdin.
- **Documentation**
  - Enhanced documentation to explain MultiAuthorization’s structure, usage, and constraints.
- **Tests**
  - Added comprehensive tests covering MultiAuthorization creation, validation, acceptance logic, CLI commands, and protobuf codec handling.


docs/proto-docs.md
```diff
@@ -8576,6 +8577,23 @@ a marker transfer on behalf of the granter's account.
...
+### MultiAuthorization
+MultiAuthorization lets you combine several authorizations.
+All sub-authorizations must accept the message for it to be allowed.
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| `msg_type_url` | [string](#string) |  | The message type this authorization is for. |
+| `sub_authorizations` | [google.protobuf.Any](#google-protobuf-Any) | repeated | A list of sub-authorizations that must all accept the message. sub_authorizations: a list of authorizations (minimum 2, maximum 10). |
+
```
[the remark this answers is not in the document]
SpicyLemon: The `docs/proto-docs.md` file is automatically generated. You can stop making formatting suggestions about it.

And you're wrong about the link being non-existent because you're ignoring that it's markdown (and not html).

x/marker/spec/11_authorization.md
```diff
@@ -22,4 +23,26 @@ message MarkerTransferAuthorization {
 With the `MarkerTransferAuthorization` a `granter` can allow a `grantee` to do transfers on their behalf.
 A `transfer_limit` is required to be set for the `grantee`.
 The `allow_list` is optional.
-An empty list means any destination address is allowed, otherwise, the destination must be in the `allow_list`.
\ No newline at end of file
+An empty list means any destination address is allowed, otherwise, the destination must be in the `allow_list`.
+
+### MultiAuthorization :
+
+The marker module supports composite authorizations using the `MultiAuthorization` type.This allows a `granter` to define multiple rules (sub-authorizations) that must all be satisfied for a `grantee` to execute a specific message.
```
SpicyLemon: There should at least be a space after the period, `"type.This"`, but maybe a little more info is handy.

```suggestion
A `MultiAuthorization` contains multiple sub-authorizations that must each be satisfied for the authorization to be valid.

For example, consider a `MultiAuthorization` with a `CountAuthorization` with `2` uses, and a `SendAuthorization` with `500nhash`.

* If the grantee uses it once to send `500nhash`, then the `SendAuthorization` will be depleted and the whole `MultiAuthorization` will become unusable (be deleted) even though there's `1` more use left on the `CountAuthorization`.
* If the grantee uses it twice to send a total of `400nhash`, then the `CountAuthorization` will be depleted and the whole `MultiAuthorization` will become unusable (be deleted) even though there's still `100nhash` left in the `SendAuthorization`.
```

x/marker/types/authz.go
```diff
@@ -99,3 +109,155 @@ func toBech32Addresses(allowed []sdk.AccAddress) []string {
 	}
 	return allowedAddrs
 }
+
+// NewMultiAuthorization creates a new MultiAuthorization
+func NewMultiAuthorization(msgTypeURL string, subAuthorizations ...authz.Authorization) (*MultiAuthorization, error) {
```
SpicyLemon, line 116: It needs to be possible to validate a `MultiAuthorization` outside of construction. I.e. the constructor should only put things together, it should NOT also be doing validation.

There needs to be a `.Validate()` method on the `MultiAuthorization` struct. Most of the time, these structs will be made using a codec (and not this constructor). We still need to run validation on them though, which means we need a `.Validate()` method (or some other validation function). By putting all of these checks in the constructor, you now have two places where you're applying the validation logic. That's bad because now there are two places that need to be updated and kept in sync if any of the validation logic needs to change.

Please make this constructor only return a `MultiAuthorization`, and make the validation a separate thing.

Edit: After reading to the end, I see that we still need this to be able to return an error due to the `Any` packing. That parts needed here, but all of the other validation stuff should still be left to a specific validation method.

x/marker/types/authz.go
```diff
@@ -99,3 +109,155 @@ func toBech32Addresses(allowed []sdk.AccAddress) []string {
...
+				return authz.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrapf(
+					"sub-authorization %d updated to wrong message type: expected %s, got %s",
+					i, m.MsgTypeUrl, resp.Updated.MsgTypeURL(),
+				)
+			}
+			if _, isMulti := resp.Updated.(*MultiAuthorization); isMulti {
+				return authz.AcceptResponse{}, sdkerrors.ErrInvalidRequest.Wrapf(
+					"nested MultiAuthorization not allowed in update for sub-authorization %d", i,
+				)
+			}
```
SpicyLemon: I think these checks can be removed.

It's safe to assume the msg type url won't change during an `Accept` because a lot of other things would break if that were possible.

Similarly, it's safe to assume that the `resp.Updated` authorization is the same type as the one that `Accept` was called on. Since we make sure there aren't any nested `MultiAuthorization`s before writing the authorization to state, there's no way it'd end up with one in here.

x/marker/types/authz.go
```diff
@@ -99,3 +109,155 @@ func toBech32Addresses(allowed []sdk.AccAddress) []string {
...
+		return authz.AcceptResponse{Accept: true, Delete: true}, nil
+	}
+	if anyUpdates {
+		updated := &MultiAuthorization{
+			MsgTypeUrl:        m.MsgTypeUrl,
+			SubAuthorizations: updatedAuths,
+		}
+		return authz.AcceptResponse{Accept: true, Updated: updated}, nil
+	}
+	return authz.AcceptResponse{Accept: true}, nil
```
SpicyLemon: This could be cleaned up a bit.

```suggestion

	rv := authz.AcceptResponse{Accept: true, Delete: anyDeleteRequested}
	if anyUpdates {
		rv.Updated = &MultiAuthorization{MsgTypeUrl: m.MsgTypeUrl, SubAuthorizations: updatedAuths}
	}
	return rv, nil
```

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+
+func GetGrantMultiAuthzCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "grant-multi-authz <grantee> <msg-type-url> <authorizations-json-or-file>",
+		Short: "Grant a MultiAuthorization with multiple sub-authorizations",
+		Long: strings.TrimSpace(`Grant a MultiAuthorization to a grantee with multiple sub-authorizations.
+    	The authorizations can be provided in one of the following ways:
+		- Inline JSON: '[{"@type": "...", ...}]'
+		- File: '@/path/to/authorizations.json'
+		- Stdin: '-'`),
```
SpicyLemon: When using ticks (\') around a string, it includes all newlines and whitespace as is. Specifically, leading whitespace is part of the string. So as you have it here, the "Grant a MultiAuthorization ..." line will start all the way on the left of the screen, then the "The authorizations ..." line will be indented by what looks like four spaces and a tab character. Then the next lines will be indented with two tabs characters. It's also not good to mix tab and space characters like this since there's no way to know how many spaces they use for tabs, so it's not going to keep things in line. And if its set at 8 spaces per tab (not too uncommon), leading with two tabs is too much. It's usually just better to only use spaces.

E.g. It'll look something like this in your terminal

```
Grant a MultiAuthorization to a grantee with multiple sub-authorizations.
            The authorizations can be provided in one of the following ways:
                - Inline JSON: '[{"@type": "...", ...}]'
                - File: '@/path/to/authorizations.json'
                - Stdin: '-'
```

Also, I know a lot of our command definitions don't do this, but since you've got the whole thing wrapped in `strings.TrimSpace`, you could even put the "Grant a ..." line on its own, all the way to the left so that it looks in the file like it'll look in the terminal.

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+					return fmt.Errorf("missing file path after '@'")
+				}
+
+				cleanPath := filepath.Clean(filePath)
+				if cleanPath != filePath {
+					return fmt.Errorf("invalid file path: contains illegal characters")
+				}
+				if strings.Contains(cleanPath, "..") {
+					return fmt.Errorf("invalid file path: contains '..'")
+				}
```
SpicyLemon: You don't need this. Just open and use the path that the user provided.

The `filepath.Clean` doesn't remove illegal characters, it "returns the shortest path name equivalent to path by purely lexical processing. " The reason to use it is if you're putting together the path yourself, and are going to communicate it back to the user and want it cleaned up. Getting back something different doesn't mean there was an illegal character in the filename.

And it's completely legitimate to open a file with `..` in the path. E.g. `grep thing ../myfile.txt`.

If there's an illegal character, or the file can't be found or opened, you'll get that error when reading the file.

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+			} else {
+				parsedTime, err := time.Parse(time.RFC3339, expStr)
+				if err != nil {
+					return fmt.Errorf("invalid expiration format (use RFC3339): %w", err)
+				}
+				if time.Until(parsedTime) <= 0 {
+					return fmt.Errorf("expiration time must be in the future")
+				}
+				expiration = &parsedTime
+			}
```
SpicyLemon: All of the other authz commands use epoch for the expiration date/time, so we should do the same here.

The command for the `MarkerTransferAuthorization` has a default expiration of 1 year, which I think makes sense for that one (since it's specialized), but for most grants (the `authz tx grant` command), the default expiration is 0 (never). I think this should also have a default of 0 since this one is so generic.

Also, you should always set the default by supplying it to the flag setup function. There's a lot of magic that happens with the flags, config files, and environment variables. Defining the default with the flag makes that stuff work better and also gives the users an example (in some cases).

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+	return cmd
+}
+
+func GetRevokeMultiAuthzCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "revoke-multi-authz <grantee> <msg-type-url>",
+		Short: "Revoke a MultiAuthorization grant",
+		Long:  strings.TrimSpace(`Revoke a MultiAuthorization previously granted to a grantee.`),
+		Example: fmt.Sprintf(`%[1]s tx marker revoke-multi-authz cosmos1grantee /provenance.marker.v1.MsgTransfer \
+--from=pbmos1granter`, version.AppName),
```
SpicyLemon: We don't need this command. The `tx authz revoke` command can already delete any authorization.

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+		return nil, fmt.Errorf("failed to unmarshal authorizations array: %w", err)
+	}
+
+	if len(rawAuths) == 0 {
+		return nil, fmt.Errorf("at least one authorization is required")
+	}
+	if len(rawAuths) > types.MaxSubAuthorizations {
+		return nil, fmt.Errorf("too many authorizations (max %d), got %d",
+			types.MaxSubAuthorizations, len(rawAuths))
+	}
```
SpicyLemon: No need to do this here since it's already done in `ValidateBasic()` which is automatically called in `GenerateOrBroadcastTxCLI`. Having it here is just another place that needs updating if we ever need to change this logic.

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+		}
+
+		var auth authz.Authorization
+		if err := clientCtx.InterfaceRegistry.UnpackAny(&anyAuth, &auth); err != nil {
+			return nil, fmt.Errorf("failed to unpack authorization %d: %w", i, err)
+		}
+
+		if _, isMulti := auth.(*types.MultiAuthorization); isMulti {
+			return nil, fmt.Errorf("nested MultiAuthorization not allowed in authorization %d", i)
+		}
```
SpicyLemon: Here too, don't worry about this here, it'll get checked later.

x/marker/client/cli/tx.go
```diff
@@ -1676,3 +1687,252 @@ func GetUpdateMarkerParamsCmd() *cobra.Command {
...
+		if auth.MsgTypeURL() != msgTypeURL {
+			return nil, fmt.Errorf("authorization %d type mismatch: expected %s, got %s",
+				i, msgTypeURL, auth.MsgTypeURL())
+		}
+
+		if validator, ok := auth.(interface{ ValidateBasic() error }); ok {
+			if err := validator.ValidateBasic(); err != nil {
+				return nil, fmt.Errorf("authorization %d failed validation: %w", i, err)
+			}
+		}
```
SpicyLemon: The `Authorization` interface includes `ValidateBasic`. So here, you should be able to just use `auth.ValidateBasic()`.

x/marker/types/authz_test.go
```diff
@@ -149,3 +151,153 @@ func TestMarkerTransferAuthorizationValidateBasic(t *testing.T) {
...
+	resp, err := multiAuth.Accept(ctx, msg)
+	require.NoError(t, err, "multiAuth.Accept")
+	require.True(t, resp.Accept)
+	require.False(t, resp.Delete)
+	require.Nil(t, resp.Updated, "multiAuth.Accept")
+
+	// Test wrong message type
+	wrongMsg := &banktypes.MsgMultiSend{}
+	_, err = multiAuth.Accept(ctx, wrongMsg)
+	require.Error(t, err, "wrongMsg")
```
SpicyLemon: The `Accept` method is complex enough that there should be a lot more tests here. That's the most crucial piece of code in here, so it's important to make sure it's correct.

Off the top of my head:
* Two authzs: wrong first type, wrong second type, wrong both types
* Two authzs: accept + accept, accept + update, accept + delete, update + accept, update + update, update + delete, delete + accept, delete + update, delete + delete.
* Three authzs: all accept, first delete, second delete, third delete, two updates, error from first, error from second, error from third

x/marker/types/authz_test.go
```diff
@@ -149,3 +151,153 @@ func TestMarkerTransferAuthorizationValidateBasic(t *testing.T) {
...
+}
+
+func TestMultiAuthorizationValidateBasic(t *testing.T) {
```
SpicyLemon: You'll probably want a mock authorization to test some of this stuff. There's a couple examples of them in `x/marker/keeper/mocks_test.go` and `x/metadata/keeper/mocks_test.go`. The former is probably a little simpler than you need (probably want to have it record if Accept or ValidateBasic is called), and the latter is probably too complex, but they should give you a good starting point. You won't be able to use those types in here (different package) though.

x/marker/types/authz_test.go
```diff
@@ -149,3 +156,613 @@ func TestMarkerTransferAuthorizationValidateBasic(t *testing.T) {
...
+			multiAuthz: newMultiAuthz(msgSendTypeURL,
+				newAny(newMockAuthorization()),
+				newAny(newMockAuthorization().ToAccept()),
+				newAny(newMockAuthorization().ToAccept()),
+			),
+			msg:     msgSend,
+			expResp: authz.AcceptResponse{Accept: false},
+		},
+		{
+			name: "three: not accept first",
```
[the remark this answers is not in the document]
SpicyLemon: Good catch.

SpicyLemon (approved):

Issue #2347, opened 2025-05-20
Create gov endpoint for unlocking a vesting account.


SpicyLemon:
## Summary

Create Tx endpoint for use in gov props that will convert one or more specified vesting accounts back to base accounts.

## Problem Definition

With v1.23.0, we converted several base accounts into vesting accounts in order to lock up their nhash. It'd be great if we had a way to convert an account back without needed a software upgrade, though.

## Proposal

In the hold module, add a msg server and give it a tx endpoint for unlocking vesting accounts. It will will have an `authority` field and a list of addresses to unlock. This will be a governance-proposal-only endpoint which is done by, in the endpoint method, ensuring that the msg's authority equals the governance module account. See the attribute module's `UpdateParams` endpoint for an example of a gov-prop-only msg.

To unlock the account, simply look it up (from the auth/accounts module), pull the `BaseAccount` out of it, and save it again as just the base account. See the `unlockVestingAccounts` function in `app/upgrades.go` for an example of doing this.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
We might not end up needing this.
I feel like, if none of the accounts need to be unlocked after a year, we can skip this issue (and close it).

SpicyLemon:
Since the hold module doesn't have a `tx.proto` yet, you'll  have to add that. You'll also need to add an entry to `client/docs/config.json` so that its included in the swagger stuff.

PR #2385, opened 2025-06-30
Added governance-only tx endpoint to the hold module for unlocking vesting accounts back to base accounts.

nagarajdivine:
## Description

* Added governance-only tx endpoint to the hold module for unlocking vesting accounts back to base accounts 
closes: #2347 

---

write a little note why.


client/docs/config.json
```diff
@@ -254,6 +254,19 @@
...
+      "tags": {
+        "add": [
+          "Hold"
+        ]
+      },
+      "operationIds": {
+        "rename": {
+          "UnlockVestingAccounts": "HoldUnlockVestingAccounts"
+        }
+      }
```
SpicyLemon: This isn't needed here.
We only need to rename an endpoint if there's another module with an endpoint of the same name, e.g. `Params`.

Also, this whole section for `hold` `tx` should be moved up to be with the hold query section.

proto/provenance/hold/v1/events.proto
```diff
@@ -24,4 +24,15 @@ message EventHoldReleased {
...
+
+// EventUnlockVestingAccounts is an event emitted when vesting accounts are
+// processed for unlock.
+message EventUnlockVestingAccounts {
+  // authority is the address that initiated the unlock (e.g., governance module account).
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // unlocked_count is the number of vesting accounts successfully unlocked.
+  uint32 unlocked_count = 2;
+  // failed_count is the number of vesting accounts that failed to unlock.
+  uint32 failed_count = 3;
```
SpicyLemon: I don't think we need this event, but we could use a similar one. Events are used by external stuff to watch for changes. This event doesn't give enough info to be actionable, though.

What we should do is emit an `EventVestingAccountUnlocked` that has a single field for `account`, and emit one of those for each account unlocked. If unlocking fails, don't emit the event for that account.

proto/provenance/hold/v1/tx.proto
```diff
@@ -0,0 +1,48 @@
...
+  repeated string addresses = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+}
+
+// MsgUnlockVestingAccountsResponse defines the response for unlocking vesting accounts
+message MsgUnlockVestingAccountsResponse {
+  // unlocked_addresses contains the addresses that were successfully unlocked
+  repeated string unlocked_addresses = 1;
+  // failed_addresses contains addresses that failed to unlock with reasons
+  repeated UnlockFailure failed_addresses = 2;
+}
```
SpicyLemon: This response message should be empty.

These endpoints aren't like webservices. The responses are almost always empty because the caller almost never receives it. Submitting Txs is done asynchronously and the response that is actually sent doesn't contain this response object. Further, since this a governance proposal endpoint, this response isn't recorded anywhere and is never returned; it's content is completely lost.

x/hold/client/cli/tx.go
```diff
@@ -0,0 +1,148 @@
...
+		Args:  cobra.NoArgs,
+		Short: "Submit a governance proposal to unlock vesting accounts",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Submit a governance proposal to convert vesting accounts back to base accounts.
+
+Supports both command-line addresses and JSON files for large address sets.
+
+Examples:
+Command-line addresses:
+$ %[1]s tx gov submit-proposal unlock-vesting-accounts-proposal \
```
SpicyLemon: These examples aren't right. They should ultimately be `provenanced tx hold unlock-vesting-accounts-proposal`

x/hold/client/cli/tx.go
```diff
@@ -0,0 +1,148 @@
...
+	txCmd.AddCommand(
+		GetCmdUnlockVestingAccountsProposal(),
+	)
+	return txCmd
+}
+
+// GetCmdUnlockVestingAccountsProposal creates a governance proposal to unlock vesting accounts
+func GetCmdUnlockVestingAccountsProposal() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unlock-vesting-accounts-proposal",
```
SpicyLemon: Let's just shorten this to `unlock-vesting-accounts` and have an alias for `unlock-vesting`.

x/hold/client/cli/tx.go
```diff
@@ -0,0 +1,148 @@
...
+			return provcli.GenerateOrBroadcastTxCLIAsGovProp(clientCtx, flagSet, unlockMsg)
+		},
+	}
+
+	govcli.AddGovPropFlagsToCmd(cmd)
+	provcli.AddAuthorityFlagToCmd(cmd)
+	flags.AddTxFlagsToCmd(cmd)
+
+	cmd.Flags().StringSlice(FlagAddresses, []string{}, "Comma-separated list of addresses to unlock")
+	cmd.Flags().String(FlagAddressesFile, "", "Path to a JSON file containing an array of addresses to unlock")
```
SpicyLemon: Since we don't have anything that would generate such a JSON file, let's just make it a text file with one account per line.

x/hold/client/cli/tx.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	addressList, err := cmd.Flags().GetStringSlice(FlagAddresses)
+	if err != nil {
+		return nil, sdkErrors.ErrInvalidAddress.Wrapf("get flag %s: %v", FlagAddresses, err)
+	}
+
+	if filePath != "" && len(addressList) > 0 {
+		return nil, sdkErrors.ErrInvalidAddress.Wrapf("only one of --addresses or --addresses-file can be specified")
+	}
```
SpicyLemon: Let's just combine both lists if both flags are provided.

Use case: User has a file they created a while ago and know of one or two more to include. They can provide both the file as a flag, and the couple extras can be added with the --addresses flag.

x/hold/codec.go
```diff
@@ -0,0 +1,13 @@
...
+	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
+	registry.RegisterImplementations(
+		(*sdk.Msg)(nil),
+		&MsgUnlockVestingAccountsRequest{},
+	)
+}
```
SpicyLemon: Even though there's only one msg type, we should stick with the pattern used in other modules here.
I.e. in `msgs.go` create a `var AllRequestMsgs = []sdk.Msg{...}` with an entry for each message type. Then in here, we convert that to a `[]proto.Message` to give it to `RegisterImplementations`.

Then, in `msgs_test.go` there should be a `TestAllMsgsGetSigners` an example of which can be seen in the exchange module.

x/hold/keeper/msg_server.go
```diff
@@ -0,0 +1,125 @@
...
+// validateAuthority ensures only governance module can call this function
+func (s msgServer) validateAuthority(authority string) error {
+	govModuleAddr := s.GetAuthority()
+	if authority != govModuleAddr {
+		return sdkerrors.ErrUnauthorized.Wrapf("invalid authority; expected %s, got %s",
+			govModuleAddr,
+			authority)
+	}
+	return nil
+}
```
SpicyLemon: This `validateAuthority` and `unlockVestingAccount` should be keeper methods instead of `msgServer` methods.

Plus, we'll want to be able to use the `unlockVestingAccount` method during an upgrade, so we'll want it to be public.

x/hold/keeper/msg_server.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, sdkerrors.ErrInvalidAddress.Wrapf("account %s is type %T, not a supported vesting account type", addr.String(), account)
+	}
+	if baseVestAcct == nil {
+		return nil, sdkerrors.ErrInvalidType.Wrapf("failed to extract BaseVestingAccount from vesting account: %s", addr.String())
+	}
+
+	if baseVestAcct.BaseAccount == nil {
+		return nil, sdkerrors.ErrInvalidType.Wrapf("BaseVestingAccount.BaseAccount is nil for account: %s", addr.String())
+	}
+	ctx.Logger().Info("Unlocking vesting account",
```
SpicyLemon: This is where we should emit the event about the vesting account being unlocked.

I like the log statement here. I probably wouldn't have thought to put one here, but it's the right thing to do. Nice.

x/hold/keeper/msg_server.go
```diff
@@ -0,0 +1,125 @@
...
+	var baseVestAcct *vesting.BaseVestingAccount
+	switch acct := account.(type) {
+	case *vesting.ContinuousVestingAccount:
+		baseVestAcct = acct.BaseVestingAccount
+	case *vesting.DelayedVestingAccount:
+		baseVestAcct = acct.BaseVestingAccount
+	case *vesting.PeriodicVestingAccount:
+		baseVestAcct = acct.BaseVestingAccount
+	default:
+		return nil, sdkerrors.ErrInvalidAddress.Wrapf("account %s is type %T, not a supported vesting account type", addr.String(), account)
```
SpicyLemon: Let's align these errors a bit.

1. This one should be an `ErrInvalidType`.
2. Let's give them the format `could not unlock account %s: <reason>` where `<reason>` is one of `unsupported account type %T`, `failed to extract base vesting account`, or `failed to extract base account`.

x/hold/keeper/msg_server_test.go
```diff
@@ -0,0 +1,347 @@
...
+		acc := s.app.AccountKeeper.GetAccount(s.ctx, addr)
+		s.T().Logf("Account %d: %s, type: %T", i, addr.String(), acc)
+	}
+}
+
+func TestMsgServerTestSuite(t *testing.T) {
+	suite.Run(t, new(MsgServerTestSuite))
+}
+
+func (s *MsgServerTestSuite) TestMsgUnlockVestingAccounts_SuccessCases() {
```
SpicyLemon: In Go (and our other tests), the pattern is to have a single `TestUnlockVestingAccounts` function/method to test all aspects of the `UnlockVestingAccounts` method. And all the different successes, failures, test cases are in a single `[]struct{...}` that defines the params of the test case such as expected errors and results.

Also, these should be named after the thing actually being tested. In these cases, we're testing the endpoint, not the `Msg`, so these tests should be named with `TestUnlockVestingAccounts`.

x/hold/keeper/msg_server_test.go
```diff
@@ -0,0 +1,347 @@
...
+			case "multiple vesting accounts":
+				addresses = []string{
+					s.vestingAddrs[0].String(),
+					s.vestingAddrs[1].String(),
+					s.vestingAddrs[2].String(),
+				}
+				expUnlocked = 3
+			default:
+				s.T().Fatalf("unsupported test case: %s", name)
+			}
```
SpicyLemon: This isn't the way to do this.

You should define the test cases with these parameters directly instead of defining them in a switch like this. The primary reason is that there are now two separate places where you have to add/remove/change something when making changes to the tests.

```golang
testCases := []struct{
    name string
    addresses []string
    expUnlocked int
}{
    {
        name: "single continuous vesting account",
        addresses: []string{s.vestingAddrs[0].String()},
        expUnlocked: 1,
    },
    ...
}
```

You'll probably end up with more fields when you start adding the failure test cases to the same list.

x/hold/keeper/msg_server_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+		msg2 := hold.MsgUnlockVestingAccountsRequest{
+			Authority: s.govAddr,
+			Addresses: []string{s.vestingAddrs[0].String()},
+		}
+		res, err := s.msgServer.UnlockVestingAccounts(ctx, &msg2)
+		s.Require().NoError(err, "s.msgServer.UnlockVestingAccounts")
+		s.Require().NotNil(res)
+		s.Assert().Len(res.UnlockedAddresses, 0)
+		s.Assert().Len(res.FailedAddresses, 1)
```
SpicyLemon: Aways, always, ALWAYS include a message with a `NotNil` assertion.
It's best practice to always include a message with all assertions, but a `NotNil` assertion is one where its even more important.

The reason you do this is to make it easier to troubleshoot when something goes wrong. These messages provide crucial context to what exactly is going wrong. The standard failure messages can only include so much information, but it doesn't know, e.g. what a variable's name is.

In this case, an appropriate message on the `NotNil` assertion is "UnlockVestingAccounts result".
Since there are two `Len` assertions, there needs to be a message on each one identifying what slice is being tested, i.e. `UnlockedAddresses` and `FailedAddresses`.

---

Details:

When a `NotNil` assertion fails, the only thing they can put in the failure message is "Expected value not to be nil." The only other context it can add is the test name and line number of the failed assertion.

It'd look something like this:
```
    msg_server_test.go:268: 
        	Error Trace:	/Users/dannywedul/git/provenance/x/hold/keeper/msg_server_test.go:268
        	Error:      	Expected value not to be nil.
        	Test:       	TestMsgUnlockVestingAccounts_EdgeCases
```

And you'll often be seeing this outside your IDE, e.g. in a terminal or github action log. There is nothing in there to indicate what is actually wrong though. So all you know is, "Something is broken."

If you were to include a message of "UnlockVestingAccounts result", the failure would look like this:
```
    msg_server_test.go:268: 
        	Error Trace:	/Users/dannywedul/git/provenance/x/hold/keeper/msg_server_test.go:268
        	Error:      	Expected value not to be nil.
        	Test:       	TestMsgUnlockVestingAccounts_EdgeCases
        	Messages:   	UnlockVestingAccounts result
```

Now you know that something went wrong with `UnlockVestingAccounts`, so you already know the area you'll be looking at as you switch back to your IDE.

x/hold/keeper/msg_server_test.go
```diff
@@ -0,0 +1,347 @@
...
+		s.Require().NoError(err, "s.msgServer.UnlockVestingAccounts")
+		s.Require().NotNil(res)
+		s.Require().Len(res.UnlockedAddresses, 1)
+		s.Require().Len(res.FailedAddresses, 0)
+
+		acc := s.app.AccountKeeper.GetAccount(ctx, addr)
+		baseAcc, ok := acc.(*authtypes.BaseAccount)
+		s.Require().True(ok, "account should be base account")
+		s.Assert().Equal(uint64(100), baseAcc.GetAccountNumber())
+		s.Assert().Equal(uint64(42), baseAcc.GetSequence())
```
SpicyLemon: When a `Equal` assertion fails comparing two unsigned integers, it includes the values in hex instead of decimal, which usually makes it harder to troubleshoot. There's two better ways to do it depending on the size and expected values.

The easiest fix is to cast the unsigned integers to signed ones for the comparison. However, with `uint64`, if any of the values are expected to be more than an `int64` can hold, it's better to convert them to strings for the comparison.

In this case, the values are small enough that you can cast the actual values to `int` and use untyped ints for the expected values.

x/hold/keeper/msg_server_test.go
```diff
@@ -0,0 +1,347 @@
...
+	})
+}
+
+func (s *MsgServerTestSuite) TestMsgUnlockVestingAccounts_Performance() {
+	s.Run("batch processing", func() {
```
SpicyLemon: When the everything in a test function is a single test, you don't need to use `s.Run`.

The `s.Run` method is primarily for running sub-tests. You don't need to call it in cases like this where the entire test func is a single test.

x/hold/keys.go
```diff
@@ -6,4 +6,10 @@ const (
 
 	// StoreKey is the store key string for the hold module.
 	StoreKey = ModuleName
+
+	// RouterKey to be used for routing msgs
+	RouterKey = ModuleName
+
+	// QuerierRoute is the querier route for authz
+	QuerierRoute = ModuleName
```
SpicyLemon: You don't need these.

The were once needed by the SDK, but not anymore.

x/hold/expected_keepers.go
```diff
@@ -11,3 +11,9 @@ type BankKeeper interface {
 	AppendLockedCoinsGetter(getter banktypes.GetLockedCoinsFn)
 	SpendableCoins(ctx context.Context, addr sdk.AccAddress) sdk.Coins
 }
+type AccountKeeper interface {
+	GetAccount(ctx context.Context, addr sdk.AccAddress) sdk.AccountI
+	SetAccount(ctx context.Context, acc sdk.AccountI)
+	HasAccount(ctx context.Context, addr sdk.AccAddress) bool
+	NewAccount(ctx context.Context, acc sdk.AccountI) sdk.AccountI
```
SpicyLemon: You don't need `HasAccount` or `NewAccount` here.

Issue #2399, opened 2025-07-16
Move the msg_execute_contract.proto file into another directory


SpicyLemon:
## Summary

Move the `proto/cosmwasm/wasm/v1beta1/msg_execute_contract.proto` file out of the `proto/` directory, into a new one meant to house legacy proto files.

## Problem Definition

The cosmwasm library no longer generates go code for the `cosmwasm.wasm.v1beta1.MsgExecuteContract` message. But we need that in order to read old `Tx`s from. So we added a copy of it so that we can generate the go code ourselves to fix that problem.

But now that message is defined in two places:
* `proto/cosmwasm/wasm/v1beta1/msg_execute_contract.proto`
* `third_party/proto/cosmwasm/wasm/v1beta1/tx.proto`

And since we publish both of those directories as our protos, we now have the same message being defined in two places which is causing some protobuf generation things to fail for others.

We can't get rid of the `third_party` one since that's where it should live. But all we need from the first one is the go code.

## Proposal

1. Move the `proto/` version into a new root directory named "legacy_protos" (maybe there's a better name?).
2. Update the proto -> go code generation to include stuff in that directory.

The proto-breaking checks will probably fail since we're removing a message, but that's okay here.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2402, opened 2025-07-17
fix(cosmwasm-legacy-types): move MsgExecuteContract proto to legacy_protos directory

nagarajdivine:
## Description

* move MsgExecuteContract proto to legacy_proto directory
closes: #2399 

---

write a little note why.

## Summary by CodeRabbit

* **Chores**
  * Updated legacy protobuf definitions and related import paths for improved organization.
  * Added a new Makefile target for generating legacy protobuf files.
  * Expanded Buf configuration to include legacy protos.
  * Updated documentation to remove references to deprecated message types.
  * Introduced a new configuration file for legacy protobuf code generation.


scripts/proto-doc-gen.sh
SpicyLemon, line 1: We don't need to update this script.

scripts/protoball.sh
SpicyLemon, line 1: We definitely don't want to do this. This will put the message back in with all the other protos (that already have this message defined).

scripts/protoc-swagger-gen.sh
SpicyLemon, line 1: We don't need this in swagger stuff either.

buf.work.yaml
```diff
@@ -1,4 +1,5 @@
 version: v1
 directories:
   - proto
+  - legacy_protos
```
SpicyLemon: I'm not quite sure if we need this.
Try removing it and running your new version of `scripts/protocgen.sh`. If it works, let's leave it out.

I honestly can't quite remember what this `buf.work.yaml` file gets used for.

SpicyLemon (commented):
The **only** thing we need/want from this proto is the go code generation. We don't want the proto (or message) actually included with anything, and we don't want the proto to affect anything else.

It's technically a duplicate definition for an existing proto message. Including it with the rest of the protos can cause problems due to this duplication. All documentation about this proto should already be included where appropriate due to the original copy of the message in the third_party/proto dir.

buf.work.yaml
nagarajdivine: buf.work.yaml is workspace configuration it support multiple proto dir, we have proto,thirdparty and new legacy_proto. we need to add it, it called inside the `protocgen.sh scripts`

Makefile
```diff
@@ -539,7 +539,14 @@ proto-update-deps:
 	@echo "Updating Protobuf files"
 	sh ./scripts/proto-update-deps.sh
 
-.PHONY: proto-all proto-checks proto-regen proto-gen proto-format proto-lint proto-check-breaking proto-check-breaking-third-party proto-update-deps proto-update-check proto-doc-gen
+proto-legacy:
+	docker run --rm \
+		-v $(PWD):/workspace \
+		--workdir /workspace \
+		ghcr.io/cosmos/proto-builder:0.15.1 \
+		buf generate --template proto/buf.gen.legacy.yaml legacy_protos
```
SpicyLemon: **issue (blocking)**: Please move `proto/buf.gen.legacy.yaml` to `legacy_protos/buf.gen.yaml` and update this make target to reflect that. That way we're keeping the legacy stuff together and there's less mention of the legacy stuff in the main proto stuff.

Also, please add this to the start of the make targe:
```
	@echo "Generating Legacy Protobuf Files"
```
(the first char there needs to be tab, as usual).

scripts/protoball.sh
```diff
@@ -8,6 +8,7 @@ fi
 
 PROTO_DIR=proto
 
+
```
SpicyLemon: **nitpick**: Remnants of changes undone.

SpicyLemon (approved):

PR #2411, opened 2025-07-29
Migrate name module from KV store to collections.

### LINKED ISSUE #2018: Switch to Collections in Name module.
## Summary

The collections package has been added to cosmos sdk v0.50, and we should see if we can leverage this. This package will help improve clarity and may improve performance of our storage usage. A simple module that we can easily integrate this into and use for gathering metrics is the `name` module.

## Problem Definition

The current way that we use state is quite cumbersome and it may not be optimal.

## Proposal

The following steps are one of many ways to implement this and can be followed loosely to incorporate this into the `name` module:
- Create a codec for Record message
- Update the prefixes to use collections.Prefix.
- Store the params as an Item
- Store the name as a Map<String, Record>
- Store the addresses with as a Map<Pair<Addr, String>, Record>
- Update pagination queries to use collections.Paginate

Some useful metrics to gather, compare, and discuss are:
- Gas usage when adding, removing, and editing entries
- Runtime performance under different loads
____

#### For Admin Use

- [ ] Not duplicate issue

nagarajdivine:
## Description

* migration of name module from kv store to collections
closes: #2018 

---

write a little note why.

## Summary by CodeRabbit

* **New Features**
  * Name records now use a more reliable storage system, with automated migration from existing data.
  * Name keys use hashed names, and reverse lookups provide stable, paginated results.

* **Bug Fixes**
  * Improved validation, error reporting, index consistency, and lookup ordering.
  * Parameter updates now report persistence errors correctly.

* **Tests**
  * Added coverage for migration, storage, indexing, key encoding, and reverse-lookup behavior.

* **Documentation**
  * Added changelog entries for the name storage and key updates.


app/upgrades.go
```diff
@@ -82,6 +83,24 @@ var upgrades = map[string]appUpgrade{
...
 	},
+	"collectionsmigration-rc1": { // Migration name module from kv store to collections (v2 -> v3)
+		Handler: func(ctx sdk.Context, app *App, vm module.VersionMap) (module.VersionMap, error) {
+			migrator := namekeeper.NewMigrator(app.NameKeeper)
+			if err := migrator.MigrateKVToCollections2to3(ctx); err != nil {
+				return nil, err
+			}
+			return app.mm.RunMigrations(ctx, app.configurator, vm)
+		},
+	},
```
SpicyLemon: **issue (blocking)**: You don't need any of the changes in this file.

Since you've set it up as a module migration, calling `app.mm.RunMigrations` will find and execute the func registered in the module's `RegisterServices` method. So as you've got it, it'll run the migration, then it'll run all module migrations, which will include re-applying the upgrade.

Additionally, we don't need a whole new upgrade for this, it'll just be part of the `bouvardia` upgrade. That upgrade already calls `runModuleMigrations` (which in turn calls `app.mm.RunMigrations`). So it's already set up to run on the next upgrade.

That being said, it'd still be a good idea to add an expected log line to the `bouvardia` upgrade test; just the line you're having your migration issue at the very start.

app/upgrades_test.go
```diff
@@ -1063,6 +1063,17 @@ func (s *UpgradeTestSuite) TestBouvardia() {
 	}
 	s.AssertUpgradeHandlerLogs("bouvardia", expInLog, nil)
 }
+func (s *UpgradeTestSuite) TestCollectionsmigration() {
+	expInLog := []string{
+		"INF Migrating name module from KV store to collections (v2 to v3)...",
```
SpicyLemon: **suggestion**: Move this line up into both the `TestBouvardiaRC1` and `TestBouvardia` expected lines, right after the "Starting module migrations" entries.

x/attribute/keeper/keeper_test.go
```diff
@@ -853,7 +853,7 @@ func (s *KeeperTestSuite) TestGetAttributesByName() {
 	s.Assert().NoError(s.app.AttributeKeeper.SetAttribute(s.ctx, attr, s.user1Addr), "should save successfully")
 	_, err := s.app.AttributeKeeper.GetAttributes(s.ctx, s.user1, "blah")
 	s.Assert().Error(err)
-	s.Assert().Equal("no address bound to name", err.Error())
+	s.Assert().Contains("no address bound to name", err.Error())
```
SpicyLemon: **issue (blocking)**: This is not doing what you want it to do. Annoyingly, the assertion functions didn't standardize on either actual then expected, nor expected then actual. So when you change the assertion function, you need to also check to make sure you're calling it correctly. In this case, the arguments are (thing, contains). So you're asserting that the contents of `err.Error()` are in the contents of `"no address bound to name"` which is backwards.

Additionally, it's usually better to update the expected error contents than to switch to just checking if it contains what it use to be expecting.

x/name/keeper/keeper.go
```diff
@@ -20,14 +22,16 @@ import (
 type Keeper struct {
 	// Key to access the key-value store from sdk.Context.
 	storeKey storetypes.StoreKey
```
SpicyLemon: **issue (blocking)**: The `storeKey` should no longer be needed since everything is in collections.

x/name/keeper/keeper.go
```diff
@@ -20,14 +22,16 @@ import (
...
 	// the signing authority for the gov proposals
-	authority string
-
-	attrKeeper types.AttributeKeeper
+	authority   string
+	Schema      collections.Schema
+	attrKeeper  types.AttributeKeeper
+	ParamsStore collections.Item[types.Params]
+	NameRecords collections.Map[[]byte, types.NameRecord] // key: 0x03 + hash(name)
+	AddrIndex   collections.Map[[]byte, types.NameRecord] // key: 0x05 + addr + name key
```
SpicyLemon: **suggestion**: Please swap the `Schema` and `attrKeeper` in the order, then put an empty line back between `authority` and `attrKeeper`, then another empty line between `attrKeeper` and `Schema` (which is now below it).

x/name/keeper/keeper.go
```diff
@@ -38,12 +42,23 @@ type Keeper struct {
 func NewKeeper(
 	cdc codec.BinaryCodec,
 	key storetypes.StoreKey,
```
SpicyLemon: **suggestion**: Since the `storeKey` isn't needed anymore, it doesn't need to be an arg here either.

x/name/keeper/keeper.go
```diff
@@ -38,12 +42,23 @@ type Keeper struct {
...
-		storeKey:  key,
-		cdc:       cdc,
-		authority: authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+	sb := collections.NewSchemaBuilder(storeService)
+	k := Keeper{
+		cdc:         cdc,
+		storeKey:    key,
+		authority:   authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		NameRecords: collections.NewMap(sb, collections.NewPrefix(types.NameKeyPrefix), "names", types.RawBytesKey, codec.CollValue[types.NameRecord](cdc)),
+		AddrIndex:   collections.NewMap(sb, collections.NewPrefix(types.AddressKeyPrefix), "addr_index", types.RawBytesKey, codec.CollValue[types.NameRecord](cdc)),
```
SpicyLemon: **issue**: You should be using an `IndexedMap` for the name records so that the index updating is automatic. It can be a bit of a bear to get set up, though. I usually use `x/bank` module's [ViewKeeper](https://github.com/provenance-io/cosmos-sdk/blob/release-pio/v0.50.x/x/bank/keeper/view.go#L73) for an example.

That one has a map of address + denom -> balance and also has an index for denom + address -> balance

For this, specially since we're doing a migration, the name records map should be namekey -> name record and the index would be addr + namekey -> emptyslice.

The auth keeper also uses an `IndexedMap` if you'd like another example.

SpicyLemon: If we absolutely need two collections here, then they need to be made private since they both always need to be updated in conjunction.

x/name/keeper/keeper.go
```diff
@@ -89,58 +104,105 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+	normalizedName, err := k.Normalize(ctx, name)
+	if err != nil {
 		return err
 	}
 	if err = types.ValidateAddress(addr); err != nil {
 		return types.ErrInvalidAddress.Wrap(err.Error())
 	}
+	nameKey, err := types.GetNameKeyPrefix(normalizedName)
+	if err != nil {
+		return types.ErrNameNotBound.Wrapf("failed to get name key prefix for name %q: %v", normalizedName, err)
```
SpicyLemon: polish: It's not a prefix here, it's the actual key.

```suggestion
		return types.ErrNameNotBound.Wrapf("failed to get key for name %q: %v", normalizedName, err)
```

SpicyLemon (commented):
Some preliminary comments.

x/name/keeper/keeper.go
nagarajdivine: Even though we now use collections, it doesn’t provide direct access to the raw keys and values stored in the old KV format — so we still need `storeKey` to read that data during migration (please correct me if I’m missing something).

x/name/keeper/keeper.go
nagarajdivine: I have made private for both collections :)

app/upgrades_test.go
```diff
@@ -1056,6 +1056,7 @@ func (s *UpgradeTestSuite) TestBouvardiaRC1() {
 func (s *UpgradeTestSuite) TestBouvardia() {
 	expInLog := []string{
 		"INF Starting module migrations. This may take a significant amount of time to complete. Do not restart node.",
+		// "INF Migrating name module from KV store to collections (v2 to v3)...",
```
SpicyLemon: **issue (blocking)**: This shouldn't be commented out.

app/upgrades_test.go
```diff
@@ -1036,7 +1036,7 @@ func (s *UpgradeTestSuite) TestUnlockVestingAccounts() {
 
 // Create strings with the log statements that start off the reusable upgrade functions.
 var (
-	LogMsgRunModuleMigrations                  = "INF Starting module migrations. This may take a significant amount of time to complete. Do not restart node."
+	LogMsgRunModuleMigrations = "INF Starting module migrations. This may take a significant amount of time to complete. Do not restart node."
```
SpicyLemon: **polish (blocking)**: Please put this spacing back.

x/attribute/keeper/query_server_test.go
```diff
@@ -14,12 +14,12 @@ import (
 	"github.com/cosmos/cosmos-sdk/types/query"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
-
```
SpicyLemon, line 17: **polish (blocking)**: This empty line should still be here.

x/name/keeper/genesis.go
```diff
@@ -1,13 +1,17 @@
...
+	storetypes "cosmossdk.io/store/types"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	types "github.com/provenance-io/provenance/x/name/types"
 )
 
 // InitGenesis creates the initial genesis state for the name module.
 func (k Keeper) InitGenesis(ctx sdk.Context, data types.GenesisState) {
+	// Set module parameters
```
SpicyLemon: **polish**: Either remove this comment or put another comment below about setting the name records.

x/name/keeper/genesis.go
```diff
@@ -22,16 +26,22 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data types.GenesisState) {
...
-	// Genesis state data structure.
-	records := types.NameRecords{}
-	// Callback func that adds records to genesis state.
-	appendToRecords := func(record types.NameRecord) error {
+	// Collect all name records
+	records := make(types.NameRecords, 0)
+
+	rng := (&collections.Range[[]byte]{}).
+		StartInclusive(types.NameKeyPrefix).
+		EndExclusive(storetypes.PrefixEndBytes(types.NameKeyPrefix))
```
SpicyLemon: **issue (blocking**: You can provide a `nil` ranger to walk over all of the entries in a collection.

The key prefix is something the collection already knows about, so you're not actually ranging over everything here.

x/name/keeper/keeper.go
```diff
@@ -38,12 +47,23 @@ type Keeper struct {
...
-		storeKey:  key,
-		cdc:       cdc,
-		authority: authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+	sb := collections.NewSchemaBuilder(storeService)
+	k := Keeper{
+		cdc:         cdc,
+		storeKey:    key,
+		authority:   authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		nameRecords: collections.NewMap(sb, collections.NewPrefix(types.NameKeyPrefix), "names", types.RawBytesKey, codec.CollValue[types.NameRecord](cdc)),
+		addrIndex:   collections.NewMap(sb, collections.NewPrefix(types.AddressKeyPrefix), "addr_index", types.RawBytesKey, codec.CollValue[types.NameRecord](cdc)),
```
SpicyLemon: **issue (blocking)**: We should be using an `IndexedMap` here instead of two different collections, and we should be able to do `nameRecords.Get(ctx, name)`.

One of the main points of switching to collections is to not have to mess with the key bytes anymore.

The SDK doesn't have a specific key codec that will work for us though, so we need to make our own for use in here. Something like `HashedStringKeyCodec`. It'll have an underlying type of `string` and be like a mix of the `StringKeyCodec` and `BytesKeyCodec`. The catch is that the decode funcs should return a stringed (base64 encoded) version of the name _hash_ instead of the original name (since we can't convert back like that).

And the `addrIndex` should be an index on that collection. See the bank module's balances collection for how to do this.

x/name/types/keys.go
```diff
@@ -72,3 +75,65 @@ func GetAddressKeyPrefix(addr sdk.AccAddress) (key []byte, err error) {
 func ValidateAddress(address sdk.AccAddress) error {
 	return sdk.VerifyAddressFormat(address)
 }
+
+type rawBytesKeyCodec struct{}
+
+// RawBytesKey is a codec for []byte keys without any length prefixing or transformation.
+var RawBytesKey codec.KeyCodec[[]byte] = rawBytesKeyCodec{}
```
SpicyLemon: **suggestion**: Redo this `rawBytesKeyCodec` as the `hashedStringKeyCodec` I mentioned earlier. The underlying type would be a `string` instead of `[]byte`

The encode-related fields will take in a string, hash it and use the bytes exactly the same way it's being done currently. Don't append the type byte or anything, just hash the name and get the bytes from that (I can't remember if we use all or a portion of the bytes, but do the same thing here).

The decode-related fields (and `Stringify`) will return a base64 encoded version of the bytes.

The non-terminal stuff should be the same as with a `StringKey`, which I think adds a null byte to the end.

SpicyLemon: The idea is that we can then provide the name string directly to the collection `Has`, `Get`, `Set`, etc without having to fuss with putting together a key.

x/name/keeper/keeper.go
SpicyLemon: In the `Keeper`, replace the `storeKey` field with the `storeService` (that's being provided to the constructor). Later, when you need the store for the migration, you can get it from the `storeService`.

x/name/keeper/migrations.go
```diff
@@ -9,3 +19,63 @@ type Migrator struct {
 func NewMigrator(keeper Keeper) Migrator {
 	return Migrator{keeper: keeper}
 }
+
+// MigrateKVToCollections2to3 migrates the name module data from the legacy KV store layout
+// to the new collections-based layout (version 2 to version 3).
+func (m Migrator) MigrateKVToCollections2to3(ctx sdk.Context) error {
+	ctx.Logger().Info("Migrating name module from KV store to collections (v2 to v3)...")
+
+	store := ctx.KVStore(m.keeper.storeKey)
```
SpicyLemon: **polish**: You can get the store from the store service (allowing the removal of the `storeKey` now).

```suggestion
	store := k.storeService.OpenKVStore(ctx)
```

x/name/keeper/migrations.go
```diff
@@ -9,3 +19,63 @@ type Migrator struct {
...
+
+	// Migrate parameters
+	if bz := store.Get(types.NameParamStoreKey); bz != nil {
+		var params types.Params
+		m.keeper.cdc.MustUnmarshal(bz, &params)
+		if err := m.keeper.paramsStore.Set(ctx, params); err != nil {
+			return err
+		}
+		ctx.Logger().Info("Migrated name module parameters to collections store.")
+	}
```
SpicyLemon: **polish**: Since the new collection for the params uses the same `types.NameParamStoreKey` key, you don't need to migrate the params.

Also, you shouldn't be using the `Must` version of stuff in here. You should be using the normal ones that return an error so that you can handle the error without blowing up. A panic during a migration will essentially halt the chain until we can get a fix released and distributed, so it's REALLY bad to panic here.

x/name/keeper/migrations.go
```diff
@@ -9,3 +19,63 @@ type Migrator struct {
...
+
+	// Migrate name records
+	ctx.Logger().Info("Migrating name records...")
+	namePrefix := types.NameKeyPrefix
+	nameIter := store.Iterator(namePrefix, storetypes.PrefixEndBytes(namePrefix))
+	defer nameIter.Close()
+	count := 0
+	for ; nameIter.Valid(); nameIter.Next() {
+		var record types.NameRecord
+		m.keeper.cdc.MustUnmarshal(nameIter.Value(), &record)
```
SpicyLemon: **issue**: If this entry can't be unmarshalled, we should log what we can about it and move on. We should **NOT** panic here.

x/name/keeper/migrations.go
```diff
@@ -9,3 +19,63 @@ type Migrator struct {
...
+
+		// Use the exact key from the store
+		key := nameIter.Key()
+
+		// Set in collections
+		if err := m.keeper.nameRecords.Set(ctx, key, record); err != nil {
```
SpicyLemon: **issue**: You **must not** update stuff as you're iterating over it.

It might feel like you're iterating over one thing and setting another, but underneath, they're both talking to the same store. Further, both the iterator and collection use `types.NameKeyPrefix`, so you're doing writes to the same stuff you're trying to iterate over. That will almost certainly cause problems.

What we should do is to switch to a new key prefix:

1. Copy the existing `types.NameKeyPrefix` value into this `migrations.go` file, maybe name it `OldNameKeyPrefix`.
2. Change the value of `types.NameKeyPrefix`.
3. Create the iterator using the old value.
4. For each value, unmarshal it and use `.Set` on the collection. This is okay here because the new key isn't something in the range of the iterator.
5. Store the key for each old entry in a slice.
6. Also store the key for each old address index entry in a slice.
7. After you're done iterating over all the entries, delete all of the old entries.

This assumes you've combined the name and address index collections.

x/name/keeper/migrations.go
```diff
@@ -9,3 +19,63 @@ type Migrator struct {
 func NewMigrator(keeper Keeper) Migrator {
 	return Migrator{keeper: keeper}
 }
+
+// MigrateKVToCollections2to3 migrates the name module data from the legacy KV store layout
+// to the new collections-based layout (version 2 to version 3).
+func (m Migrator) MigrateKVToCollections2to3(ctx sdk.Context) error {
+	ctx.Logger().Info("Migrating name module from KV store to collections (v2 to v3)...")
```
SpicyLemon: **praise**: Woo! Thank you for proper sentences (capital first letter and ending punctuation) on these log statements.

It's sometimes confusing because errors are supposed to NOT start with a capital letter or have ending punctuation, but that's so that it's easier to place them in log statements that SHOULD have those things.

I'm glad to see it here.

x/name/keeper/keeper.go
SpicyLemon: **issue**: One mistake we made with the existing address name index is that we should have left the value empty instead of duplicating the name record. We should fix that here. The address index should have the same key format as now, but the value should be the full name. This should be possible by using an `IndexedMap`.

There might be some keeper lookup stuff that will need tweaking now, though. Rather than just getting a value from the address index, you'll get the value (which is now the name), then get the name record from the name collection.

SpicyLemon (commented):
I haven't looked at the tests yet, but I'm certain that we can and should be using an `IndexedMap` for both the name records and the address index.

app/app.go
```diff
@@ -998,7 +998,7 @@ func (app *App) setPostHandler() {
 func (app *App) registerUpgradeHandlers() {
 	// Add the upgrade handlers for each release.
 	InstallCustomUpgradeHandlers(app)
-
+	app.injectUpgrade("bouvardia")
```
SpicyLemon: **issue (blocking)**: Remove this.

x/name/types/keys.go
```diff
@@ -29,46 +32,115 @@ var (
...
+
+func (c HashedStringKeyCodec) EncodeNonTerminal(buffer []byte, key string) (int, error) {
+	// Use collections.StringKey for non-terminal encoding
+	strKey := collections.StringKey
+	size := strKey.SizeNonTerminal(key)
+	if len(buffer) < size {
+		return 0, fmt.Errorf("buffer too small")
+	}
+	return strKey.EncodeNonTerminal(buffer, key)
+}
```
SpicyLemon: **issue**: This needs to hash the key too, just like `Encode`. You should use a byte slice key here instead of a StringKey.

When the collections stuff is combining two fields into a single string, the  first one is encoded using `EncodeNonTerminal`, and the second is encoded using `Encode`. If there are three or more fields, the last one uses `Encode`, and the others all use `EncodeNonTerminal`.

They should both return the same thing except `EncodeNonTerminal` might have an extra byte to help distinguish one field from another in the resulting byte slice.

The `DecodeNonTerminal` and `SizeNonTerminal` funcs need to be updated to reflect this too.

This should essentially have the same as underlying functionality as the `bytesKey`, except the key itself is a string that it hashes into the byte slice. All the encoding should be doing that. All the decoding should be returning the base64 encoded version of the bytes.

x/name/keeper/migrations.go
```diff
@@ -9,3 +24,40 @@ type Migrator struct {
...
+	// 1. Migrate parameters
+	if bz, _ := store.Get(types.NameParamStoreKey); bz != nil {
+		var params types.Params
+		m.keeper.cdc.MustUnmarshal(bz, &params)
+		if err := m.keeper.paramsStore.Set(ctx, params); err != nil {
+			return fmt.Errorf("failed to migrate params: %w", err)
+		}
+		ctx.Logger().Info("Migrated parameters")
+		store.Delete(types.NameParamStoreKey)
+	}
```
SpicyLemon: **Issue**: You shouldn't need to migrate the params.

They're currently stored as `<NameParamsStoreKey> => protobuf(params)`. The new `paramsStore` collection also uses `types.NameParamStoreKey` as the key and protobuf encodes the params. It's the exact same entry in the exact same place as before. That also means that the `store.Delete` line here is deleting the params completely (i.e. `m.keeper.paramsStore.Get` won't return anything). So once this is done running, there won't be any name params anymore.

Using the `store` with custom keys and values is like doing low-level reads and writes, while the collections are a framework that manages all those low-level reads and writes for us. But it's the same data either way, regardless of how you're interacting with it.

x/name/keeper/migrations.go
```diff
@@ -9,3 +24,40 @@ type Migrator struct {
...
+	ctx.Logger().Info("Migrating name records...")
+	addrPrefix := types.AddressKeyPrefix
+	iter, _ := store.Iterator(addrPrefix, storetypes.PrefixEndBytes(addrPrefix))
+	defer iter.Close()
+
+	count := 0
+	for ; iter.Valid(); iter.Next() {
+		store.Set(iter.Key(), []byte{})
+		count++
+	}
```
SpicyLemon: **issue (blocking)**: Don't ever write to the same thing you're iterating over. You're also not even migrating stuff; all you're doing is clearing out all the values.

To confuse things even more, the comment says we're migrating the name records here, but it looks like it's only doing stuff with the index.

What we should be doing here is move the old name records to the new place, then delete all the old existing index entries.

app/app.go
```diff
@@ -998,7 +998,6 @@ func (app *App) setPostHandler() {
 func (app *App) registerUpgradeHandlers() {
 	// Add the upgrade handlers for each release.
 	InstallCustomUpgradeHandlers(app)
-
```
SpicyLemon, line 1062: **polish**: Please keep this empty line since there's a comment on `injectUpgrade` directing people to put a line of code here.

x/name/keeper/keeper.go
```diff
@@ -278,130 +281,154 @@ func (k Keeper) Normalize(ctx sdk.Context, name string) (string, error) {
...
 	var toDelete [][]byte
 
-	extractNameKey := func(key []byte) []byte {
-		// byte 1 is the type byte (0x05), it's ignored here.
-		// The 2nd byte is the length of the address that immediately follows it.
-		// The name key starts directly after the address, and is the rest of the key.
-		addrLen := int(key[1])
-		nameKeyStart := addrLen + 2
-		return key[nameKeyStart:]
+	store := k.storeService.OpenKVStore(ctx)
```
SpicyLemon: **polish**: Let's just delete this `DeleteInvalidAddressIndexEntries` method. It's not needed anymore.

Background: Originally, name records couldn't be updated. At some point, we added that functionality, but forgot to update the setter to delete the old index entry when it was changing. So there were some left-over index entries that weren't correct. So this was used as part of a migration to go through all index entries, find the ones that weren't correct, and delete them. We ran that as part of the upgrade that fixed the bug in the setter. So we're good on this front, and don't need this method anymore.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+	normalizedName, err := k.Normalize(ctx, name)
+	if err != nil {
 		return err
 	}
 	if err = types.ValidateAddress(addr); err != nil {
 		return types.ErrInvalidAddress.Wrap(err.Error())
 	}
-
-	if err = k.addRecord(ctx, name, addr, restrict, false); err != nil {
+	if err := k.addRecord(ctx, name, addr, restrict, false); err != nil {
```
SpicyLemon: **issue**: You need to supply `normalizedName` here now. Originally, the `name` variable was being updated, so `name` was used below here. But now that you're storing the normalized version in a new variable, that new variable is what you need to supply here.

**polish**: Go back to just `err = ...`. The `err` variable already exists in the scope, so now you're needlessly creating a shadow version of it here for this if block.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 		return err
 	}
-
-	nameBoundEvent := types.NewEventNameBound(addr.String(), name, restrict)
-
```
SpicyLemon, line 104: **polish**: Please leave an empty line here after the if block and before the event emission/return.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-	if err := k.IterateRecords(ctx, addrPrefix, appendToRecords); err != nil {
-		return records, err
-	}
-	return records, nil
+
+		return false, nil
+	})
+
+	return records, err
 }
```
SpicyLemon, line 244: **issue (blocking)**: It's WAY too much work to look through all name entries each time. This is exactly what the index is for, finding the name records for a specific address without having to look at all the name records.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 	if err != nil {
 		return err
 	}
-	// Delete the main name record
-	key, err := types.GetNameKeyPrefix(name)
+	record, err := k.nameRecords.Get(ctx, normalizedName)
 	if err != nil {
-		return err
+		return types.ErrNameNotBound
 	}
```
SpicyLemon, line 255: **issue (blocking)**: When `collection.Get` doesn't find an entry, it returns a not-found error that you need to also handle. There are other errors that it might return too, though, so you can't just assume every error is a not-found error.

You'll need to update the error handling of all your `.Get` calls.

I suggest updating `GetRecordByName` to correctly handle the results of `.Get`, then always use the `GetRecordByName` method instead of `.Get`.

x/name/types/keys.go
```diff
@@ -29,46 +34,125 @@ var (
 	NameParamStoreKey = []byte{0x06}
```
SpicyLemon: **issue (blocking)**: Since we're changing storage stuff, we should also update the prefixes.

Change the prefixes here to `0x07`, `0x08`, and `0x09`. Move these old values into the migration stuff. The migration process will then be to copy everything from the old place/way to the new place/way, then delete all of the old stuff. You'll be able to loop over the existing stuff, writing each new record at the same time since it will **NOT** be updating the state that you're also iterating over. Store each of the keys, though, so that you don't have to iterate over the state again in order to delete the old entries.

x/name/keeper/keeper.go
```diff
@@ -278,130 +281,154 @@ func (k Keeper) Normalize(ctx sdk.Context, name string) (string, error) {
...
+			logger.Error("failed to delete invalid index entry", "key", fmt.Sprintf("%x", key), "error", err)
+		}
 	}
-
 	logger.Info(fmt.Sprintf("Done checking address -> name index entries. Deleted %d invalid entries and kept %d valid entries.", len(toDelete), keepCount))
 }
 
+func (k *Keeper) GetNameRecord(ctx sdk.Context, key string) (types.NameRecord, error) {
+	return k.nameRecords.Get(ctx, key)
+}
```
SpicyLemon: **issue (blocking)**: Please delete this. There is already a `GetRecordByName` that does this.

x/name/keeper/keeper.go
```diff
@@ -278,130 +281,154 @@ func (k Keeper) Normalize(ctx sdk.Context, name string) (string, error) {
...
+
+func (k Keeper) GetAddrIndex() *indexes.Multi[collections.Pair[[]byte, string], string, types.NameRecord] {
+	return k.nameRecords.Indexes.AddrIndex
+}
+
+func (k Keeper) addRecord(ctx sdk.Context, name string, addr sdk.AccAddress, restrict bool, isModifiable bool) error {
```
SpicyLemon: **suggestion**: Let's get rid of this `addRecord` method, and just put what's needed directly in the set/update methods.

The `.Normalize` method is expensive, and it's best to minimize its use. The way things are, the set/update methods do the normalization, then send it here where it's done again.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-
 	nameUpdateEvent := types.NewEventNameUpdate(addr.String(), name, restrict)
-
 	return ctx.EventManager().EmitTypedEvent(nameUpdateEvent)
 }
 
 // GetRecordByName resolves a record by name.
 func (k Keeper) GetRecordByName(ctx sdk.Context, name string) (record *types.NameRecord, err error) {
-	key, err := types.GetNameKeyPrefix(name)
+	normalizedName, err := k.Normalize(ctx, name)
```
SpicyLemon: **suggestion**: Don't normalize the name here, just look for it with what was provided.

The idea is that all the names are normalized as they're stored in state. From that point on, everyone knows exactly what it **should** be, so we don't have to normalize it anymore.

The thing is that `.Normalize` requires a state lookup, which is expensive, so should be avoided.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-	return getNameRecord(ctx, k, key)
-}
-
-func getNameRecord(ctx sdk.Context, keeper Keeper, key []byte) (record *types.NameRecord, err error) {
-	store := ctx.KVStore(keeper.storeKey)
-	if !store.Has(key) {
+	namerecord, err := k.nameRecords.Get(ctx, normalizedName)
+	if err != nil {
 		return nil, types.ErrNameNotBound
 	}
```
SpicyLemon, line 210: **issue (blocking)**: Incorrect handling of `.Get` result.

It's possible for other errors to be returned from `.Get`; you need to check if it's a not-found error in order to know that that's the case.

x/name/keeper/keeper.go
```diff
@@ -16,34 +20,88 @@ import (
...
+		collections.NewPrefix(types.AddressKeyPrefix),
+		"addr_index",
+		collections.PairKeyCodec(collections.BytesKey, collections.StringKey),
+		types.HashedStringKeyCodec{}, // Primary key codec
+		func(name string, record types.NameRecord) (collections.Pair[[]byte, string], error) {
+			addr, err := sdk.AccAddressFromBech32(record.Address)
+			if err != nil {
+				return collections.Pair[[]byte, string]{}, err
+			}
+			return collections.Join([]byte(addr), name), nil
```
SpicyLemon: **issue**: The `name` here, should be hashed like it is for normal records. I.e. it should be the same value used for the main collection key.

x/name/keeper/keeper.go
```diff
@@ -16,34 +20,88 @@ import (
...
-		cdc:       cdc,
-		authority: authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+	sb := collections.NewSchemaBuilder(storeService)
+
+	// Create address index
+	addrIndex := indexes.NewMulti(
+		sb,
+		collections.NewPrefix(types.AddressKeyPrefix),
+		"addr_index",
+		collections.PairKeyCodec(collections.BytesKey, collections.StringKey),
```
SpicyLemon: **issue**: I think we want this to be a `collections.PairKeyCodec(sdk.AccAddressKey, collections.BytesKey)`

Since the first field is an `AccAddress`, we should use the appropriate codec.
The second field should be the same value that the main collection uses, which ends up being a byte slice. Or maybe you should re-use the `HashedStringKeyCodec` there too.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 // UpdateNameRecord updates the owner address and restricted flag on a name.
 func (k Keeper) UpdateNameRecord(ctx sdk.Context, name string, addr sdk.AccAddress, restrict bool) error {
-	var err error
-	if name, err = k.Normalize(ctx, name); err != nil {
+	normalizedName, err := k.Normalize(ctx, name)
+	if err != nil {
 		return err
 	}
-	if err = types.ValidateAddress(addr); err != nil {
+	if err := types.ValidateAddress(addr); err != nil {
```
SpicyLemon: **polish**: The `err` variable already exists in the scope, so you should just use `=` here (not `:=`).

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-		oldAddrKey, err = types.GetAddressKeyPrefix(oldAddr)
-		if err != nil {
-			return types.ErrInvalidAddress.Wrapf("invalid existing %s record address format: %v", name, err)
-		}
-		oldAddrKey = append(oldAddrKey, oldNameKeyPre...)
-		store := ctx.KVStore(k.storeKey)
-		store.Delete(oldAddrKey)
+	_, err = k.nameRecords.Get(ctx, normalizedName)
+	if err != nil {
+		return types.ErrNameNotBound
```
SpicyLemon: **issue**: If you don't need the record, you should just use `.Has`. That way, you don't need to worry about checking for a not-found error from the `.Get`.

x/name/keeper/keeper.go
```diff
@@ -88,173 +146,118 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-	record = &types.NameRecord{}
-	err = keeper.cdc.Unmarshal(bz, record)
-	return record, err
+	return &namerecord, nil
 }
 
 // NameExists returns true if store contains a record for the given name.
 func (k Keeper) NameExists(ctx sdk.Context, name string) bool {
-	key, err := types.GetNameKeyPrefix(name)
+	normalizedName, err := k.Normalize(ctx, name)
```
SpicyLemon: **issue**: You shouldn't normalize the name here. We only want to normalize it as its being written. After that, everyone knows exactly what it should be, so we only accept exact matches.

x/name/keeper/params.go
```diff
@@ -8,21 +8,18 @@ import (
...
-	params = types.DefaultParams() // Assuming DefaultParams initializes all defaults
-
-	bz := store.Get(types.NameParamStoreKey) // General key for all parameters
-	if bz != nil {
-		k.cdc.MustUnmarshal(bz, &params) // Deserialize parameters from bytes
+	params, err := k.paramsStore.Get(ctx)
+	if err != nil {
+		return types.DefaultParams()
 	}
 	return params
```
SpicyLemon, line 23: **issue**: You should only return the default if the `err` is a not-found error. If it's any other type of error, you should panic. That matches the `MustUnmarshal` behavior.

x/name/keeper/query_server.go
```diff
@@ -26,48 +27,90 @@ func (k Keeper) Resolve(c context.Context, request *types.QueryResolveRequest) (
...
 		return nil, err
 	}
-	record, err := k.GetRecordByName(ctx, name)
+	record, err := k.nameRecords.Get(ctx, name)
 	if err != nil {
-		return nil, err
-	}
-	if record == nil {
 		return nil, types.ErrNameNotBound
 	}
```
SpicyLemon: **issue (blocking)**: You need to differentiate the "not-found" error from the other errors in order to determine if you should return `types.ErrNameNotBound` or the returned error.

**suggestion**: Just switch this back to how it was.

x/name/keeper/query_server.go
```diff
@@ -26,48 +27,90 @@ func (k Keeper) Resolve(c context.Context, request *types.QueryResolveRequest) (
...
 		}
-		if accumulate {
-			names = append(names, record.Name)
+
+		if len(pageReq.Key) > 0 {
+			pageKey := string(pageReq.Key)
+			for i, name := range allNames {
+				if name == pageKey {
+					start = i + 1
+					break
```
SpicyLemon: **issue (blocking)**: The purpose of paginating a query is to save the node from doing work. This doesn't do that.

**suggestion**: Use `query.CollectionPaginate` using the address index to get all the data you need, and only the data requested.

x/name/types/keys.go
```diff
@@ -29,46 +34,125 @@ var (
...
+}
+
+// GetAddressKeyBytes returns the full address key
+func GetAddressKeyBytes(addr sdk.AccAddress, nameKey []byte) ([]byte, error) {
+	prefix, err := GetAddressKeyPrefix(addr)
+	if err != nil {
+		return nil, err
+	}
+	return append(prefix, nameKey...), nil
+}
```
SpicyLemon: **suggestion**: Move these to the migration.

The only place where these might still be needed is in the migration. You should move them there so that when we go to clean up the migration, everything to delete is all in one place.

This should go for anything that's only still needed for the migration.

x/name/keeper/keeper.go
```diff
@@ -88,173 +145,158 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-			return err
+	count := 0
+	err := k.nameRecords.Walk(ctx, nil, func(_ string, record types.NameRecord) (bool, error) {
+		count++
+		if prefix != nil && !bytes.Equal(prefix, types.NameKeyPrefix) {
+			nameBytes := []byte(record.Name)
+			if !bytes.HasPrefix(nameBytes, prefix) {
+				return false, nil
+			}
 		}
```
SpicyLemon: **issue (blocking)**: This is not how you do a prefix iteration, and this change narrows the focus of what this method current does (before this change).

Currently `IterateRecords` is used in two ways:
1. The `prefix` is `types.NameKeyPrefix` so this iterates over all name records.
2. The `prefix` is `types.GetAddressKeyPrefix(address)` so this iterates over the just the name records for the given address.

With this change, it's only case 1 because it'll only ever iterate over all name records. It can never handle case 2 because there's no way to provide the address anymore; it can't be in the `prefix` because you're not looping over the index which would have the address at the start of the key.

The purpose of all of this stuff is so that we can look up specific things. The point is to limit the amount of work the node needs to do to get this info.

**suggestion**: Refactor `IterateRecords` to take out the `prefix []byte` argument, and remove that check from the loop. Most places that call this are calling it with just the type byte for all the records. For those ones, you just take out that argument. But there is one call that is for the address index, but that's in `GetRecordsByAddress` which is getting overhauled anyway and should no longer call `IterateRecords`.

Also, the `prefix` was for the the keys, not the names. So the prefix check in the loop isn't correct anyway.

x/name/keeper/keeper.go
```diff
@@ -88,173 +145,158 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 		if err := handle(record); err != nil {
-			return err
+			ctx.Logger().Error("IterateRecords handle error", "error", err)
```
SpicyLemon: **issue**: Please remove this log line. Returning an error here causes that error to come back from `Walk`; so it'll get used/seen if needed. As a log line, it'll just show up to node operators and just be noise to them.

x/name/keeper/keeper_test.go
```diff
@@ -255,8 +251,8 @@ func (s *KeeperTestSuite) TestGetName() {
 		r, err := s.app.NameKeeper.GetRecordByName(s.ctx, "..name")
 		s.Require().Error(err)
 		s.Require().Nil(r)
-		s.Require().Equal("name segment cannot be empty: value provided for name is invalid", err.Error())
-		s.Require().False(s.app.NameKeeper.NameExists(s.ctx, "..name"))
+		s.Require().Equal("segment of name is too short", err.Error())
+		s.Require().True(s.app.NameKeeper.NameExists(s.ctx, "..name"))
 	})
```
SpicyLemon, line 254: **issue (blocking)**: This is not right. This test demonstrates that `NameExists` is **NOT** working correctly.

x/name/keeper/keeper_test.go
```diff
@@ -272,7 +268,7 @@ func (s *KeeperTestSuite) TestDeleteRecord() {
 	s.Run("delete invalid name", func() {
 		err := s.app.NameKeeper.DeleteRecord(s.ctx, "undefined.name")
 		s.Require().Error(err)
-		s.Require().Equal("no address bound to name", err.Error())
+		s.Require().ErrorContains(err, "collections: not found: key")
 	})
```
SpicyLemon, line 270: **issue**: The old error is better than the new one. The new one lacks context.

**suggestion**: In `DeleteRecord`, when the record already doesn't exist, make sure the error it returns contains `"no address bound to name"`. It can also contain `"collections: not found: key"`, but it should still also have the original error message.

x/name/keeper/keeper_test.go
```diff
@@ -317,12 +317,12 @@ func (s *KeeperTestSuite) TestModifyRecord() {
 		}
 		addr1Recs, err := s.app.NameKeeper.GetRecordsByAddress(s.ctx, s.user1Addr)
 		s.Require().NoError(err, "GetRecordsByAddress(user1)")
-		s.Assert().Equal(expUser1Recs, addr1Recs, "GetRecordsByAddress(user1)")
+		s.Assert().ElementsMatch(expUser1Recs, addr1Recs, "GetRecordsByAddress(user1)")
```
SpicyLemon: **issue (blocking)**: We need these results to be deterministic, so `ElementsMatch` is a bad assertion to use; it allows the order of the elements to change.

**suggestion**: Fix the order of the items in `expUser1Recs` and leave this check as `Equal`.

It is concerning, though, that the order is changing. It doesn't seem like that should happen.

x/name/keeper/keeper_test.go
```diff
@@ -638,7 +479,7 @@ func (s *KeeperTestSuite) TestCreateRootNameProposals() {
 			name:       "..badroot",
 			owner:      s.user1Addr,
 			restricted: false,
-			err:        fmt.Errorf("segment of name is too short"),
+			err:        fmt.Errorf("name is already bound to an address"),
 		},
```
SpicyLemon, line 485: issue: This doesn't make any sense to me. How can that be the expected error when the name should never be bound in the first place?

x/name/types/keys_test.go
```diff
@@ -117,3 +104,95 @@ func mustHexDecode(h string) []byte {
...
 	return result
 }
+
+func (s *NameKeyTestSuite) TestHashedStringKeyCodec() {
+	codec := HashedStringKeyCodec{}
+
+	tests := []struct {
+		name      string
+		input     string
+		expectErr bool
```
SpicyLemon: **issue (blocking)**: When testing something with errors, it is insufficient to merely test whether an error has occurred, you MUST check the actual contents of the error against an expected value.

I.e. `expectErr bool` is very bad, don't do it. Use `expErr string`. If, in your test code, you need to do something different when expecting an error, use `if len(tc.expErr) > 0` (or `== 0` if appropriate).

In this case, all test cases have `false` for it anyway, so you can just remove that field altogether.

x/name/types/keys_test.go
```diff
@@ -117,3 +104,95 @@ func mustHexDecode(h string) []byte {
...
+	}{
+		{"empty string", "", false},
+		{"single char", "a", false},
+		{"short string", "short", false},
+		{"domain style", "example.domain", false},
+		{"multi-level domain", "one.two.three.four", false},
+		{"long string", strings.Repeat("x", 100), false},
+		{"whitespace only", "   ", false},
+		{"contains whitespace", "some domain.name", false},
+		{"special chars", "!@#$%^&*()_+{}|:\"<>?", false},
```
SpicyLemon: **issue (blocking)**: Always always always name your fields when instantiating a struct.

E.g. instead of `{"empty string", "", false}`, use `{name: "empty string", input: "", expectErr: false}

We even have an old issue to go through all of our unit tests and fix this exact thing.

x/name/types/keys_test.go
```diff
@@ -117,3 +104,95 @@ func mustHexDecode(h string) []byte {
...
+			buffer := make([]byte, size)
+			n, err := codec.Encode(buffer, tc.input)
+
+			if tc.expectErr {
+				s.Error(err)
+				return
+			}
+
+			s.Require().NoError(err)
+			s.Equal(size, n)
```
SpicyLemon: **issue (blocking)**: Missing `msgAndArgs`.

I'm just going to make this comment once, but it applies to everything, and there are many more places that need to address this.

x/name/types/keys_test.go
```diff
@@ -117,3 +104,95 @@ func mustHexDecode(h string) []byte {
...
+			s.NotEqual(tc.input, out, "decoded string should be base64 of hash, not the original input")
+
+			nonTermSize := codec.SizeNonTerminal(tc.input)
+			s.Equal(sha256.Size, nonTermSize, "non-terminal size must also match sha256 size")
+
+			nonTermBuf := make([]byte, nonTermSize)
+			n2, err := codec.EncodeNonTerminal(nonTermBuf, tc.input)
+			s.Require().NoError(err)
+			s.Equal(nonTermSize, n2)
+			s.Equal(expectedHash, nonTermBuf[:n2], "non-terminal hash must match terminal hash")
```
SpicyLemon: suggestion: Each of the little blocks in here look like independent tests. Wrap each in `s.Run(...)` so that they're run independently, but still report back on the test case as a whole. You'll end up with nested `s.Run(...)` calls.

There's more stuff below this comment that I'm referring to; GitHub wouldn't let me select all the lines though.

x/name/keeper/query_server.go
```diff
@@ -26,48 +27,90 @@ func (k Keeper) Resolve(c context.Context, request *types.QueryResolveRequest) (
...
 		return nil, types.ErrInvalidAddress
 	}
-	key, err := types.GetAddressKeyPrefix(accAddr)
+
+	var pageReq *query.PageRequest
+	if request.Pagination != nil {
+		pageReq = request.Pagination
+	}
+
+	recordByAddress, err := k.GetRecordsByAddress(ctx, accAddr)
```
SpicyLemon: **issue (blocking)**: This cannot be done this way.

The purpose of pagination is to save the node from doing work. As you've got it, you're doing all the work every time, and then just limiting what's seen. You need to use the index to get only the records being requested by the pagination params.

x/name/keeper/migrations.go
```diff
@@ -1,11 +1,104 @@
...
 // Migrator is a struct for handling in-place store migrations.
 type Migrator struct {
 	keeper Keeper
 }
 
+// Legacy prefixes - OLD values moved here for migration
+var (
+	LegacyNameKeyPrefix     = []byte{0x03}
+	LegacyAddressKeyPrefix  = []byte{0x05}
+	LegacyNameParamStoreKey = []byte{0x06}
```
SpicyLemon: All of the existing entries need to be deleted as part of the migration.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+
+func (c HashedStringKeyCodec) DecodeNonTerminal(buffer []byte) (int, string, error) {
+	if len(buffer) < sha256.Size {
+		return 0, "", fmt.Errorf("%w: a hashed name key needs %d bytes, got %d",
+			ErrNameInvalid, sha256.Size, len(buffer))
+	}
+	return sha256.Size, base64.StdEncoding.EncodeToString(buffer[:sha256.Size]), nil
+}
+func (c HashedStringKeyCodec) DecodeJSON(b []byte) (string, error) {
```
SpicyLemon, line 100: Missing empty line.

x/name/types/keys.go
```diff
@@ -2,11 +2,16 @@ package types
 
 import (
 	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"strings"
 
+	"cosmossdk.io/collections"
+	callcodec "cosmossdk.io/collections/codec"
```
SpicyLemon: Should be `collcodec`.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+	AddressKeyPrefix = collections.NewPrefix(0x08)
 	// NameParamStoreKey key for marker module's params
-	NameParamStoreKey = []byte{0x06}
+	NameParamStoreKey = collections.NewPrefix(0x09)
+)
+
+// Legacy prefixes - OLD values moved here for migration
+var (
+	LegacyNameKeyPrefix     = []byte{0x03}
+	LegacyAddressKeyPrefix  = []byte{0x05}
```
SpicyLemon, line 40: Part of the migration needs to go through and delete all the old data. These shouldn't be needed anymore other than that.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+	err := json.Unmarshal(b, &s)
+	return s, err
+}
+
+func (c HashedStringKeyCodec) SizeNonTerminal(_ string) int {
+	return sha256.Size
+}
+
+func (c HashedStringKeyCodec) ComputeHash(name string) []byte {
+	sum := sha256.Sum256([]byte(trimNameSegments(name)))
```
SpicyLemon: The thing being hashed should also be lower-cased. Use `NormalizeName` here instead of `trimNameSegments`.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+}
+
+// GetNameKeyBytes returns the name key in the same format as before (0x03 + sha256(name))
+func GetNameKeyBytes(name string) ([]byte, error) {
+	hash, err := ComputeNameHash(name)
+	if err != nil {
 		return nil, err
 	}
+	return append(append([]byte{}, NameKeyPrefix...), hash...), nil
+}
```
SpicyLemon: Please delete this. Since we're using collections now, we should never be manually creating any keys. This is only used in unit tests. Those unit tests are technically not doing the right thing anymore.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+
+// trimNameSegments removes whitespace around each segment so " a . b " and "a.b"
+// produce the same key.
+func trimNameSegments(name string) string {
+	segments := strings.Split(name, ".")
+	for i, segment := range segments {
+		segments[i] = strings.TrimSpace(segment)
+	}
+	return strings.Join(segments, ".")
 }
```
SpicyLemon: Please delete this and replace uses of it with `NormalizeName`.

x/name/types/keys.go
```diff
@@ -22,53 +27,157 @@ const (
...
+			return "", fmt.Errorf("name segment cannot be empty: %w", ErrNameInvalid)
+		}
+	}
+	return normalized, nil
+}
+
+// LegacyComputeNameHash reproduces the old name-key hash by hashing segments in reverse order.
+//
+// Deprecated: used only by the 3->4 migration and its tests. Use ComputeNameHash instead.
+func LegacyComputeNameHash(name string) ([]byte, error) {
```
SpicyLemon: Please move `LegacyComputeNameHash` and `LegacyGetNameKeyBytes` into the migrations so that they get deleted with all that stuff. Or better yet, delete them if they're not needed anymore.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+		return err
+	}
+	nameUpdateEvent := types.NewEventNameUpdate(addr.String(), normalizedName, restrict)
 	return ctx.EventManager().EmitTypedEvent(nameUpdateEvent)
 }
 
 // GetRecordByName resolves a record by name.
 func (k Keeper) GetRecordByName(ctx sdk.Context, name string) (record *types.NameRecord, err error) {
-	key, err := types.GetNameKeyPrefix(name)
+	key, err := types.GetNameKey(name)
```
SpicyLemon: Please use `k.Normalize` here for consistency.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-	record = &types.NameRecord{}
-	err = keeper.cdc.Unmarshal(bz, record)
-	return record, err
+	return &namerecord, nil
 }
 
 // NameExists returns true if store contains a record for the given name.
 func (k Keeper) NameExists(ctx sdk.Context, name string) bool {
-	key, err := types.GetNameKeyPrefix(name)
+	key, err := types.GetNameKey(name)
```
SpicyLemon: Please use `k.Normalize` here instead of `GetNameKey`.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 
 // DeleteRecord removes a name record from the kvstore.
 func (k Keeper) DeleteRecord(ctx sdk.Context, name string) error {
-	// Need the record to clear the address index
-	record, err := k.GetRecordByName(ctx, name)
-	if err != nil {
-		return err
-	}
-	address, err := sdk.AccAddressFromBech32(record.Address)
+	key, err := types.GetNameKey(name)
```
SpicyLemon: Please use `k.Normalize` here instead of `GetNameKey`.

SpicyLemon (commented):
I'm not quite done looking through all the stuff, but here's a few things.

The big one is that the migration is missing the part where it deletes all the old data.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-func getNameRecord(ctx sdk.Context, keeper Keeper, key []byte) (record *types.NameRecord, err error) {
-	store := ctx.KVStore(keeper.storeKey)
-	if !store.Has(key) {
-		return nil, types.ErrNameNotBound
+	namerecord, err := k.nameRecords.Get(ctx, normalizedName)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return nil, types.ErrNameNotBound
+		}
+		return nil, err
```
SpicyLemon, line 211: This error needs to have context added to it.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+	record := types.NewNameRecord(normalizedName, addr, restrict)
+	if err := record.Validate(); err != nil {
 		return err
 	}
-
-	nameUpdateEvent := types.NewEventNameUpdate(addr.String(), name, restrict)
-
+	// Update the record.
+	if err := k.nameRecords.Set(ctx, normalizedName, record); err != nil {
+		return err
```
SpicyLemon, line 194: This error needs context added to it.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-		oldAddrKey, err = types.GetAddressKeyPrefix(oldAddr)
-		if err != nil {
-			return types.ErrInvalidAddress.Wrapf("invalid existing %s record address format: %v", name, err)
-		}
-		oldAddrKey = append(oldAddrKey, oldNameKeyPre...)
-		store := ctx.KVStore(k.storeKey)
-		store.Delete(oldAddrKey)
+	exists, err := k.nameRecords.Has(ctx, normalizedName)
+	if err != nil {
+		return err
```
SpicyLemon, line 181: This error needs context added to it.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 		return err
 	}
 	if err = types.ValidateAddress(addr); err != nil {
 		return types.ErrInvalidAddress.Wrap(err.Error())
 	}
-
-	if err = k.addRecord(ctx, name, addr, restrict, false); err != nil {
+	exists, err := k.nameRecords.Has(ctx, normalizedName)
+	if err != nil {
+		return err
```
SpicyLemon, line 152: This error needs context.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+		return types.ErrNameAlreadyBound
+	}
+	record := types.NewNameRecord(normalizedName, addr, restrict)
+	if err := record.Validate(); err != nil {
 		return err
 	}
 
-	nameBoundEvent := types.NewEventNameBound(addr.String(), name, restrict)
+	if err := k.nameRecords.Set(ctx, normalizedName, record); err != nil {
+		return err
```
SpicyLemon, line 163: This error needs more context.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-			records = append(records, record)
-		}
-		return nil
-	}
-	// Calculate address prefix
-	addrPrefix, err := types.GetAddressKeyPrefix(address)
+
+	iter, err := k.nameRecords.Indexes.AddrIndex.MatchExact(ctx, address)
 	if err != nil {
 		return nil, err
```
SpicyLemon, line 232: This error needs some context.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 		return err
 	}
-	// Delete the main name record
-	key, err := types.GetNameKeyPrefix(name)
+	record, err := k.nameRecords.Get(ctx, normalizedName)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return types.ErrNameNotBound
+		}
 		return err
```
SpicyLemon, line 261: This error needs some context too.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
 	}
-	store := ctx.KVStore(k.storeKey)
-	store.Delete(key)
-	// Delete the address index record
-	addrPrefix, err := types.GetAddressKeyPrefix(address)
-	if err != nil {
+	if err := k.nameRecords.Remove(ctx, normalizedName); err != nil {
 		return err
```
SpicyLemon, line 264: This error needs some context.

SpicyLemon (commented):
Founds some spots that need to have extra context on the returned errors. Basically, anywhere where the error comes from the collections, it should be wrapped with added context.

x/name/keeper/migrations.go
```diff
@@ -9,3 +21,139 @@ type Migrator struct {
...
+	for _, prefix := range [][]byte{types.LegacyNameKeyPrefix, types.LegacyAddressKeyPrefix} {
+		iter, err := kvStore.Iterator(prefix, storetypes.PrefixEndBytes(prefix))
+		if err != nil {
+			return fmt.Errorf("could not iterate legacy prefix %X: %w", prefix, err)
+		}
+		var keys [][]byte
+		for ; iter.Valid(); iter.Next() {
+			keys = append(keys, append([]byte{}, iter.Key()...))
+		}
+		defer iter.Close() //nolint:errcheck // close error safe to ignore in this context.
```
SpicyLemon, line 119: You don't need the `defer` here. It's okay to just call `iter.Close()` directly.

x/name/keeper/migrations.go
```diff
@@ -9,3 +21,139 @@ type Migrator struct {
...
+	return hsh.Sum(nil), nil
+}
+
+// LegacyGetNameKeyBytes returns the full store key a name occupied before the 3->4 migration.
+func LegacyGetNameKeyBytes(name string) ([]byte, error) {
+	hash, err := LegacyComputeNameHash(name)
+	if err != nil {
+		return nil, err
+	}
+	return append(append([]byte{}, types.NameKeyPrefix...), hash...), nil
```
SpicyLemon, line 158: This needs to use the legacy name key prefix.

x/name/types/keys.go
```diff
@@ -22,53 +27,101 @@ const (
...
 	// NameParamStoreKey key for marker module's params
-	NameParamStoreKey = []byte{0x06}
+	NameParamStoreKey = collections.NewPrefix(0x09)
+)
+
+// Legacy prefixes - OLD values moved here for migration
+var (
+	LegacyNameKeyPrefix     = []byte{0x03}
+	LegacyAddressKeyPrefix  = []byte{0x05}
+	LegacyNameParamStoreKey = []byte{0x06}
```
SpicyLemon, line 41: These should be moved to the migration file. They shouldn't be needed anywhere outside the migration, and should be deleted when we delete the migration.

x/name/keeper/migrations.go
SpicyLemon: Also, I feel like this should have been caught in a unit test (e.g. `TestMigration()`).

Any state entries stored using `LegacyGetNameKeyBytes` as it is would be ignored by the migration. You'd just end up with the name record written with the wrong key.

x/name/keeper/migration_test.go
```diff
@@ -0,0 +1,94 @@
...
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	"github.com/cosmos/cosmos-sdk/testutil"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/provenance-io/provenance/x/name/keeper"
+	"github.com/provenance-io/provenance/x/name/types"
+	"github.com/stretchr/testify/suite"
+)
+
+type MigrationTestSuite struct {
```
SpicyLemon, line 17: There is no `func TestMigrationTestSuite(t *Testing.T) { suite.Run(t, new(MigrationTestSuite)) }` test/func defined, so none of these tests are even being run.

x/name/keeper/migrations.go
```diff
@@ -9,3 +21,139 @@ type Migrator struct {
...
+	}
+	if bz == nil {
+		return nil
+	}
+
+	// Start with the default params, not a zero-value Params. Proto3 omits false and 0
+	// values, so unmarshalling into a zero-value Params could turn a stored
+	// AllowUnrestrictedNames=false into the default true and change the chain's params
+	// during the upgrade.
+	params := types.DefaultParams()
```
SpicyLemon, line 95: I don't think this is correct.

If you start with the default params, then `AllowUnrestrictedNames` starts out as `true`. Then, if it's `false` in state, the unmarshal ignores it and it and will stay `true` (but should be false). If it's `true` in state, then the unmarshal sees it and sets it to `true`. Either way, it's always going to be `true` after the unmarshal.

If you start with a zero-value, though, the `AllowUnrestrictedNames` field starts out `false`. Then, if it's `false` in state, unmarshal ignores it and it stays `false`. If it's `true` in state, the unmarshal sees it and sets it to `true`. This maintains the value through the migration.

x/name/keeper/params.go
```diff
@@ -1,28 +1,33 @@
...
 
 // SetParams sets the name parameters to the store.
 func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
-	store := ctx.KVStore(k.storeKey)
-	bz := k.cdc.MustMarshal(&params)
-	store.Set(types.NameParamStoreKey, bz)
+	if err := k.paramsStore.Set(ctx, params); err != nil {
+		k.Logger(ctx).Error("failed to set params", "error", err)
+	}
 }
```
SpicyLemon, line 31: This new version of `SetParams` should now also return an error instead of panicking.

`SetParams` is called in two places:
1. `InitGenesis`: It's okay to make it panick here.
2. `UpdateParams`: This is a governance endpoint. If `SetParams` encounters an error, we want that error to end up returned by this endpoint too so that it's easier to find out what went wrong (and that something went wrong).

x/attribute/keeper/query_server_test.go
```diff
@@ -44,6 +45,8 @@ func (s *QueryServerTestSuite) SetupTest() {
 	s.ctx = s.app.BaseApp.NewContext(true)
 	s.app.AccountKeeper.Params.Set(s.ctx, authtypes.DefaultParams())
 	s.app.BankKeeper.SetParams(s.ctx, banktypes.DefaultParams())
+	s.app.NameKeeper.SetParams(s.ctx, nametypes.DefaultParams())
+	s.Require().NoError(s.app.NameKeeper.SetParams(s.ctx, nametypes.DefaultParams()), "NameKeeper.SetParams")
```
SpicyLemon: You can delete the first call (that doesn't have the error check on it).

x/name/keeper/export_test.go
```diff
@@ -4,8 +4,19 @@ import sdk "github.com/cosmos/cosmos-sdk/types"
...
+)
 
 // AddRecord is a TEST ONLY exposure of addRecord.
 func (k Keeper) AddRecord(ctx sdk.Context, name string, addr sdk.AccAddress, restrict, isModifiable bool) error {
-	return k.addRecord(ctx, name, addr, restrict, isModifiable)
+	if isModifiable {
+		return k.UpdateNameRecord(ctx, name, addr, restrict)
+	}
+	return k.SetNameRecord(ctx, name, addr, restrict)
 }
```
SpicyLemon: This should be deleted since it's not used. It's not correct anyway.

x/name/keeper/msg_server.go
```diff
@@ -207,7 +207,9 @@ func (s msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
 		return nil, err
 	}
 
-	s.SetParams(ctx, msg.Params)
+	if err := s.SetParams(ctx, msg.Params); err != nil {
+		panic(err)
```
SpicyLemon: This shouldn't be a panic. The error should be returned.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+		return fmt.Errorf("could not update name record %q: %w", normalizedName, err)
+	}
+	nameUpdateEvent := types.NewEventNameUpdate(addr.String(), normalizedName, restrict)
 	return ctx.EventManager().EmitTypedEvent(nameUpdateEvent)
 }
 
 // GetRecordByName resolves a record by name.
 func (k Keeper) GetRecordByName(ctx sdk.Context, name string) (record *types.NameRecord, err error) {
-	key, err := types.GetNameKeyPrefix(name)
+	normalizedName, err := k.Normalize(ctx, name)
```
SpicyLemon: We don't want to normalize the name here using params, just use `types.NormalizeName` here.

The problem is that, if the params ever change there might be previously added names that no longer agree with the new params (but were fine with the old ones). But by enforcing normalization here (against the params), you'd be blocking the lookup of the old entry.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-	record = &types.NameRecord{}
-	err = keeper.cdc.Unmarshal(bz, record)
-	return record, err
+	return &namerecord, nil
 }
 
 // NameExists returns true if store contains a record for the given name.
 func (k Keeper) NameExists(ctx sdk.Context, name string) bool {
-	key, err := types.GetNameKeyPrefix(name)
+	normalizedName, err := k.Normalize(ctx, name)
```
SpicyLemon: Here too, you should just use `types.NormalizeName` (and don't involve the params).

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
-			records = append(records, record)
-		}
-		return nil
-	}
-	// Calculate address prefix
-	addrPrefix, err := types.GetAddressKeyPrefix(address)
+
+	iter, err := k.nameRecords.Indexes.AddrIndex.MatchExact(ctx, address)
 	if err != nil {
 		return nil, err
```
SpicyLemon: This error still needs some context.

x/name/keeper/keeper.go
```diff
@@ -88,321 +140,207 @@ func (k Keeper) ResolvesTo(ctx sdk.Context, name string, addr sdk.AccAddress) bo
...
+		records = append(records, record)
 	}
 	return records, nil
 }
 
 // DeleteRecord removes a name record from the kvstore.
 func (k Keeper) DeleteRecord(ctx sdk.Context, name string) error {
-	// Need the record to clear the address index
-	record, err := k.GetRecordByName(ctx, name)
+	normalizedName, err := k.Normalize(ctx, name)
```
SpicyLemon: Don't normalize it using the module's params, just use `types.NormalizeName`.

SpicyLemon (approved):

Issue #902, opened 2022-06-27
Publish container images to GHCR.


iramiller:
## Summary

Currently containers are published to docker hub.  In addition to the hub the GitHub container registry directly associated with the Provenance project should be used.

## Problem Definition

The docker hub is not the only container registry available.  To increase awareness of the published container images for the `provenanced` project we should also publish on the `ghcr` within the `provenance` project.

## Proposal

Add an additional publishing step for GitHub container registry along with the current standard docker hub version.

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2435, opened 2025-08-28
Publish docker images to GitHub registry.

nagarajdivine:
## Description

* (GHRC) github container registry publishing for docker images
closes: #902 Publish container images to GHCR

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
  - Official container images now published to GitHub Container Registry (GHCR) alongside Docker Hub with multi-architecture builds (linux/amd64, linux/arm64) and edge/PR/release tagging.

- **Chores**
  - CI workflow updated to add GHCR publishing job, environment variable for GHCR image, metadata tagging, registry login, and build-and-push steps.

- **Documentation**
  - Added unreleased changelog entry documenting GHCR publishing.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -120,6 +120,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+
+      - run: go mod vendor
+
       - uses: docker/setup-buildx-action@v3
 
       - uses: docker/setup-qemu-action@v3


.github/workflows/docker.yml
```diff
@@ -43,11 +44,30 @@ jobs:
...
+        id: meta-ghcr
+        if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
+        with:
+          images: ${{ env.GHCR_IMAGE }}
+          flavor: |
+            prefix=heighliner-,onlatest=true
+          tags: |
+            type=edge
+            type=ref,event=pr
+            type=ref,event=tag
```
SpicyLemon: **issue (blocking)**: Nothing should be changing for the heighliner image. It's special, and managed by strangelove.

.github/workflows/docker.yml
```diff
@@ -43,11 +44,30 @@ jobs:
...
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+        
+      - uses: docker/login-action@v3
+        if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
SpicyLemon: **issue (blocking)**: Nothing should be changing for the heighliner image. It's special, and managed by strangelove.

.github/workflows/docker.yml
```diff
@@ -68,8 +88,35 @@ jobs:
...
+            if [[ "$dockerhub_tag" == "provenanceio/provenance:"* ]]; then
+              # Extract the tag suffix
+              TAG_SUFFIX="${dockerhub_tag#provenanceio/provenance:}"
+              GHCR_TAG="ghcr.io/provenance-io/provenance:${TAG_SUFFIX}"
+              
+              echo "Mirroring $dockerhub_tag to $GHCR_TAG"
+              docker buildx imagetools create --tag "$GHCR_TAG" "$dockerhub_tag"
+            fi
+          done
+
```
SpicyLemon: **issue (blocking)**: Nothing should be changing for the heighliner image. It's special, and managed by strangelove.

.github/workflows/docker.yml
```diff
@@ -92,11 +139,30 @@ jobs:
...
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - uses: docker/login-action@v3
+        if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
```
SpicyLemon, line 151: **issue (blocking)**: You need to create a whole new job to do this publishing.

As you've got it, you've duplicated some setup actions that will conflict with each other. So one or both of the publish steps isn't going to happen like it should. Additionally, this `docker` job takes a long time already (sometimes over 30 minutes), so it'd be good to separate the two jobs. 

Another reason to make it two jobs is so that it's easier to retry stuff if only one of them fails.

.github/workflows/docker.yml
```diff
@@ -107,3 +173,17 @@ jobs:
...
+        if: (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && github.event_name != 'pull_request'
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          target: run
+          build-args: |
+            VERSION=${{ steps.meta-ghcr.outputs.version }}
+          platforms: linux/amd64,linux/arm64
+          file: docker/blockchain/Dockerfile
+          push: true
```
SpicyLemon: **issue (blocking)**: If the other steps run, this one needs to run to. It just shouldn't do the **push** on pull requests.

There's lots of stuff that this step does, and is the main step where something will go wrong. It's important to see that it completes as much as possible, e.g. the building of the container.

**suggestion**: Take the `&& github.event_name != 'pull_request'` out of the `if` and make `push` the same as the others: `${{ github.event_name != 'pull_request' }}`.

docker/blockchain/Dockerfile
```diff
@@ -17,15 +17,16 @@ COPY vendor/ ./vendor/
 COPY testutil/ ./testutil/
 COPY .git/ ./.git/
 COPY contrib/ ./contrib/
+COPY legacy_protos/ ./legacy_protos/
```
SpicyLemon: **issue**: You don't need this since there's no go code in it.

Please undo all changes to this file. The whitespace changes below aren't wanted since they're untabbing code in an `if` block.

docker/blockchain/Dockerfile
```diff
@@ -24,8 +24,8 @@ COPY Makefile sims.mk ./
 ENV VERSION=$VERSION
 RUN ARCH=$(uname -m) && \
     if [ "$ARCH" != "x86_64" ] && [ "$ARCH" != "aarch64" ]; then \
-        echo "Unsupported architecture (required: x86_64 or aarch64): $ARCH"; \
-        exit 1; \
+    echo "Unsupported architecture (required: x86_64 or aarch64): $ARCH"; \
+    exit 1; \
```
SpicyLemon: Please undo this change too as well as the other changes in this file.

This PR shouldn't need to change anything in here.

Also, please notice that these two lines are the block content of an `if` statement and **should** be tabbed in.

.github/workflows/docker.yml
```diff
@@ -110,3 +111,48 @@ jobs:
           file: docker/blockchain/Dockerfile
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
+
+  ghcr-publish:
+    runs-on: ubuntu-latest
+    needs: [docker]  # Wait for docker job to complete
```
SpicyLemon: **Suggestion**: Remove this `needs` line.

I don't see anything in this `ghcr-publish` job that relies on the `docker` job. And that `docker` job can take quite a long time (over 30 minutes sometimes). It'd be best to not wait on it if we don't have to. Also, by having it as a `needs` here, it won't run if `docker` fails, which is not something we want to have happen. We want this publishing to be independent of the docker publishing so that we can replay one or the other if needed.

.github/workflows/docker.yml
```diff
@@ -110,3 +111,48 @@ jobs:
...
+          labels: |
+            org.opencontainers.image.source=https://github.com/provenance-io/provenance
+            org.opencontainers.image.description=Provenance Blockchain Node
+            org.opencontainers.image.licenses=Apache-2.0
+
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
SpicyLemon: **Question**: I haven't interacted with the GitHub docker repository before. Is it going to be weird to have these uploaded by the `github.actor`? The `github.actor` is just the last account that ran the workflow. I.e. it's one of us. Should we be using the bot account for this?

nagarajdivine: @SpicyLemon, after checking,  `github.actor` is just the `username` of whoever merged to main, so their name will show up as the one who ran the workflow. I think it’s better to use a `secret` tied to a bot account like `provenance-bot` or `provenance-io`, so the actions come from the organization instead of a person.

SpicyLemon: Okay, let's use the following then:
username: `129784868+provenanceio-bot@users.noreply.github.com`
password: `${{ secrets.BOT_GHCR_PAT }}`

I just created that `BOT_GHCR_PAT` secret using a classic personal access token with both read and write for packages.

nagarajdivine: will update it :)

.github/workflows/docker.yml
```diff
@@ -110,3 +111,47 @@ jobs:
           file: docker/blockchain/Dockerfile
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
+
+  ghcr-publish:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')
```
SpicyLemon: **suggestion**: Let's get rid of this `if` line so that it runs the same as the other docker jobs (and so we can see it run in this PR before merging).

Issue #2499, opened 2025-10-17
Switch to collections in the trigger module.


SpicyLemon:
## Summary

Update the `x/trigger` module to use collections for state storage.

## Problem Definition

Collections make it easier to interact with state, and we should switch to them for our state storage interactions.

## Proposal

Update the trigger module to use collections for state storage. 

It should be possible to do this without needing a migration.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2546, opened 2025-11-21
Migrate trigger module from KV store to collections.

nagarajdivine:
## Description

* Migrate trigger module from KV store to collections.
closes: #2499 

---

write a little note why.

## Summary by CodeRabbit

* **Changelog**
  * Migrated trigger module to collection-backed state storage.

* **New Features**
  * Collection-backed APIs for triggers, listeners, and queue: existence checks, counts, bulk removals, hashed event keys, enqueue/dequeue/peek, iteration, and paginated trigger queries.
  * Stable trigger ID sequencing and strict 32-byte event-hash key support.

* **Bug Fixes**
  * Clearer not-found mapping, explicit error propagation, improved logging and resilience.

* **Chores**
  * Storage migration, init/export validation, and migration tooling added.

* **Tests**
  * Expanded coverage for collections, queue, listeners, migrations and backward compatibility.

### DIFF
--- a/x/trigger/keeper/event_listener.go
+++ b/x/trigger/keeper/event_listener.go
@@ -94,7 +94,7 @@ func (k Keeper) IterateEventListeners(ctx sdk.Context, eventName string, handle 
 	if err != nil {
 		return err
 	}
-	defer iter.Close()
+	defer iter.Close() //nolint:errcheck // ignoring close error on iterator: not critical for this context.
 
 	for ; iter.Valid(); iter.Next() {
 		key, err := iter.Key()
--- a/x/oracle/module/module.go
+++ b/x/oracle/module/module.go
@@ -74,6 +74,7 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, _ sdkclient.TxEncodin
 }
 
 // RegisterRESTRoutes registers the REST routes for the oracle module.
+//
 // Deprecated: RegisterRESTRoutes is deprecated.
 func (AppModuleBasic) RegisterRESTRoutes(_ sdkclient.Context, _ *mux.Router) {}
 
--- a/x/trigger/keeper/migrations_v2.go
+++ b/x/trigger/keeper/migrations_v2.go
@@ -34,9 +34,12 @@ func (m Migrator) DeleteGasLimits(ctx sdk.Context, logger log.Logger) {
 		logger.Error("Failed to create iterator", "err", err)
 		return
 	}
+
 	closeIter := func() {
 		if iter != nil {
-			iter.Close()
+			if err := iter.Close(); err != nil {
+				m.keeper.Logger(ctx).Error("failed to close iterator during migration", "err", err)
+			}
 			iter = nil
 		}
 	}
--- a/x/trigger/keeper/queue.go
+++ b/x/trigger/keeper/queue.go
@@ -12,7 +12,7 @@ import (
 
 // QueueTrigger Creates a QueuedTrigger and Enqueues it
 func (k Keeper) QueueTrigger(ctx sdk.Context, trigger types.Trigger) {
-	item := types.NewQueuedTrigger(trigger, ctx.BlockTime().UTC(), uint64(ctx.BlockHeight()))
+	item := types.NewQueuedTrigger(trigger, ctx.BlockTime().UTC(), uint64(ctx.BlockHeight())) //nolint:gosec // G115: block height guaranteed non-negative in Cosmos SDK.
 	err := k.Enqueue(ctx, item)
 	if err != nil {
 		ctx.Logger().Error("failed to queue trigger", "trigger_id", trigger.Id, "err", err)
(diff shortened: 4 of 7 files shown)


x/trigger/types/keys.go
```diff
@@ -94,38 +96,38 @@ func GetTriggerKey(id TriggerID) []byte {
 	triggerIDBytes := make([]byte, TriggerIDLength)
 	binary.BigEndian.PutUint64(triggerIDBytes, id)
 
-	key := TriggerKeyPrefix
+	key := []byte(TriggerKeyPrefix)
```
SpicyLemon: Should this be `.Bytes()` instead?

x/trigger/keeper/event_listener.go
```diff
@@ -1,58 +1,108 @@
...
+func (k Keeper) SetEventListener(ctx sdk.Context, trigger triggertypes.Trigger) error {
+	event, err := trigger.GetTriggerEventI()
+	if err != nil {
+		return err
+	}
+	eventHash, err := getEventNameHash(event.GetEventPrefix())
+	if err != nil {
+		return err
+	}
+	key := collections.Join3(eventHash, event.GetEventOrder(), trigger.GetId())
```
SpicyLemon: This is done the same way in a lot of places. Its probably best to just refactor `GetEventListenerKey` to return the `collections.Join3` result.

x/trigger/keeper/event_listener.go
```diff
@@ -1,58 +1,108 @@
...
 
 // SetEventListener Adds the trigger to the event listener store.
-func (k Keeper) SetEventListener(ctx sdk.Context, trigger triggertypes.Trigger) {
-	store := ctx.KVStore(k.storeKey)
-	event, _ := trigger.GetTriggerEventI()
-	store.Set(triggertypes.GetEventListenerKey(event.GetEventPrefix(), event.GetEventOrder(), trigger.GetId()), []byte{})
+func (k Keeper) SetEventListener(ctx sdk.Context, trigger triggertypes.Trigger) error {
+	event, err := trigger.GetTriggerEventI()
+	if err != nil {
+		return err
```
SpicyLemon: Please get into the habit of wrapping errors before returning them. e.g. `fmt.Errorf("could not get trigger event: %w", err)`.

Doing this helps you reconstruct what exactly went wrong from the contents of an error message.

There are a few places where it's okay to not wrap your error, e.g. a private validation helper that only exists to segment a tricky piece of validation from the rest of the validation, where any errors the helper returns provide all the needed context already.

x/trigger/keeper/event_listener.go
```diff
@@ -62,3 +112,43 @@ func (k Keeper) IterateEventListeners(ctx sdk.Context, eventName string, handle
...
+
+// RemoveAllEventListenersForTrigger removes all event listeners for a specific trigger
+func (k Keeper) RemoveAllEventListenersForTrigger(ctx sdk.Context, triggerID uint64) error {
+	trigger, err := k.GetTrigger(ctx, triggerID)
+	if err != nil {
+		return err
+	}
+	_, err = k.RemoveEventListener(ctx, trigger)
+	return err
+}
```
SpicyLemon: I'm confused by this. It only removes one event listener. From what I can tell, all it does is delete a trigger event based on a trigger id. I suggest renaming it to `RemoveEventListenerForTriggerID`.

I also don't think it's in use anywhere, so consider just deleting it.

x/trigger/keeper/genesis.go
```diff
@@ -30,18 +39,25 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
 		panic(err)
 	}
 
-	k.setTriggerID(ctx, data.TriggerId)
-	k.setQueueStartIndex(ctx, data.QueueStart)
+	must := func(err error, msg string) {
+		if err != nil {
+			panic(fmt.Errorf("%s: %w", msg, err))
+		}
+	}
```
SpicyLemon: Please make this an actual function instead of an anonymous one.

Anonymous functions require more computing power than normal ones. Since this one doesn't use any variables from its containing func, it can just as easily be made into a normal function.

x/trigger/keeper/query_server.go
```diff
@@ -39,28 +43,23 @@ func (k Keeper) Triggers(ctx context.Context, req *types.QueryTriggersRequest) (
...
-
+	triggers, pageRes, err := query.CollectionPaginate[uint64, types.Trigger, *collections.Map[uint64, types.Trigger], types.Trigger](
+		sdkCtx,
+		&k.TriggersMap,
+		pagination,
+		func(key uint64, trigger types.Trigger) (types.Trigger, error) {
+			trigger.Id = key
+			return trigger, nil
+		},
+	)
```
SpicyLemon: Since all of the types are defined in the arguments provided to `CollectionPaginate`, you don't need to list them in the call.
```suggestion
	triggers, pageRes, err := query.CollectionPaginate(sdkCtx, &k.TriggersMap, pagination,
		func(key uint64, trigger types.Trigger) (types.Trigger, error) {
			trigger.Id = key
			return trigger, nil
		},
	)
```

Issue #2550, opened 2025-12-01
Allow oracles to update the flatfees conversion factor


SpicyLemon:
## Summary

Allow one or more oracle addresses to be able to update the flatfees conversion factor without needing to use a governance proposal.

## Problem Definition

Currently, the only way to update the flatfees conversion factor is through a governance proposal. But that's a whole lot of work for such a small task. It would be better if there were also a set of addresses (oracles) that were allowed to update the conversion factor.

## Proposal

In the flatfees module, add two new governance endpoints, one for adding, and one for removing oracle addresses. The module will keep track of this list in its params. Then, in the `UpdateConversionFactor` endpoint, make sure that the `authority` is either the gov module account (current behavior) **or** one of these oracle addresses.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2575, opened 2025-12-17
Allow oracles to update the flatfees conversion factor

nagarajdivine:
## Description

* Allow oracles to update the flatfees conversion factor.
closes: #2550 

---

write a little note why.

## Summary by CodeRabbit

* **New Features**
  * Oracle addresses can now be configured to adjust the flatfees conversion factor without requiring governance approval.
  * Added CLI commands to add and remove oracle addresses for managing trusted oracles.
  * Extended API with new oracle address management endpoints and corresponding governance operations.

### DIFF
--- a/x/flatfees/keeper/msg_server_test.go
+++ b/x/flatfees/keeper/msg_server_test.go
@@ -145,6 +145,7 @@ func (k *MockKeeper) WithExpSetConversionFactor(conversionFactors ...types.Conve
 	k.SetConversionFactorExp = append(k.SetConversionFactorExp, conversionFactors...)
 	return k
 }
+
 func (k *MockKeeper) WithIsOracleAddressResults(results ...bool) *MockKeeper {
 	k.IsOracleAddressResp = append(k.IsOracleAddressResp, results...)
 	return k


x/flatfees/keeper/genesis_test.go
```diff
@@ -57,6 +57,7 @@ func (s *GenesisTestSuite) TestInitExportGenesis() {
 						DefinitionAmount: sdk.NewInt64Coin("apple", 7),
 						ConvertedAmount:  sdk.NewInt64Coin("orange", 3),
 					},
+					OracleAddresses: nil,
```
SpicyLemon: Please add some addresses here. This test is named "non-defaults" yet `nil` is the default for this field.

x/flatfees/keeper/keeper.go
```diff
@@ -268,3 +268,75 @@ func (k Keeper) SetConversionFactor(ctx sdk.Context, conversionFactor types.Conv
...
+	}
+
+	params := k.GetParams(ctx)
+
+	if params.IsOracleAddress(address) {
+		return fmt.Errorf(
+			"%s: address: %s",
+			types.ErrOracleAlreadyExists.Error(),
+			address,
+		)
```
SpicyLemon: This should be one line. Same with the similar ones above and below.

x/flatfees/keeper/keeper.go
```diff
@@ -268,3 +268,75 @@ func (k Keeper) SetConversionFactor(ctx sdk.Context, conversionFactor types.Conv
...
+func (k Keeper) RemoveOracleAddress(ctx sdk.Context, address string) error {
+	params := k.GetParams(ctx)
+
+	if len(params.OracleAddresses) == 0 {
+		return fmt.Errorf(
+			"%s: address: %s",
+			types.ErrOracleNotFound.Error(),
+			address,
+		)
+	}
```
SpicyLemon: Let's take this part out. I know it saves a little work if the list is already empty, but it's going to be really rare that this work is done and it would be better to only emit this error from one place.

x/flatfees/keeper/keeper.go
```diff
@@ -268,3 +268,75 @@ func (k Keeper) SetConversionFactor(ctx sdk.Context, conversionFactor types.Conv
...
+	params := k.GetParams(ctx)
+	return params.IsOracleAddress(address)
+}
+
+// AddOracleAddress adds an oracle address to the params
+func (k Keeper) AddOracleAddress(ctx sdk.Context, address string) error {
+	if _, err := sdk.AccAddressFromBech32(address); err != nil {
+		return fmt.Errorf(
+			"%s: invalid address format: %w",
+			types.ErrInvalidOracleAddr.Error(),
```
SpicyLemon: By only using the error string for these errors, you're negating the reason for making them their own error types. By using them like this, you lose the stack trace and error codes associated with them.

Let's swap this around and wrap your defined errors, and just include the contents of `err`.

E.g. `fmt.Errorf("%w: invalid address format: %v", types.ErrInvalidOracleAddr, err)`

This goes for all the errors in here like this.

x/flatfees/keeper/keeper.go
```diff
@@ -268,3 +268,75 @@ func (k Keeper) SetConversionFactor(ctx sdk.Context, conversionFactor types.Conv
...
+		return fmt.Errorf(
+			"%s: address: %s",
+			types.ErrOracleNotFound.Error(),
+			address,
+		)
+	}
+
+	if len(newOracles) == 0 {
+		newOracles = []string{}
+	}
```
SpicyLemon: This isn't needed.

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,335 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+		{
+			name:       "add invalid address format",
+			addAddress: "invalid",
+			expErr:     "invalid oracle address: invalid address format: decoding bech32 failed: invalid bech32 string length 7",
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+
```
SpicyLemon: Please stop putting a blank space at the top of code blocks.

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,335 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+				OracleAddresses:  []string{oracle1, "invalid"},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			s.Require().NotPanics(func() {
+				_ = s.kpr.SetParams(s.ctx, tc.params)
+			}, "SetParams(%+v)", tc.params)
```
SpicyLemon: Our `assertions` package (under `testutil`) has a `RequireNotPanicsNoError` assertion that would be good to use here.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -610,3 +681,501 @@ func (s *MsgServerTestSuite) TestUpdateMsgFees() {
...
+func (s *MsgServerTestSuite) TestAddOracleAddress() {
+	oracle1 := sdk.AccAddress("oracle1_____________").String()
+	oracle2 := sdk.AccAddress("oracle2_____________").String()
+
+	tests := []struct {
+		name    string
+		kpr     *MockKeeper
+		req     *types.MsgAddOracleAddressRequest
+		expErr  string
+		expCall bool
```
SpicyLemon: This field doesn't seem to be used. It gets set, but is never used.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -610,3 +681,501 @@ func (s *MsgServerTestSuite) TestUpdateMsgFees() {
...
+	oracle1 := sdk.AccAddress("oracle1_____________").String()
+	oracle2 := sdk.AccAddress("oracle2_____________").String()
+	nonOracle := sdk.AccAddress("non_oracle__________").String()
+
+	tests := []struct {
+		name    string
+		kpr     *MockKeeper
+		req     *types.MsgRemoveOracleAddressRequest
+		expErr  string
+		expCall bool
```
SpicyLemon: Same here, this field doesn't seem to be used.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -610,3 +681,501 @@ func (s *MsgServerTestSuite) TestUpdateMsgFees() {
...
+	}
+
+	tests := []struct {
+		name     string
+		kpr      *MockKeeper
+		requests []struct {
+			authority string
+			cf        types.ConversionFactor
+			isOracle  bool
+		}
```
SpicyLemon: Give this struct type a name so you don't have to repeat it in every single instance.

You can define types inside functions, that are only visible to those functions:

```
type testReq struct {
	authority string
	cf        types.ConversionFactor
	isOracle  bool
}

Comment (SpicyLemon) [reply x/flatfees/keeper/msg_server_test.go]: Also, add a field for the expected error so that the expected error is defined with the test case rather than afterwards in a slice of strings.

File: x/flatfees/keeper/query_server_test.go [hunk 11]
```diff
@@ -156,6 +156,7 @@ func (s *QueryServerTestSuite) TestParams() {
 					DefinitionAmount: sdk.NewInt64Coin("apple", 44),
 					ConvertedAmount:  sdk.NewInt64Coin("orange", 78),
 				},
+				OracleAddresses: nil,
```
SpicyLemon: Please define something here.

x/flatfees/types/errors.go
```diff
@@ -7,6 +7,8 @@ import (
 // The x/flatfees module sentinel errors.
 
 var (
-	// ErrMsgFeeDoesNotExist is returned when a message fee entry does not exist for a given message type.
-	ErrMsgFeeDoesNotExist = cerrs.Register(ModuleName, 5, "fee for type does not exist")
+	ErrMsgFeeDoesNotExist  = cerrs.Register(ModuleName, 5, "fee for type does not exist")
+	ErrOracleAlreadyExists = cerrs.Register(ModuleName, 1100, "oracle address already exists")
+	ErrOracleNotFound      = cerrs.Register(ModuleName, 1101, "oracle address not found")
+	ErrInvalidOracleAddr   = cerrs.Register(ModuleName, 1102, "invalid oracle address")
```
SpicyLemon: Codes are unique to a namespace. The namespace in this instance is the `ModuleName` variable, which is just `flatfees`.

The `ErrMsgFeeDoesNotExist` was left at `5` since that's what it was for that error in the old `msgfees` module. 

For these new ones, it's fine to just make them 6, 7, and 8.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +43,50 @@ func (p Params) Validate() error {
...
+	}
+
+	seen := make(map[string]bool)
+	for _, addr := range addresses {
+		if addr == "" {
+			return errors.New("oracle address cannot be empty")
+		}
+
+		if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+			return fmt.Errorf("invalid oracle address format %q: %w", addr, err)
```
SpicyLemon: I'd remove `format ` from this error message so that it's just `"invalid oracle address %q: %w"`.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +43,50 @@ func (p Params) Validate() error {
...
+	if len(addresses) == 0 {
+		return nil
+	}
+
+	seen := make(map[string]bool)
+	for _, addr := range addresses {
+		if addr == "" {
+			return errors.New("oracle address cannot be empty")
+		}
+
```
SpicyLemon: There's no need to do this check on its own since `AccAddressFromBech32` also does it.

x/flatfees/types/flatfees.go
SpicyLemon: We should return all possible errors, not just the first one.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +43,50 @@ func (p Params) Validate() error {
...
+		if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+			return fmt.Errorf("invalid oracle address format %q: %w", addr, err)
+		}
+
+		if seen[addr] {
+			return fmt.Errorf("duplicate oracle address: %q", addr)
+		}
+		seen[addr] = true
```
SpicyLemon: Put this check at the start of the loop.

Even though we want to return all errors, if someone puts the same invalid address in the list twice, we should get one error first about the invalid address, then another error should get included about the duplication.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +43,50 @@ func (p Params) Validate() error {
...
+func (p Params) IsOracleAddress(address string) bool {
+	if p.OracleAddresses == nil {
+		return false
+	}
+	for _, oracle := range p.OracleAddresses {
+		if oracle == address {
+			return true
+		}
+	}
+	return false
```
SpicyLemon: You should be able simplify this with something like `return slices.Contains(p.OracleAddresses, address)`.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +43,50 @@ func (p Params) Validate() error {
...
+	}
+
+	return nil
+}
+
+// IsOracleAddress checks if the given address is in the oracle list
+func (p Params) IsOracleAddress(address string) bool {
+	if p.OracleAddresses == nil {
+		return false
+	}
```
SpicyLemon: This isn't needed. If `p.OracleAddresses` is nil (or empty), the loop below it simply doesn't iterate over anything.

x/flatfees/types/genesis.go
```diff
@@ -67,5 +67,12 @@ func GetGenesisStateFromAppState(cdc codec.JSONCodec, appState map[string]json.R
 		}
 	}
 
+	if genesisState.Params.OracleAddresses == nil {
+		genesisState.Params.OracleAddresses = []string{}
+	}
+	if genesisState.MsgFees == nil {
+		genesisState.MsgFees = []*MsgFee{}
+	}
+
```
SpicyLemon: Why is this needed?

I ask because the comment on this method states that "If the appState doesn't have anything for this module, a zero-value (empty) GenesisState is returned." But assigning empty slices makes it not the zero-value anymore.

x/flatfees/types/genesis_test.go
```diff
@@ -146,6 +146,7 @@ func TestGetGenesisStateFromAppState(t *testing.T) {
 				DefinitionAmount: sdk.NewInt64Coin("apple", 25),
 				ConvertedAmount:  sdk.NewInt64Coin("banana", 3),
 			},
+			OracleAddresses: []string{},
```
SpicyLemon: Please put some entries in here.

x/flatfees/types/msgs.go
```diff
@@ -72,3 +74,31 @@ func (m MsgUpdateMsgFeesRequest) ValidateBasic() error {
...
 }
+
+// ValidateBasic performs basic validation on MsgAddOracleAddressRequest
+func (m MsgAddOracleAddressRequest) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(m.Authority); err != nil {
+		return fmt.Errorf("invalid authority: %w", err)
+	}
+	if m.OracleAddress == "" {
+		return fmt.Errorf("oracle address cannot be empty")
+	}
```
SpicyLemon: You don't need this since `AccAddressFromBech32` does it for you. Same with below.

docs/proto-docs.md
```diff
@@ -531,6 +531,10 @@
     - [UpdateNhashPerUsdMilProposal](#provenance-msgfees-v1-UpdateNhashPerUsdMilProposal)
   
 - [provenance/flatfees/v1/tx.proto](#provenance_flatfees_v1_tx-proto)
+    - [MsgAddOracleAddressRequest](#provenance-flatfees-v1-MsgAddOracleAddressRequest)
+    - [MsgAddOracleAddressResponse](#provenance-flatfees-v1-MsgAddOracleAddressResponse)
+    - [MsgRemoveOracleAddressRequest](#provenance-flatfees-v1-MsgRemoveOracleAddressRequest)
+    - [MsgRemoveOracleAddressResponse](#provenance-flatfees-v1-MsgRemoveOracleAddressResponse)
```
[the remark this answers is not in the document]
SpicyLemon: Ignore this file. Always. And forever. It's auto-generated, so your comments on it are worthless and going to be ignored anyway.

Ignore this file for all PRs.

proto/provenance/flatfees/v1/tx.proto
```diff
@@ -67,4 +73,32 @@ message MsgUpdateMsgFeesRequest {
 }
 
 // MsgUpdateMsgFeesResponse is the response for the UpdateMsgFees governance endpoint.
-message MsgUpdateMsgFeesResponse {}
\ No newline at end of file
+message MsgUpdateMsgFeesResponse {}
+
```
SpicyLemon, line 77: Please update the comment in `MsgUpdateConversionFactorRequest` for the `authority` field to indicate that an oracle address is also valid. Something like "authority should be either the governance module account address or an oracle address."

x/flatfees/types/genesis.go
nagarajdivine: I added this because `TestGetGenesisStateFromAppState` was failing in the `empty app state`, `empty flatfees state`, and `unmarshal error cases`, as those tests expect non-nil slices.

SpicyLemon: Please get rid of this here, and fix the unit tests accordingly.

The `UnmarshalJSON` function creates empty slices like this. So you should only expect these empty slices when the app state has something for the module.

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,332 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+		},
+		{
+			name:       "add invalid address format",
+			addAddress: "invalid",
+			expErr:     "invalid oracle address: invalid address format: decoding bech32 failed: invalid bech32 string length 7",
+		},
+	}
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+
```
SpicyLemon: Please stop putting blank lines at the start of code blocks.

Usually there's a blank line between the definition of the test cases and the `for` loop iterating over them, though.

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,332 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+				if params.OracleAddresses == nil {
+					params.OracleAddresses = []string{}
+				}
+				s.Assert().Equal(tc.expOracles, params.OracleAddresses, "oracle addresses after remove")
+				s.Assert().False(s.kpr.IsOracleAddress(s.ctx, tc.removeAddress), "removed address should not be oracle")
+			}
+		})
+	}
+}
+func (s *KeeperTestSuite) TestKeeper_SetParams_WithOracleAddresses() {
```
SpicyLemon, line 1343: There should be a blank line between these two.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -143,6 +208,8 @@ func shiftErr(errs []string) ([]string, error) {
 		switch {
 		case errMsg == "ErrMsgFeeDoesNotExist":
 			err = types.ErrMsgFeeDoesNotExist
+		case errMsg == "not gov":
+			err = sdkerrors.ErrUnauthorized
```
SpicyLemon, line 214: Can we stick with the same pattern here, and make `ErrUnauthorized` the magic string (instead of "not gov")?

It helps make individual test cases easier to understand.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +45,49 @@ func (p Params) Validate() error {
...
+		}
+
+		if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+			allErrors = append(allErrors, fmt.Sprintf("invalid oracle address %q: %v", addr, err))
+			continue
+		}
+
+		if seen[addr] > 1 {
+			allErrors = append(allErrors, fmt.Sprintf("duplicate oracle address: %q", addr))
+		}
```
SpicyLemon: Put this check at the start of the loop.

If the list has the same invalid address listed twice, we want the first one to get an invalid address, and the 2nd one to get a duplicate error. But the order these checks are in means that there'll be two invalid address errors.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +45,49 @@ func (p Params) Validate() error {
...
 	return nil
 }
 
+// validateOracleAddresses validates the oracle addresses list
+func validateOracleAddresses(addresses []string) error {
+	if len(addresses) == 0 {
+		return nil
+	}
+
+	var allErrors []string
```
SpicyLemon: Please make this a `[]error`.

Then, at the end, you just do `return errors.Join(allErrors...)`.

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,335 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+		{
+			name:       "add invalid address format",
+			addAddress: "invalid",
+			expErr:     "invalid oracle address: invalid address format: decoding bech32 failed: invalid bech32 string length 7",
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+
```
SpicyLemon: Seriously. Delete this empty line.

```suggestion
```

x/flatfees/keeper/keeper_test.go
```diff
@@ -1128,3 +1128,335 @@ func (s *KeeperTestSuite) TestKeeper_SetConversionFactor() {
...
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			assertions.RequireNotPanicsNoErrorf(s.T(), func() error {
+				return s.kpr.SetParams(s.ctx, tc.params)
+			}, "SetParams(%+v)", tc.params)
+
+			act := s.kpr.GetParams(s.ctx)
+			if act.OracleAddresses == nil {
+				act.OracleAddresses = []string{}
+			}
```
SpicyLemon: You shouldn't be altering the actual, you should be setting your expected accordingly.

x/flatfees/keeper/msg_server_test.go
SpicyLemon: Looks like I accidentally messed these checks up, but now's a good time to fix them.

Instead of comparing `k.SetParamsExp` to itself, the `actual` here should be `k.SetParamsArgs` (in both assertions here.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -610,3 +681,480 @@ func (s *MsgServerTestSuite) TestUpdateMsgFees() {
 		})
 	}
 }
+
+// TestUpdateConversionFactor_WithOracles tests the dual authorization (governance OR oracle)
+func (s *MsgServerTestSuite) TestUpdateConversionFactor_WithOracles() {
```
SpicyLemon: These tests should all be part of the existing `TestUpdateConversionFactor` instead of on their own.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +44,43 @@ func (p Params) Validate() error {
...
+	if len(addresses) == 0 {
+		return nil
+	}
+
+	var allErrors []error
+	seen := make(map[string]int)
+
+	for _, addr := range addresses {
+		if seen[addr] > 0 {
+			allErrors = append(allErrors, fmt.Errorf("duplicate oracle address: %q", addr))
```
SpicyLemon, line 64:
```suggestion
			allErrors = append(allErrors, fmt.Errorf("duplicate oracle address: %q", addr))
			continue
```

If it's a duplicate and invalid, we only want the duplicate error here because we already have an error stating that it's invalid.

x/flatfees/types/flatfees.go
```diff
@@ -44,9 +44,43 @@ func (p Params) Validate() error {
...
+
+		if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+			allErrors = append(allErrors, fmt.Errorf("invalid oracle address %q: %w", addr, err))
+		}
+	}
+
+	if len(allErrors) > 0 {
+		return errors.Join(allErrors...)
+	}
 	return nil
```
SpicyLemon:
```suggestion
	return errors.Join(allErrors...)
```

The `errors.Join` function returns `nil` if no errors are provided to it (or if all of the error entries provided to it are `nil`). But you don't have to check the length on your own.

x/flatfees/types/genesis_test.go
```diff
@@ -131,21 +131,121 @@ func TestDefaultGenesisState(t *testing.T) {
 	assert.Empty(t, genState.MsgFees, "genState.MsgFees")
 }
 
+// func TestGetGenesisStateFromAppState(t *testing.T) {
```
SpicyLemon: This test should be uncommented (and fixed) or deleted.

x/flatfees/types/genesis_test.go
```diff
@@ -203,21 +306,19 @@ func TestGetGenesisStateFromAppState(t *testing.T) {
...
 			testFunc := func() {
-				actState, err = GetGenesisStateFromAppState(tc.cdc, tc.appState)
+				actState, err = GetGenesisStateFromAppState(cdc, tc.appState)
 			}
 			require.NotPanics(t, testFunc, "GetGenesisStateFromAppState")
 			assertions.AssertErrorValue(t, err, tc.expErr, "GetGenesisStateFromAppState error")
 			assert.Equal(t, tc.expState, actState, "GetGenesisStateFromAppState result")
-			if t.Failed() && tc.appState != nil {
-				t.Logf("%s json:\n%s", ModuleName, string(tc.appState[ModuleName]))
-			}
```
SpicyLemon, line 220: Please put this back. Stuff like this helps you when this test starts to fail later on in life.

x/flatfees/keeper/msg_server.go
```diff
@@ -48,11 +52,23 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+	if err := m.ValidateAuthority(req.Authority); err == nil {
+		isGov = true
 	}
 
-	err := m.SetConversionFactor(sdk.UnwrapSDKContext(goCtx), req.ConversionFactor)
+	isOracle := m.IsOracleAddress(ctx, req.Authority)
+
+	if !isGov && !isOracle {
+		return nil, govtypes.ErrInvalidSigner.Wrapf(
+			"expected governance authority or an oracle address, got %s",
```
SpicyLemon: This needs to be `%q` here (instead of `%s`), specially since it's a raw value provided by the user.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -610,3 +681,480 @@ func (s *MsgServerTestSuite) TestUpdateMsgFees() {
...
+			assertions.AssertErrorValue(s.T(), err, tc.expErr, "RemoveOracleAddress error")
+			s.Assert().Equal(expResp, actResp, "RemoveOracleAddress response")
+
+			tc.kpr.AssertCalls(s.T())
+		})
+	}
+}
+
+// TestUpdateConversionFactor_MultipleUpdates tests multiple sequential updates
+func (s *MsgServerTestSuite) TestUpdateConversionFactor_MultipleUpdates() {
```
SpicyLemon: Please relocate this test in this file so that it is directly after the `TestUpdateConversionFactor` test. This keeps all the tests for that endpoint together.

x/flatfees/keeper/msg_server_test.go
```diff
@@ -132,6 +145,59 @@ func (k *MockKeeper) WithExpSetConversionFactor(conversionFactors ...types.Conve
 	k.SetConversionFactorExp = append(k.SetConversionFactorExp, conversionFactors...)
 	return k
 }
+func (k *MockKeeper) WithIsOracleAddressResults(results ...bool) *MockKeeper {
```
SpicyLemon, line 149:
```suggestion
}

func (k *MockKeeper) WithIsOracleAddressResults(results ...bool) *MockKeeper {
```

Issue #2585, opened 2026-01-15
Set up the circuit breaker.


SpicyLemon:
## Summary

Set up the `x/circuit` module.

## Problem Definition

We already have the `x/circuit` module included in our chain, but we don't have any addresses set up to manage it.

## Proposal

Add a process to the upgrades that sets gives the Provenance Foundation account(s) `SUPER_ADMIN` access and gives a handful of us `ALL_MSGS` access (with the circuit breaker). These account addresses will be added to this issue later.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2598, opened 2026-01-22
Add circuit breaker permission setup for upgrades.

nagarajdivine:
## Description

* Add circuit breaker permission setup for upgrades.
closes: #2585 

---

write a little note why.

## Summary by CodeRabbit

* **New Features**
  * Circuit breaker admin permissions are now automatically configured during upgrades, granting elevated access to designated foundation accounts and operational access to team accounts for consistent security across networks.

* **Tests**
  * Added tests for the permission setup covering full success, partial failures with invalid addresses, and no-op empty-input cases; tests validate expected log sequences including per-address outcomes.

* **Documentation**
  * Added an unreleased changelog entry documenting the upgrade behavior.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -411,60 +411,6 @@ var (
 	_ = unlockVestingAccounts
 )
 
-// setupCircuitBreakerPermissions grants circuit breaker permissions
-// during an upgrade using the caller-provided address lists.
-// The upgrade handlers are responsible for selecting the correct
-// mainnet/testnet address sets.
-func setupCircuitBreakerPermissions(ctx sdk.Context, app *App, foundationAccounts, teamAccounts []string) {
-	ctx.Logger().Info("Setting up circuit breaker permissions.")
-
-	grant := func(addresses []string, level circuittypes.Permissions_Level) {
-		for i, addrStr := range addresses {
-			addr, err := sdk.AccAddressFromBech32(addrStr)
-			if err != nil {
-				ctx.Logger().Error(
-					fmt.Sprintf(
-						"Invalid address at index %d (%s): decoding bech32 failed. Skipping.",
-						i, addrStr,
-					),
-				)
-				continue
-			}
-
-			perms := circuittypes.Permissions{
-				Level:         level,
-				LimitTypeUrls: []string{},
-			}
-
-			if err := app.CircuitKeeper.Permissions.Set(ctx, addr, perms); err != nil {
-				ctx.Logger().Error(
-					fmt.Sprintf("Failed to grant permissions to %s. Skipping.", addrStr),
-				)
-				continue
-			}
-
-			ctx.Logger().Info(fmt.Sprintf("Granted Level %s to %s.", level, addrStr))
-		}
-	}
-	// Apply Foundation Permissions (SUPER_ADMIN)
-	grant(foundationAccounts, circuittypes.Permissions_LEVEL_SUPER_ADMIN)
-
-	// Apply Team Permissions (ALL_MSGS)
-	grant(teamAccounts, circuittypes.Permissions_LEVEL_ALL_MSGS)
-
-	ctx.Logger().Info("Circuit breaker setup configured.")
-}
-
-// Create a use of the standard helpers so that the linter neither complains about it not being used,
-// nor complains about a nolint:unused directive that isn't needed because the function is used.
-var (
-	_ = runModuleMigrations
-	_ = removeInactiveValidatorDelegations
-	_ = pruneIBCExpiredConsensusStates
-	_ = convertFinishedVestingAccountsToBase
-	_ = unlockVestingAccounts
-)
-
 // setupMsgStoreCodeFee sets the flat fee for MsgStoreCode to $100 (100,000 musd).
 // This allows smart contracts to be stored directly without governance proposals.
 func setupMsgStoreCodeFee(ctx sdk.Context, app *App) error {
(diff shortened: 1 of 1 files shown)


app/upgrades.go
```diff
@@ -368,6 +376,87 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
...
+func setupCircuitBreakerPermissions(ctx sdk.Context, app *App) error {
+	ctx.Logger().Info("Setting up circuit breaker permissions")
+
+	chainID := ctx.ChainID()
+
+	var foundationAccounts []string
+	var teamAccounts []string
+
+	switch {
+	case strings.Contains(chainID, "mainnet") || chainID == "pio-1":
```
SpicyLemon: The `daisy-rc1` upgrade is specifically for testnet, and `daisy` is for mainnet.

Instead of doing this detection here, the calls should be slightly different from each upgrade.

We often have methods like, `GetMainnetData` or `GetTestnetAddresses` to differentiate the data for each environment. These are then either called directly in the upgrade handler, or passed as a function reference from the upgrade handler.

In this case, you'd probably just pass in each list, e.g. `setupCircuitBreakerPermissions(ctx sdk.Context, app *App, superUserAccounts, normalUserAccounts []string)` Then, you'd have the two different functions: `GetMainnetCBAddrs() (superUserAccounts, normalUserAccounts []string)` and `GetTestnetCBAddrs() (superUserAccounts, normalUserAccounts []string)`.

app/upgrades.go
```diff
@@ -85,6 +87,9 @@ var upgrades = map[string]appUpgrade{
 			if vm, err = runModuleMigrations(ctx, app, vm); err != nil {
 				return nil, err
 			}
+			if err = setupCircuitBreakerPermissions(ctx, app); err != nil {
+				return nil, err
+			}
```
SpicyLemon: Since this setup isn't crucial to the running of the chain, we don't want an error to interrupt the upgrade (which halts the chain). So, all errors encountered in `setupCircuitBreakerPermissions` should just be logged instead of returned.

app/upgrades.go
```diff
@@ -368,6 +376,87 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
...
+			chainID,
+		)
+	}
+
+	// Validate and Grant Permissions.
+	grant := func(addresses []string, level circuittypes.Permissions_Level) error {
+		for i, addrStr := range addresses {
+			addr, err := sdk.AccAddressFromBech32(addrStr)
+			if err != nil {
+				return fmt.Errorf("invalid address at index %d (%s): %w", i, addrStr, err)
```
SpicyLemon: This should just be logged, then it should move on to the next address.
Please also include the permission this is for.

app/upgrades.go
```diff
@@ -368,6 +376,87 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
...
+			}
+
+			perms := circuittypes.Permissions{
+				Level:         level,
+				LimitTypeUrls: []string{},
+			}
+
+			if err := app.CircuitKeeper.Permissions.Set(ctx, addr, perms); err != nil {
+				return fmt.Errorf("failed to grant permissions to %s: %w", addrStr, err)
```
SpicyLemon: This should just be logged, then move on to the next address.
Please also include which permission this was for and the index of the addr.

app/upgrades.go
```diff
@@ -368,6 +376,87 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
...
 }
 
+// setupCircuitBreakerPermissions grants circuit breaker admin permissions
+// during an upgrade based on the chain ID.
+//
+// Mainnet and testnet require explicit handling for safety.
+// Local/dev chains (including empty chain IDs used in tests) skip setup.
+// Unknown non-local chain IDs fail fast to prevent unsafe launches.
+func setupCircuitBreakerPermissions(ctx sdk.Context, app *App) error {
+	ctx.Logger().Info("Setting up circuit breaker permissions")
```
SpicyLemon: Please include punctuation in log statements.

app/upgrades_test.go
```diff
@@ -1186,3 +1186,115 @@ func (s *UpgradeTestSuite) TestExecuteStoreCodeMsg() {
...
+				LogMsgCircuitMainnet,
+				LogMsgCircuitTestnet,
+				"Skipping",
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+
```
SpicyLemon, line 1246: There shouldn't be a blank line here.

app/upgrades_test.go
```diff
@@ -1186,3 +1186,115 @@ func (s *UpgradeTestSuite) TestExecuteStoreCodeMsg() {
 		})
 	}
 }
+func (s *UpgradeTestSuite) TestSetupCircuitBreakerPermissions() {
+
```
SpicyLemon: There shouldn't be a blank line here.

app/upgrades.go
```diff
@@ -368,6 +377,74 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
 		resp.CodeID, fmt.Sprintf("%x", resp.Checksum)))
 }
 
+// getTestnetCircuitBreakerAddrs returns the list of accounts for Testnet.
```
SpicyLemon: Please move all of these additions to the bottom of the file.

app/upgrades.go
```diff
@@ -368,6 +377,50 @@ func executeStoreCodeMsg(ctx sdk.Context, wasmMsgServer wasmMsgSrvr, msg *wasmty
 		resp.CodeID, fmt.Sprintf("%x", resp.Checksum)))
 }
 
+// setupCircuitBreakerPermissions grants circuit breaker permissions
```
SpicyLemon: Please move this to the bottom too.

The way this file is ordered is this:

1) Upgrades (oldest first).
2) Standard stuff used in most upgrades.
3) A `var` block where we provide fake uses of the stuff used in most upgrades.
4) Stuff for specific upgrades (oldest first).

This should be defined either directly before `getTestnetCircuitBreakerAddrs` or directly after `getMainnetCircuitBreakerAddrs`. But it doesn't belong in this area of the file.

app/upgrades_test.go
```diff
@@ -1186,3 +1186,70 @@ func (s *UpgradeTestSuite) TestExecuteStoreCodeMsg() {
 		})
 	}
 }
+
```
SpicyLemon, line 1192: Please add an expected log line to both `TestDaisyRC1` and `TestDaisy`.

Since this is a 1-time thing, don't create a variable with the expected log content, just put the expected strings in the lists.

app/upgrades.go
```diff
@@ -87,6 +88,10 @@ var upgrades = map[string]appUpgrade{
 			if vm, err = runModuleMigrations(ctx, app, vm); err != nil {
 				return nil, err
 			}
+
+			foundation, team := getTestnetCircuitBreakerAddrs()
+			setupCircuitBreakerPermissions(ctx, app, foundation, team)
+
```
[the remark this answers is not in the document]
SpicyLemon: I don't know what you're doing wrong, but `getTestnetCircuitBreakerAddrs` and `getMainnetCircuitBreakerAddrs` very much exist and are even defined in this same file.

app/upgrades.go
```diff
@@ -402,6 +411,60 @@ var (
...
+		}
+	}
+	// Apply Foundation Permissions (SUPER_ADMIN)
+	grant(foundationAccounts, circuittypes.Permissions_LEVEL_SUPER_ADMIN)
+
+	// Apply Team Permissions (ALL_MSGS)
+	grant(teamAccounts, circuittypes.Permissions_LEVEL_ALL_MSGS)
+
+	ctx.Logger().Info("Circuit breaker setup configured.")
+}
```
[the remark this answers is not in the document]
SpicyLemon: No. We specifically do NOT want to fail fast when one fails. That way, it'll still move on to the next one and we have less to fix later.

Issue #2593, opened 2026-01-16
Update Sims to match v0.53.x style


SpicyLemon:
## Summary

Update the sims to match what is done in v0.53.x of the SDK.

## Problem Definition

With `v0.53.x` of the SDK, they overhauled how the sims run. To begin, they no longer use the `runsim` external tool. We should update our sims to run the same way.

## Proposal

Overhaul our sim tests to make them match what the SDK does.
Make sure they all pass. This shouldn't require any changes to specific module simulation code since they all pass right now.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SpicyLemon:
We'll want to wait on this until we get the SDK bumped to `v0.53.x`.

SpicyLemon:
We'll also need to fix `runsim (import-export, goleveldb, ubuntu-latest)` i.e. `make test-sim-import-export`.

Currently, several seeds fail from either the attribute or wasm modules.

Seeds `1`, `3232`, `7601778` fail in the attribute module with an error like this:
`panic: unexpected attribute key 03C77E39831277885E2F7A28D0D414CB8F8A322A4E6411351C77713AEE4E2B81EA14B8ED70194ECCE83581E15301CECD795D9A8F4C15`
That `03` at the start indicates that it's an attribute address lookup key. with the format: `<type byte 03 (one byte)><name sha256 (32 bytes)><address length (1 byte)><address>` (note that hex `14` is equal to decimal 20 indicating that the address has 20 bytes). The name and address is different for each seed (but consistent for each seed).

Seeds `32`, `2989`, `3012`, `37827`, `29892989` fail in the wasm module with an error like this:
```
compared 2 different key/value pairs between KVStoreKey{0x14003717250, wasm} and KVStoreKey{0x14002ece580, wasm}
    sim_test.go:448:
        	Error Trace:	/Users/dannywedul/git/provenance/app/sim_test.go:448
        	Error:      	Not equal:
        	            	expected: 0
        	            	actual  : 2
        	Test:       	TestAppImportExport/wasm
[...]

PR #2645, opened 2026-03-11
update Sims to match v0.53.x style.

nagarajdivine:
## Description

* update Sims to match v0.53.x style.
closes: #2593 

---

write a little note why.

## Summary by CodeRabbit

* **Bug Fixes**
  * Genesis export now validates attribute expirations and omits invalid entries
  * Attribute lookup counters fixed to avoid double-counting on updates

* **Improvements**
  * Sims updated to v0.53.x style with expanded scenarios and stronger nondeterminism checks
  * Decoder and attribute handling made more tolerant and robust
  * CI/test logging streamlined and test matrix broadened

* **Tests**
  * Simulation decoder tests extended to cover new key types and params

* **Chores**
  * Removed legacy simulation runner tooling and simplified build/test targets

### DIFF
--- a/sims.mk
+++ b/sims.mk
@@ -79,7 +79,7 @@ test-sim-simple:
 test-sim-benchmark-invariants:
 	@echo "Running simulation invariant benchmarks..."
 	$(GO) test -mod=readonly -tags sims -run=^$$ $(SIMAPP) -benchmem -bench=BenchmarkInvariants \
-		-NumBlocks=1000 -BlockSize=200 \
+		-NumBlocks=450 -BlockSize=200 \
 		-Period=1 -Commit=true $(SEED_ARG) -v -timeout 24h
 
 SIM_NUM_BLOCKS ?= 500
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -140,7 +140,7 @@ func mutateAppState[G proto.Message](appState json.RawMessage, cdc codec.JSONCod
 func setupSimulation(dirPrefix string, dbName string) (simtypes.Config, dbm.DB, string, log.Logger, bool, error) {
 	config := simcli.NewConfigFromFlags()
 	config.ChainID = pioconfig.SimAppChainID
-	db, dir, logger, skip, err := simtestutil.SetupSimulation(config, dirPrefix, dbName, simcli.FlagVerboseValue, simcli.FlagEnabledValue)
+	db, dir, logger, skip, err := simtestutil.SetupSimulation(config, dirPrefix, dbName, simcli.FlagVerboseValue, true)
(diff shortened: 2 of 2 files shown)


sims.mk
```diff
@@ -22,73 +22,80 @@ ifneq ($(DB_BACKEND),goleveldb)
   $(error unknown DB_BACKEND value [$(DB_BACKEND)]. Must be goleveldb)
 endif
```
SpicyLemon: We don't need the `DB_BACKEND` variable anymore in here since that's no longer an option for an application. You can remove it this chunk and the argument from each of the targets below.

sims.mk
```diff
@@ -22,73 +22,80 @@ ifneq ($(DB_BACKEND),goleveldb)
...
-# flags: [-Jobs maxprocs] [-ExitOnFail] [-Seeds comma-separated-seed-list]
-#        [-Genesis file-path] [-SimAppPkg file-path] [-Github] [-Slack] [-LogObjPrefix string]
-
-test-sim-import-export: runsim
+test-sim-import-export:
 	@echo "Running application import/export simulation. This may take several minutes..."
-	$(BINDIR)/runsim -Jobs=4 -SimAppPkg='$(SIMAPP)' -ExitOnFail 30 3 'TestAppImportExport'
+	$(GO) test -mod=readonly -tags sims $(SIMAPP) -run TestAppImportExport \
+		-NumBlocks=30 -BlockSize=100 -DBBackend=$(DB_BACKEND) \
+		-Commit=true -Seed=57 -Period=3 -v -timeout 1h
```
SpicyLemon: These should not have a hard-coded seed value. However, it will make things easier down the road if we can provide the seed with the make command so that we can run a specific seed at will (e.g. when the sim fails in a github runner and we want to run it locallally).

Ideally, the command would look something like this `SEED=57 make test-sim-import-export`. To do this, we'll create a `SEED_ARG` variable that is either an empty string, or has `-Seed=$(SEED)` in it. Then, in these targets, include it in the command (unquoted), e.g. just `$(SEED_ARG)`. `SEED_ARG` should be an empty string unless `SEED` was provided with the `make` command.

All of these targets should have that `SEED_ARG` in it.

sims.mk
```diff
@@ -6,7 +6,6 @@
 ###
 ### Environment Variables:
 ###   GO:             The command to use to execute go. Default: go
-###   DB_BACKEND:     Dictates which db backend to use: goleveldb.
 ###                   The test-sim-nondeterminism is hard-coded to use memdb though.
```
SpicyLemon: This line should also be deleted.

Also, the new `SEED` env var should have an entry in this list.

Issue #2648, opened 2026-03-18
Fix admin check in CreateSecuritization


nagarajdivine:
## Summary of Bug
Ensure CreateSecuritization only allows admins to update marker permissions.

## Problem Definition
CreateSecuritization currently updates marker (pool) permissions without verifying that the signer has admin access. This allows unauthorized users to modify permissions.

## Proposal

- Add a validation step in CreateSecuritization to check that the signer has admin role for all involved markers before updating permissions.
- Reject the request if the signer is not an admin.
- Include unit tests to verify both valid and invalid cases.

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2649, opened 2026-03-18
Fix admin check in CreateSecuritization.

nagarajdivine:
## Description

* Fix admin check in CreateSecuritization.
closes: #2648
---

write a little note why.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced per-pool Admin authorization during securitization creation; prevents unauthorized operations and ensures permissions are checked for each pool.
  * Adjusted unauthorized error text formatting for clearer messages.

* **Tests**
  * Added tests verifying per-pool admin checks and that failed requests do not mutate pool access.

* **Documentation**
  * Added changelog entry documenting the admin-check fix.

### DIFF
--- a/.changelog/unreleased/improvements/2586-swagger-file.md
+++ b/.changelog/unreleased/improvements/2586-swagger-file.md
@@ -1,0 +1,1 @@
+* Update 3rd party swagger file [#2586](https://github.com/provenance-io/provenance/issues/2586).
--- a/client/docs/swagger_third_party.yaml
+++ b/client/docs/swagger_third_party.yaml
@@ -7,7 +7,6 @@ paths:
   /cosmos/auth/v1beta1/account_info/{address}:
     get:
       summary: AccountInfo queries account info which is common to all account types.
-      description: 'Since: cosmos-sdk 0.47'
       operationId: AccountInfo
       responses:
         '200':
@@ -24,7 +23,7 @@ paths:
                   pub_key:
                     type: object
                     properties:
-                      type_url:
+                      '@type':
                         type: string
                         description: >-
                           A URL/resource name that uniquely identifies the type
@@ -75,19 +74,17 @@ paths:
                           protobuf release, and it is not used for type URLs
                           beginning with
 
-                          type.googleapis.com.
+                          type.googleapis.com. As of May 2023, there are no
+                          widely used type server
+
+                          implementations and no plans to implement one.
 
 
                           Schemes other than `http`, `https` (or the empty
                           scheme) might be
 
                           used with implementation specific semantics.
-                      value:
-                        type: string
-                        format: byte
-                        description: >-
-                          Must be a valid serialized protocol buffer of the
-                          above specified type.
+                    additionalProperties: {}
                     description: >-
                       `Any` contains an arbitrary serialized protocol buffer
                       message along with a
@@ -125,7 +122,7 @@ paths:
                             foo = any.unpack(Foo.getDefaultInstance());
                           }
 
-                      Example 3: Pack and unpack a message in Python.
+                       Example 3: Pack and unpack a message in Python.
 
                           foo = Foo(...)
                           any = Any()
@@ -135,7 +132,7 @@ paths:
                             any.Unpack(foo)
                             ...
 
-                      Example 4: Pack and unpack a message in Go
+                       Example 4: Pack and unpack a message in Go
 
                            foo := &pb.Foo{...}
                            any, err := anypb.New(foo)
@@ -165,6 +162,7 @@ paths:
 
                       JSON
 
+                      ====
 
                       The JSON representation of an `Any` value uses the regular
 
@@ -207,17 +205,12 @@ paths:
                   sequence:
                     type: string
                     format: uint64
-            description: |-
-              QueryAccountInfoResponse is the Query/AccountInfo response type.
-
-              Since: cosmos-sdk 0.47
+            description: QueryAccountInfoResponse is the Query/AccountInfo response type.
         default:
           description: An unexpected error response.
           schema:
             type: object
             properties:
-              error:
-                type: string
               code:
                 type: integer
                 format: int32
@@ -228,7 +221,7 @@ paths:
                 items:
                   type: object
                   properties:
-                    type_url:
+                    '@type':
                       type: string
                       description: >-
                         A URL/resource name that uniquely identifies the type of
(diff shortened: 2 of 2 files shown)


x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
...
+		s.Require().NoError(err, "create asset %s", id)
+	}
+
+	// user1 creates two pools. user1 gets Admin on both.
+	for _, tc := range []struct {
+		denom  string
+		nft    string
+		amount int64
+	}{
+		{"pool-auth-a", "nft-auth-1", 500}, {"pool-auth-b", "nft-auth-2", 750}} {
```
SpicyLemon: Always separate the definition of the test cases and the loop over them.
Also, always name your fields when defining each entry.

I get that this is just setup stuff, but it's confusing to read because the loop code is in line with entry definitions.

x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
...
+				Id:       fmt.Sprintf("sec-auth-%d", counter),
+				Tranches: []*sdk.Coin{{Denom: "tranche-x", Amount: sdkmath.NewInt(100)}},
+				Pools:    tc.pools,
+				Signer:   tc.signer.String(),
+			}
+
+			_, err := msgServer.CreateSecuritization(s.ctx, msg)
+
+			if tc.expectErr {
+				s.Require().Error(err, "expected an error but got none")
```
SpicyLemon: Using a message of `"expected an error but got none"` is useless since the standard failure message states that already. Instead, you should use the `msgAndArgs` to tell you what you were doing that generated the error. In this case, the best thing to put there is just `"CreateSecuritization error"`. That way, when `make test` or github actions get a failure here, the failure message tells you where to start looking.

x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
...
+	s.Require().NoError(err)
+
+	// user1 tries to securitize [pool-per-1 (own), pool-per-2 (user2's)].
+	_, err = msgServer.CreateSecuritization(s.ctx, &types.MsgCreateSecuritization{
+		Id:       "sec-perpool-1",
+		Tranches: []*sdk.Coin{{Denom: "tr-pp", Amount: sdkmath.NewInt(50)}},
+		Pools:    []string{"pool-per-1", "pool-per-2"},
+		Signer:   s.user1Addr.String(),
+	})
+	s.Require().Error(err)
```
SpicyLemon: You should always be setting a `msgAndArgs` with every assertion. The `Error` assertion is one of the ones that needs it more than others. Without any context, all you're going to get is a failure stating that an error was expected, but didn't get one. That doesn't tell you anything about what went wrong or how you might start addressing it.

x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
 	s.Require().Error(err)
 	s.Require().Contains(err.Error(), "is not the owner of asset")
 }
+
+func (s *MsgServerTestSuite) TestCreateSecuritization_RequiresAdminOnEachPool() {
+	msgServer := keeper.NewMsgServerImpl(s.app.AssetKeeper)
+
+	// Create a second address that will NOT own any pool.
```
SpicyLemon: nit: this is the first address.

x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
...
+		signer      sdk.AccAddress
+		pools       []string
+		expectErr   bool
+		errContains string
+	}{
+		{
+			name:        "pool owner (admin) can securitize own pools",
+			signer:      s.user1Addr,
+			pools:       []string{"pool-auth-a", "pool-auth-b"},
+			expectErr:   false,
```
SpicyLemon: You don't ever need an `expectErr` boolean in a test definition. Wherever you'd use one, just check if the expected error is empty or not.

x/asset/keeper/msg_server_test.go
```diff
@@ -501,3 +503,153 @@ func (s *MsgServerTestSuite) TestBurnAsset_NotOwner() {
...
+			errContains: "",
+		},
+		{
+			name:        "address with no access on pool is rejected",
+			signer:      otherAddr,
+			pools:       []string{"pool-auth-a"},
+			expectErr:   true,
+			errContains: "does not have Admin access on pool marker",
```
SpicyLemon: You should also be making sure that the error message contains the offending address as well as the marker denom.

x/asset/keeper/msg_server.go
```diff
@@ -277,7 +281,11 @@ func (m msgServer) CreateSecuritization(goCtx context.Context, msg *types.MsgCre
 		if err != nil {
 			return nil, types.NewErrCodeNotFound(fmt.Sprintf("pool marker with denom %s", pool))
 		}
-
+		// Verify the signer holds Admin access on this pool marker before modifying
+		// any permissions.
+		if !poolMarker.AddressHasAccess(signerAddr, markertypes.Access_Admin) {
+			return nil, types.NewErrCodeUnauthorized(fmt.Sprintf("signer %q does not have Admin access on pool marker %q", msg.Signer, pool))
```
SpicyLemon: Please go into the `NewErrCodeUnauthorized` function and change the `%q` in there to `%s`.

You might have to update some unit tests because those errors are double-quoting things, so they'll have escaped double quotes that need to be cleaned up.

x/asset/keeper/msg_server.go
```diff
@@ -277,7 +281,11 @@ func (m msgServer) CreateSecuritization(goCtx context.Context, msg *types.MsgCre
...
+		}
```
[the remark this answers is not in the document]
SpicyLemon: As I've told you before (and you've already recognized), the code inside any `msgServer` function is automatically executed as a transaction. So if there's an error at any point, any changed state is automatically thrown out. So there's no need to do this in separate loops.

PR #2711, opened 2026-05-01
Make GetWhitelistedQuery return a copy of the query response

### LINKED ISSUE #2701: Make GetWhitelistedQuery return a copy of the query response.
## Description

Make the `GetWhitelistedQuery` function (in the internal stargate stuff) return a copy of the response object instead of the response object directly.

## Problem

The stargate whitelist stores proto.Message instances in a sync.Map during init(). These same instances are retrieved by GetWhitelistedQuery and then passed directly to ConvertProtoToJSONMarshal, which unmarshals response bytes into them, marshals to JSON, and then calls Reset().

The problem is that GetWhitelistedQuery returns a pointer to the shared instance stored in the map. Under concurrent stargate queries (multiple smart contracts querying simultaneously, or even a single contract in simulation + execution), two goroutines could call ConvertProtoToJSONMarshal with the same proto.Message pointer. One goroutine could Unmarshal data into the shared instance while another is MarshalJSON-ing it, producing corrupted JSON responses. The Reset() call exacerbates this—it could clear the message while another goroutine is marshalling it.

This is a data race that can produce incorrect query results returned to smart contracts. Incorrect data could cause contracts to make wrong decisions about transfers, approvals, or state changes. The sync.Map only protects the map operations themselves, not the shared proto.Message values stored within it.

## Proposed Solution

In `GetWhitelistedQuery`, instead of just returning `protoResponseType`, return `reflect.New(reflect.TypeOf(protoResponseType).Elem()).Interface().(proto.Message)`.

That will make it return a new zero-value object to store the response. So anything changed later isn't connected to the original.

____

## For Admin Use

- [ ] Not duplicate issue

nagarajdivine:
## Description

* Make GetWhitelistedQuery return a copy of the query response
closes: #2701 

---

write a little note why.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a data-race by ensuring query responses are returned as independent copies rather than shared instances.
* **Tests**
  * Added unit and concurrency tests confirming each query call yields a fresh, distinct response instance under sequential and concurrent use.
* **Documentation**
  * Added an unreleased changelog entry documenting the fix.

### DIFF
--- a/internal/provwasm/stargate_whitelist.go
+++ b/internal/provwasm/stargate_whitelist.go
@@ -2,6 +2,7 @@ package provwasm
 
 import (
 	"fmt"
+	"reflect"
 	"sync"
 
 	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
@@ -358,7 +359,17 @@ func GetWhitelistedQuery(queryPath string) (proto.Message, error) {
 	if !ok {
 		return nil, wasmvmtypes.Unknown{}
 	}
-	return protoResponseType, nil
+	// Build a fresh zero-value instance of the same concrete type.
+	rt := reflect.TypeOf(protoResponseType)
+	if rt.Kind() != reflect.Ptr {
+		return nil, wasmvmtypes.Unknown{}
+	}
+	freshAny := reflect.New(rt.Elem()).Interface()
+	fresh, ok := freshAny.(proto.Message)
+	if !ok {
+		return nil, wasmvmtypes.Unknown{}
+	}
+	return fresh, nil
 }
 
 func setWhitelistedQuery(queryPath string, protoType proto.Message) {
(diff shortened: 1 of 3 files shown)


internal/provwasm/stargate_whitelist.go
```diff
@@ -358,7 +359,17 @@ func GetWhitelistedQuery(queryPath string) (proto.Message, error) {
 	if !ok {
 		return nil, wasmvmtypes.Unknown{}
 	}
-	return protoResponseType, nil
+	// Build a fresh zero-value instance of the same concrete type.
```
SpicyLemon, line 362: While in here, can you please also fix the return type for the `"/cosmwasm.wasm.v1.Query/ContractHistory"`, query. It should return a `&wasmtypes.QueryContractHistoryResponse{}`.

SpicyLemon (approved):

PR #2714, opened 2026-05-05
fix metadata account-data REST query.

### LINKED ISSUE #2708: Metadata Account-data REST query doesn't work
## Summary of Bug

The metadata module's account-data REST query doesn't work.

## Version

v1.27

## Steps to Reproduce

`curl -X GET 'https://api.provenance.io/provenance/metadata/v1/accountdata/scope1qp4dzfye0pcy2g5qwt5gvf9wkyasvkmn0u'`

That returns this error:
`{"code":13, "message":"grpc: error unmarshalling request: invalid metadata address type: 177", "details":[]}%`

To contrast, here's the same query via command line:
```console
> provenanced q metadata account-data scope1qp4dzfye0pcy2g5qwt5gvf9wkyasvkmn0u
{"value":""}
```

## Extra Details

The problem is that the `metadata_addr` field in the `AccountDataRequest` is a `bytes` with the `(gogoproto.customtype) = "MetadataAddress"` option.

That causes the query to use [MetadataAddress#Unmarshal](https://github.com/provenance-io/provenance/blob/main/x/metadata/types/address.go#L322-L331) method, which calls [VerifyMetadataAddressFormat](https://github.com/provenance-io/provenance/blob/main/x/metadata/types/address.go#L47-L94). When a scope id is provided, it's treated as a raw byte array. As a byte array, the first byte is the metadata address type and needs to be 0 for a scope id. But it's not in this case, so we get the error returned from [this line](https://github.com/provenance-io/provenance/blob/main/x/metadata/types/address.go#L79).

The command line version works because it takes in the scope id, converts it to bytes, puts it in the request and uses GRPC.

## Proposed Fix

Update the `metadata_addr` field (in the `AccountDataRequest`) to be a `string` that expects the bech32 (scope id). Have the query then do the decoding (if needed).

____

## For Admin Use

- [ ] Not duplicate issue

nagarajdivine:
## Description

* fix metadata account-data REST query.
closes: #2708 

---

write a little note why.

## Bug Fixes
* Fixed metadata account-data REST API query functionality to properly retrieve and handle account information across various supported data formats
* Enhanced address parsing with improved fallback mechanisms to support additional address format types for better compatibility and robustness

## Tests
* Expanded test coverage for address serialization, format conversion, and validation scenarios


.changelog/unreleased/improvements/2708-metadata-account-query.md
```diff
@@ -0,0 +1 @@
+* Fix metadata account-data REST query [#2708](https://github.com/provenance-io/provenance/issues/2708).
```
SpicyLemon, line 1: We should **also** have an entry in the `client-breaking` section about the `AccountDataRequest.metadata_addr` field being changed to a `string` (from a byte array).

nagarajdivine: IMO, main error occurs in gateway file near `protoReq.MetadataAddr, err = runtime.Bytes()`
so, if we replaces runtime.Bytes(val) with MetadataAddressFromBech32(val) directly in the gateway file, which correctly decodes the bech32 string into a MetadataAddress. In this cases, we can revert all changes this PR and just change only gateway file.

SpicyLemon: You can't make manual changes to any of the `.pb.go` or `.pb.gw.go` files because all changes will be lost the next time files are generated from protos. Additionally, if we leave it as a `byte[]` we need to also maintain current functionality.

However, in `x/metadata/types/address.go`, there's a `func (ma *MetadataAddress) Unmarshal(data []byte) error` that gets used here, and that we can change.

```
func (ma *MetadataAddress) Unmarshal(data []byte) error {
	*ma = data
	if len(data) == 0 {
		return nil
	}
	_, err := VerifyMetadataAddressFormat(data)
	return err
}
```

If we want to keep the proto field as a `[]byte`, then we should add to this function. It still needs to maintain it's current behavior when handed a byte array that's the actual bytes of a metadata address. But if `VerifyMetadataAddressFormat(data)` returns an error, then try to decode it as a bech32 (using `MetadataAddressFromBech32`). If bech32 decoding also fails, return just the error from `VerifyMetadataAddressFormat`.

If you do this, please also turn `TestMetadataAddressMarshal()` into a table-driven design and include both the bytes and bech32 versions of each of the metadata types (as well as negative test cases too).

SpicyLemon (approved):

Issue #248, opened 2021-04-13
Denom Metadata: Require unique denom strings and delimiter after SI prefix.


dwedul-figure:
## Summary

A denom string must uniquely identify a marker. If a denom string ambiguously refers to two or more markers, then we don't actually know the coin in question.

## Problem Definition

Current restrictions require a denom unit name or alias to be either the root coin name, or an SI Prefix + the root coin name. This will cause a collision in some cases when new markers are being created. For example, if an existing marker has a root coin name of `oocoin`, it might have a denom unit for `foocoin` (being one "femto oocoin"). Then a new coin comes along named "foocoin". We create the marker for it including a denom unit for the root coin name "foocoin". Now, if someone is trying to do a transaction with "1000 foocoin", we don't know which one they're talking about.

## Proposal

- [ ] Require either a `-` or `.` between the SI prefix and root coin name of all denom unit names and aliases. Since the base name and display name must both be denom units, this applies to those as well (for free).
- [ ] Update `GetRootCoinName` to remove any leading `-` or `.` (for cases when the root coin name isn't in a denom unit entry).
- [ ] None of the denom unit names or aliases can start with a `-` or `.`.
- [ ] Restrict denom names and aliases to 63 characters long (including any prefix and delimiter).
- [ ] Require all denom unit names and aliases to be unique across the chain.
    - [ ] Create an index for all denom unit names and aliases in the kvstore.
        - [ ] The key should have this format: `<key byte><denom name><delimiter byte><base name>`
            - [ ] The <key byte> should just be a single byte that uniquely identifies the index entries.
            - [ ] The <denom name> should be the name as bytes.
            - [ ] The <delimiter byte> can be anything as long as it's not possible in a denom name. Maybe just reuse the <key byte> here.
            - [ ] The <base name> should be the base name as bytes (the same base name used in the kvstore key for the marker).
        - [ ] The value should just be `0x01`.
        - [ ] Index entries should be made for all denom `metadata.Denom` strings, and all `metadata.Aliases` entry strings.
        - [ ] Create an iterator for the index that takes in a prefix to iterate over, breaks each key into the denom and base name, and calls a function with the denom and base name as parameters.
    - [ ] During `ValidateDenomMetadataExtended` check the index for existing entries for each denom name and alias that refer to different denom base names.
- [ ] Update `DenomMetadata` query to use the new index for looking up the denom metadata by denom string.

## Reasoning

[...]

PR #240, opened 2021-04-09
Add rpath lookup for pwd libwasmvm

mtps:
## Description

Add cgo ldflags to make the binary look in pwd for the libwasmvm.so

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #239, opened 2021-04-08
Fix the input status mapping for p8e memorialize contract.

dwedul-figure:
## Description

The input status enum has 3 entries: Unknown, Proposed, and Record.
It is being set by the result type of record specification which has four possible values: Unknown, Proposed, Record, and Record list.

This PR updates the mapping to treat record and record list as the same. Prior to this, a record list would end up getting treated as unknown.

This PR also updates the location of the contract specification id to only allow look at `contract.Definition.ResourceLocation.Ref.Hash`.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #2779, opened 2026-07-20
Update vault dependency for v1.2.1 release

nullpointer0x00:
## Description

Bumps `github.com/provlabs/vault` to `v1.2.1` (from `v1.1.0`) and wires in the changes that came with it. Since the last vault bump in provenance was `v1.1.0`, this PR picks up both the `v1.2.0` and `v1.2.1` releases; the release notes for both are included below. In `app/app.go`, `vaultkeeper.NewKeeper` now takes the exchange keeper and an exchange query server, which the vault module uses to settle and list exchange payments targeting a vault. In `app/upgrades.go`, the `MigrateVaultAccountPaymentDenomDefaults` calls were removed because that method was replaced upstream by a registered `Migrate1to2` (ConsensusVersion 1 -> 2) that will run via `runModuleMigrations` in the next upgrade handler after edelweiss. The vault protos under `third_party/proto` were synced from the release, swagger was regenerated, and the four new vault queries (`VaultNavs`, `NavValue`, `VaultPayment`, `VaultPayments`) were added to the provwasm stargate whitelist.

## Vault v1.2.1 Release Notes

This is a fast-follow patch release to v1.2.0 containing two minor fixes: a genesis fix so a `DefaultAumFeeBips` of zero survives export and import instead of being replaced with the module default, and removal of an unused import from `tx.proto` that tripped downstream `IMPORT_USED` lint checks.

No state migrations are required.

### Improvements

* Removed the unused `provlabs/vault/v1/vault.proto` import from `tx.proto` so downstream consumers linting with `IMPORT_USED` no longer report it [#252](https://github.com/provlabs/vault/issues/252).

### Bug Fixes

* Honor a genesis `DefaultAumFeeBips` of zero in `InitGenesis` instead of silently replacing it with the module default, so params round-trip through genesis export and import [#253](https://github.com/provlabs/vault/issues/253).

### Full Commit History

* https://github.com/provlabs/vault/compare/v1.2.0...v1.2.1

## Vault v1.2.0 Release Notes

This release introduces the internal NAV system and the peer-to-peer multi-asset
vault workflow, allowing a vault to accept external assets in exchange for its
underlying asset through `x/exchange` settlement, priced against a per-vault NAV
table managed by a designated NAV authority.

It also removes the mixed-denom (payment denom) vault functionality in favor of
the p2p workflow: vaults are now strictly single-denom on their underlying asset,
with a v1 to v2 migration that flattens existing mixed-denom vaults in place.
Vault share markers now require deposit access, restricting principal deposits to
the vault itself.

### Features

* Added an internal NAV system: a per-vault, per-denom NAV table with a designated NAV authority, `UpdateVaultNAV` and `UpdateNAVAuthority` transactions, `VaultNavs` and `NavValue` queries, genesis import/export, and `EventNAVUpdated`/`EventNAVAuthorityUpdated` events [#189](https://github.com/provlabs/vault/issues/189).
* Added a peer-to-peer multi-asset vault workflow that lets a vault take in external assets in exchange for its payment denom: new `AcceptAsset` and `RejectAsset` transactions settle or decline pending `x/exchange` payments targeting the vault, with supporting queries and simulations. Either the vault admin or asset manager may authorize a settlement. Built on the newly integrated Provenance `x/exchange` module (with its `x/hold` and `x/metadata` dependencies) [#192](https://github.com/provlabs/vault/issues/192).
* Added settlement-time responsibilities to the p2p multi-asset workflow: `AcceptAsset` now enforces an exact-price guardrail against the vault's internal NAV, records the settlement price in the internal NAV table (removing the entry with a new `EventNAVRemoved` when an outbound settlement drains the denom), publishes the updated NAV to the marker module attributed to the vault, and triggers a vault reconcile before settling. `UpdateVaultNAV` likewise reconciles first and publishes the updated NAV to the marker module [#193](https://github.com/provlabs/vault/issues/193).
* Add a `force` flag to `PauseVault` that pauses a vault best-effort when the pre-pause reconcile, valuation, or account persistence fails, falling back to an unvalidated write and recording every tolerated error on `EventVaultPaused.forced_error`; without it the pause stays strict and aborts on failure [#228](https://github.com/provlabs/vault/issues/228).

### Improvements

* Classify swap-out refund reasons with typed errors instead of fragile error-string matching [#90](https://github.com/provlabs/vault/issues/90).
* Add simulation tests to validate swap estimation accuracy [#112](https://github.com/provlabs/vault/issues/112).
* Aligned `AutoCLI` positional argument order with proto definitions for `CreateVault` (`[#135](https://github.com/provlabs/vault/issues/135).
* Switched the valuation engine to read prices exclusively from the per-vault Internal NAV table; `UnitPriceFraction` now performs a single `GetVaultNAV` lookup, no longer consults `MarkerKeeper.GetNetAssetValue`, no longer applies the temporary `uylds.fcc` 1:1 peg, and surfaces a distinct `no internal NAV entry for denom` error when an entry is missing. `MsgCreateVaultRequest` now carries an `initial_payment_nav` field and `Keeper.CreateVault` seeds the payment-denom NAV in the same cache context; the field is required (and is rejected as a vault-wide invariant) whenever `payment_denom` is set and differs from `underlying_asset`, so a vault can never enter an operational state that would silently stall fee collection or payment-denom swap-outs [#190](https://github.com/provlabs/vault/issues/190).
* Add an absolute interest-rate ceiling and overflow-safe interest calculations [PR 205](https://github.com/provlabs/vault/pull/205).
* Document and add test coverage for the bridge mint/burn supply-of-record model, where bridge operations never mutate the vault's `TotalShares` [#207](https://github.com/provlabs/vault/issues/207).
* Add a net TVV calculation that subtracts the outstanding AUM fee liability from gross vault value, and use it as the valuation basis for share NAV publication, NAV per share, deposit/redeem conversions, and the captured paused balance [PR 213](https://github.com/provlabs/vault/pull/213).
* Fix the lint workflow and run it against pull requests, upgrade golangci-lint to v2, and resolve the outstanding lint findings [PR 214](https://github.com/provlabs/vault/pull/214).
* Reworked `GetTVV` to value the accepted denoms directly and iterate the per-vault Internal NAV table for held assets instead of walking every principal-marker balance, so Total Vault Value cost scales with the number of valued denoms rather than with whatever is parked at the principal marker [#223](https://github.com/provlabs/vault/issues/223).
* Share NAV publication is now uint64-safe: when a vault's total shares exceed the `uint64` marker volume ceiling, the volume is capped at the uint64 maximum and the price is scaled down proportionally (computed via `math/big` so the intermediate product cannot overflow), so a correct price-per-share is always published instead of being skipped. Supplies that already fit in uint64 publish their exact share count unchanged. When a vault's supply drops to zero, a previously published share NAV is overwritten with a zero price so a stale price is not served for a share denom that no longer has supply [#233](https://github.com/provlabs/vault/issues/233).
* Disallow vault creation with a payment denom that differs from the underlying asset. The payment denom must now be empty (defaulting to the underlying) or equal to it. This creation-time check does not itself alter existing vaults; a separate state-machine-breaking migration flattens pre-existing mixed-denom vaults [#239](https://github.com/provlabs/vault/issues/239).
* Split the expected `ExchangeKeeper` interface into a keeper part and a new `ExchangeQueryServer` dependency so consuming apps can wire the exchange keeper and `exchangekeeper.NewQueryServer` directly instead of writing an adapter struct [#247](https://github.com/provlabs/vault/issues/247).

### Bug Fixes

* Guard NAV valuation multiplications with `SafeMul` so an oversized net asset value returns an error instead of panicking the block hook [#206](https://github.com/provlabs/vault/issues/206).
* Enforce per-block visit budgets on all ABCI queue processors (interest timeouts, fee timeouts, payout verification set, and pending swap-outs) and dequeue/refund pending swap-outs for paused vaults so they cannot camp at the front of the queue [#225](https://github.com/provlabs/vault/issues/225).
* Preserve a pending swap-out and its escrowed shares when a payout or refund fails critically, removing the queue entry only when its payout or refund commits [#226](https://github.com/provlabs/vault/issues/226).
* Reject swap-in deposits that are too small to mint at least one share, so no funds move when the computed share amount is zero [#237](https://github.com/provlabs/vault/issues/237).

### Deprecated

* Remove the mixed-denom request fields (`payment_denom` and `initial_payment_nav` on `MsgCreateVaultRequest`, `redeem_denom` on `MsgSwapOutRequest` and `QueryEstimateSwapOutRequest`) with their field numbers reserved, while `VaultAccount.payment_denom` and `PendingSwapOut.redeem_denom` remain deprecated on the wire for migration decoding [#240](https://github.com/provlabs/vault/issues/240).

### State Machine Breaking

* Remove the mixed-denom (payment denom) vault functionality in favor of the p2p workflow, making vaults strictly single-denom on their underlying asset, with a v1->v2 migration that flattens existing mixed-denom vaults in place [#240](https://github.com/provlabs/vault/issues/240).
* Enable require_deposit_access on every vault share marker, at creation and via the 1->2 migration for existing vaults, and grant the vault address deposit access so only the vault can move coins into its principal marker [#248](https://github.com/provlabs/vault/issues/248).

### Full Commit History

* https://github.com/provlabs/vault/compare/v1.1.0...v1.2.0

---

write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#pr-targeting)).

## Summary by CodeRabbit

* **New Features**
  * Added APIs to view vault NAV records and payment details.
  * Added transactions for updating NAV values, changing NAV authority, and accepting or rejecting settled assets.
  * Added NAV and asset settlement events for improved tracking.
  * Added governance support for updating marker `require_deposit_access`.
  * Added optional forced vault pauses when valuation or reconciliation encounters issues.
* **Improvements**
  * Clarified single-underlying-asset behavior; legacy denomination fields are now deprecated but retained for compatibility.
  * Added NAV data to vault state export and import.


.changelog/unreleased/dependencies/2779-bump-vault.md
```diff
@@ -0,0 +1 @@
+* `github.com/provlabs/vault` bumped to v1.0.13-0.20260720162054-5c2922c12d0c, a pre-v1.2.0 release candidate commit (from v1.1.0) [PR 2779](https://github.com/provenance-io/provenance/pull/2779).
```
SpicyLemon: Please run this command to generate the changelog for the dependencies section:
```
.changelog/get-dep-changes.sh --pr 2779 --id vault
```

nullpointer0x00: Ok, I did the command and removed this file.  Thanks!

SpicyLemon (approved):

PR #2799, opened 2026-07-31
separate marker and account permissions storage.

### LINKED ISSUE #2771: Separate marker and account permissions storage
## Summary

Move the storage of marker permissions out of the marker account and into its own key/value storage in the module.

## Problem Definition

The `MarkerAccount` has an `access_control` field that stores account addresses associated with permissions lists. We don't always need that information when looking up a marker, though. But because it's in the marker account, it gets read/parsed every time we look up the marker. This can be a significant amount of data as the list can grow as big as needed. Conversely, there are times where we are retrieving the entire marker just to look up marker permissions.

## Proposal

Store the `access_control` data on its own and stop storing it in the marker itself.

1. Choose a new type byte and store the permissions using this scheme: `<type byte><marker addr length><marker addr><addr length><addr> = <proto([]Access)>`.
2. Create a module migration that will move these permissions out of each marker and into the new storage place.
3. Refactor the keeper methods accordingly. Leave `GetMarkerByDenom` and `GetMarker` as they are so that calling them does NOT look up permissions. Instead, create `...withPerms` versions that look up both the marker and permissions. Create new keeper methods for checking account permissions.
4. Find places that use the `MarkerAccount` methods like `.HasAccess`, `.ValidateHasAccess` etc. Update them to look up the least amount of data possible to do what they need to do.

____

#### For Admin Use

- [ ] Not duplicate issue

nagarajdivine:
## Description

* separate marker and account permissions storage.
closes: #2771 

---

write a little note why.

## Summary by CodeRabbit

* **New Features**
  * Marker permissions are now stored and managed separately from marker account data.
  * Added permission-aware marker queries and access validation.
  * Added support for checking, granting, revoking, and listing marker permissions.
  * Added documentation for marker permission data.
  * Added a migration to preserve existing permissions during the storage upgrade.

* **Bug Fixes**
  * Improved authorization checks across marker operations, transfers, and securitization.
  * Marker exports now retain permission data correctly.


x/marker/keeper/keeper.go
```diff
@@ -214,17 +224,91 @@ func (k Keeper) GetMarker(ctx sdk.Context, address sdk.AccAddress) (types.Marker
 		if !ok {
 			return nil, fmt.Errorf("account at %s is not a marker account", address.String())
 		}
+		if ma, ok := macc.(*types.MarkerAccount); ok {
+			ma.AccessControl = nil
+		}
```
SpicyLemon: You don't need to do the casting twice. We've got a type assertion that makes sure that `*types.MarkerAccount` satisfies `types.MarkerAccountI`. Since nothing else satisfies `types.MarkerAccountI`, you're okay to just check for the concrete type instead of the interface.

```golang
macc, ok := mac.(*types.MarkerAccount)
if !ok {
	return nil, fmt.Errorf("account at %s is not a marker account", address.String())
}
macc.AccessControl = nil
```

x/marker/keeper/keeper.go
```diff
@@ -214,17 +224,91 @@ func (k Keeper) GetMarker(ctx sdk.Context, address sdk.AccAddress) (types.Marker
...
+// separately via SetAccess/RevokeAccessEntry/AddAccess/RemoveAccess.
 func (k Keeper) SetMarker(ctx sdk.Context, marker types.MarkerAccountI) error {
+	ma, isMA := marker.(*types.MarkerAccount)
+
+	// Validate needs the real access list — populate for validation only.
+	if isMA {
+		if err := k.PopulateMarkerPerms(ctx, ma); err != nil {
+			return fmt.Errorf("could not populate marker permissions for validation: %w", err)
+		}
+	}
```
SpicyLemon: This shouldn't be in here. It's a needless state lookup that is only for the purposes of validation. 

Instead, the `Validate()` method should be updated. Add a boolean argument for whether access control is required. In here, provide it as `false`, but for most (all?) the other calls to it, provide `true`.

x/marker/keeper/keeper.go
```diff
@@ -214,17 +224,91 @@ func (k Keeper) GetMarker(ctx sdk.Context, address sdk.AccAddress) (types.Marker
...
 	}
 	return nil, nil
 }
 
-// SetMarker sets a marker in the auth account store will panic if the marker account is not valid or
-// if the auth module account keeper fails to marshall the account.
+// SetMarker persists the marker WITHOUT touching permissions.
+// This is the normal path — perms live in their own store and are managed
+// separately via SetAccess/RevokeAccessEntry/AddAccess/RemoveAccess.
 func (k Keeper) SetMarker(ctx sdk.Context, marker types.MarkerAccountI) error {
```
SpicyLemon, line 239: Calling `SetMarker` should now do the following:
1. Ensure it's valid (allowing the access list to be either nil or valid).
2. Extract the access list from the marker.
3. Set the marker's access list to `nil`.
4. Save the marker acount.
5. If the access list isn't `nil`, save the access list.

We shouldn't have a public separate `SetMarkerWithPerms`. This `SetMarker` func should store everything. That way, the next time something comes up where we need to store a marker, we don't accidentally choose the wrong one.

You'll probably want to add a method to the `MarkerAccountI` interface like `ClearAccessList()`. Then you can use that on all the stuff that's a `MarkerAccountI` without having to cast it to a `*MarkerAccount`.

There should be two private methods for storing just marker account and just the marker's access list, and both of these should be called from `SetMarker`. But they shouldn't be public.

x/marker/keeper/keeper.go
```diff
@@ -214,17 +224,91 @@ func (k Keeper) GetMarker(ctx sdk.Context, address sdk.AccAddress) (types.Marker
...
+		// them here so persisting doesn't silently drop earlier grants.
+		type mergedGrant struct {
+			addr  sdk.AccAddress
+			perms map[types.Access]bool
+		}
+		order := make([]string, 0, len(ma.AccessControl))
+		merged := make(map[string]*mergedGrant)
+		for _, g := range ma.AccessControl {
+			mg, ok := merged[g.Address]
+			if !ok {
```
SpicyLemon: The `ok` variable should be named `seen` or `has` here. As it is, it looks like this `if` block should be handling an error case.

x/marker/keeper/marker.go
```diff
@@ -114,7 +114,7 @@ func (k Keeper) AddAccess(
 		if err := m.Validate(); err != nil {
 			return err
 		}
-		if err := k.SetMarker(ctx, m); err != nil {
+		if err := k.SetMarkerWithPerms(ctx, m); err != nil {
```
SpicyLemon: Here (and in Remove Access) there's no reason to re-save the marker account. You should only be saving permissions. In fact, you should only be adding (or removing) the access grants provided; there's no need to delete and re-save all of the other existing permissions.

x/marker/keeper/marker.go
```diff
@@ -558,9 +545,9 @@ func (k Keeper) CancelMarker(ctx sdk.Context, caller sdk.AccAddress, denom strin
 				" ensure marker account holds the entire supply of %s", inCirculation, totalSupply, denom)
 		}
 	case types.StatusProposed:
-		// for a proposed marker either the manager or someone assigned `delete` can perform this action
```
SpicyLemon, line 561: Please keep this comment. Comments should only ever be deleted if they are no longer true.

x/marker/keeper/permission.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	m, err := k.GetMarkerWithPerms(ctx, addr)
+	if err != nil {
+		return nil, err
+	}
+	if m == nil {
+		return nil, fmt.Errorf("marker %s not found for address: %s", denom, addr)
+	}
+	return m, nil
+}
```
SpicyLemon, line 113: Instead of calling `k.GetMarkerWithPerms`, this should look just like `GetMarkerWithPerms`, but call `GetMarkerByDenom` instead of `GetMarker`.

That way, the differences between `GetMarker` and `GetMarkerByDenom` remain in those methods and you don't have to try to replicate their behavior here.

x/marker/types/key.go
```diff
@@ -103,3 +107,12 @@ func GetMarkerFromNetAssetValueKey(key []byte) sdk.AccAddress {
...
 }
+
+// MarkerPermsKey returns the full store key for one address' permissions on a marker.
+func MarkerPermsKey(markerAddr, granteeAddr sdk.AccAddress) []byte {
+	key := make([]byte, 0, len(MarkerPermsKeyPrefix)+len(markerAddr)+len(granteeAddr)+2)
+	key = append(key, MarkerPermsKeyPrefix...)
+	key = append(key, address.MustLengthPrefix(markerAddr)...)
+	key = append(key, address.MustLengthPrefix(granteeAddr)...)
+	return key
+}
```
SpicyLemon, line 118: This isn't used or needed. The collections stuff takes care of al the key bytes now.

x/marker/keeper/msg_server.go
```diff
@@ -761,8 +765,14 @@ func (k msgServer) UpdateSendDenyList(goCtx context.Context, msg *types.MsgUpdat
 		if !marker.HasGovernanceEnabled() {
 			return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
 		}
-	} else if err = marker.ValidateHasAccess(msg.Authority, types.Access_Transfer); err != nil {
-		return nil, err
+	} else {
+		callerAddr, err := sdk.AccAddressFromBech32(msg.Authority)
+		if err != nil {
+			return nil, err
```
SpicyLemon: This error needs to be wrapped with context data.

x/marker/keeper/msg_server.go
```diff
@@ -731,7 +731,11 @@ func (k msgServer) SetAccountData(goCtx context.Context, msg *types.MsgSetAccoun
 			return nil, fmt.Errorf("%s marker does not allow governance control", msg.Denom)
 		}
 	} else {
-		if err = marker.ValidateHasAccess(msg.Signer, types.Access_Deposit); err != nil {
+		signerAddr, err := sdk.AccAddressFromBech32(msg.Signer)
+		if err != nil {
+			return nil, err
```
SpicyLemon: This error also needs two be wrapped to provide context.

x/marker/keeper/query_server.go
```diff
@@ -216,7 +216,7 @@ func accountForDenomOrAddress(ctx sdk.Context, keeper Keeper, lookup string) (ty
 	if addr, addrErr = sdk.AccAddressFromBech32(lookup); addrErr != nil {
 		account, err = keeper.GetMarkerByDenom(ctx, lookup)
 	} else {
-		account, err = keeper.GetMarker(ctx, addr)
+		account, err = keeper.GetMarkerWithPerms(ctx, addr)
```
SpicyLemon: You should leave this the same as it was. That way, this `accountForDenomOrAddress` function always returns a marker without permissions.

Most queries don't need the access list. For those that do, create a new `accountForDenomOrAddressWithPerms` that calls `accountForDenomOrAddress` and if the marker account isn't `nil` from that, call `PopulateMarkerPerms` for it.

x/marker/keeper/permission.go
```diff
@@ -0,0 +1,161 @@
...
+
+// PopulateMarkerPerms looks up and sets the marker's AccessControl from the perms store.
+func (k Keeper) PopulateMarkerPerms(ctx sdk.Context, ma *types.MarkerAccount) error {
```
SpicyLemon: Since everything else passes the marker account around as an interface, it's probably easiest if you take in the interface here. Add a `SetAccessList` to the `MarkerAccountI` interface so that you can set it without having to cast it to the concrete type. That'd probably get rid of the need to cast to `*MarkerAccount` in a lot of places.

.changelog/unreleased/features/2771-marker-perms-storage.md
```diff
@@ -0,0 +1 @@
+* separate marker and account permissions storage [#2771](https://github.com/provenance-io/provenance/issues/2771).
```
SpicyLemon: Please capitalize the first letter here.

SpicyLemon (changes requested):
There's a few things to refactor here. The key thing is that there shouldn't be a separate (public) `SetMarkerWithPerms`. There should be two private methods, one for setting the marker account, and one for setting the marker perms. The `SetMarker` method should set the marker account, then, if there are perms in the marker, set the marker perms.

Also, the `GetMarker` and `GetMarkerByDenom` need to have their comments updated to indicate that the returned markers no longer have the permissions populated, and that you need to use either `GetMarkerWithPerms`, `GetMarkerByDenomWithPerms` or use the new permissions methods directly.

x/marker/keeper/keeper.go
```diff
@@ -209,26 +219,91 @@ func (k Keeper) NewMarker(ctx sdk.Context, marker types.MarkerAccountI) types.Ma
...
+	if len(accessList) > 0 {
+		if err := k.setMarkerAccessList(ctx, marker.GetAddress(), accessList); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// setMarkerAccount persists just the marker's account fields (no permissions).
+// Only ever called from SetMarker.
```
SpicyLemon: This comment isn't actually true. It sets the marker as provided, regardless of whether it has permissions or not. The comment should instruct people to clear the access list before calling `setMarkerAccount`.

x/marker/keeper/keeper.go
```diff
@@ -209,26 +219,91 @@ func (k Keeper) NewMarker(ctx sdk.Context, marker types.MarkerAccountI) types.Ma
...
+
+	accessList := marker.GetAccessList()
+	marker.ClearAccessList()
+
+	k.setMarkerAccount(ctx, marker)
+
+	if len(accessList) > 0 {
+		if err := k.setMarkerAccessList(ctx, marker.GetAddress(), accessList); err != nil {
+			return err
+		}
```
SpicyLemon, line 253: You should put the access list back in the marker here. Otherwise, calling `SetMarker` will secretly remove the access list from the provided `marker` argument, affecting it outside the `SetMarker` call too.

I'd do this by adding a `SetAccessList` method to the `MarkerAccountI`. That way you don't have to do any iteration or anything.

x/marker/keeper/permission.go
```diff
@@ -0,0 +1,156 @@
...
+	return false
+}
+
+// SetAccess sets (or deletes, if empty) the permission list for addr on a marker.
+func (k Keeper) SetAccess(ctx sdk.Context, markerAddr, addr sdk.AccAddress, perms types.AccessList) error {
+	key := collections.Join(markerAddr, addr)
+	if len(perms) == 0 {
+		err := k.markerPerms.Remove(ctx, key)
+		if err != nil && !errors.Is(err, collections.ErrNotFound) {
+			return err
```
SpicyLemon: This error needs to be wrapped.

x/marker/keeper/permission.go
```diff
@@ -0,0 +1,156 @@
...
+func (k Keeper) SetAccess(ctx sdk.Context, markerAddr, addr sdk.AccAddress, perms types.AccessList) error {
+	key := collections.Join(markerAddr, addr)
+	if len(perms) == 0 {
+		err := k.markerPerms.Remove(ctx, key)
+		if err != nil && !errors.Is(err, collections.ErrNotFound) {
+			return fmt.Errorf("could not remove the permissions that %s has on marker %s: %w", addr, markerAddr, err)
+		}
+		return nil
+	}
+	return k.markerPerms.Set(ctx, key, types.MarkerPermissions{Permissions: perms})
```
SpicyLemon, line 44: This error needs to be wrapped too.

Basically, any error returned by the collections needs to be wrapped. Those errors don't contain any context, so if a user gets just that error, it doesn't really help point out what the problem is.

There's other spots in here too.

SpicyLemon:
There's some stuff in the `proposal_handler.go` file that need to be addressed.
As it is, the `SetAdministrator` method wipes out every existing grant on the marker, and `RemoveAdministrator` is a silent no-op.

Also, `FinalizeMarker` no longer fails if the supply is zero and there aren't any minters.

x/marker/keeper/marker.go
```diff
@@ -155,12 +158,19 @@ func (k Keeper) RemoveAccess(ctx sdk.Context, caller sdk.AccAddress, denom strin
 		if err = m.RevokeAccess(remove); err != nil {
 			return fmt.Errorf("access revoke failed: %w", err)
 		}
-		if err := m.Validate(); err != nil {
+		if err := m.ValidateWithAccessControl(); err != nil {
 			return err
 		}
 		if err := k.SetMarker(ctx, m); err != nil {
```
SpicyLemon, line 164: There's no longer a reason to set the marker here. Nothing changed except permissions. Just update the one permissions record.

x/marker/keeper/marker.go
SpicyLemon, line 123: Here too. Don't store the marker again (it didn't change). Just update the single permissions record.

x/marker/keeper/permission.go
```diff
@@ -0,0 +1,156 @@
...
+	denom := ""
+	if m, err := k.GetMarker(ctx, markerAddr); err == nil && m != nil {
+		denom = m.GetDenom()
+	}
+	strs := make([]string, len(addrs))
+	for i, addr := range addrs {
+		strs[i] = addr.String()
+	}
+	return fmt.Errorf("none of %q have permission %s on %s marker (%s)", strs, role, denom, markerAddr)
+}
```
SpicyLemon, line 156: Please create a `AtLeastOneHasAccess` helper here that just does the checking and returns a boolean.

Then, in here, you call that one and create the error only if needed.

Once you have that, in the send restriction, instead of calling `ValidateAtLeastOneHasAccess`, call the new `AtLeastOneHasAccess` instead.

The reason why, is because in the send restriction, we throw away the error anyway. But in order to create that error, it has to do a lookup on the marker. So we do a lookup of the marker for no reason.

x/asset/keeper/msg_server.go
```diff
@@ -274,19 +274,22 @@ func (m msgServer) CreateSecuritization(goCtx context.Context, msg *types.MsgCre
 		trancheCount++
 	}
 
-	// Reassign the pools permissions to the asset module account (prevent the pools from being transferred)
-	var poolCount uint32
+	poolMarkers := make([]markertypes.MarkerAccountI, 0, len(msg.Pools))
 	for _, pool := range msg.Pools {
-		poolMarker, err := m.markerKeeper.GetMarkerByDenom(ctx, pool)
+		poolMarker, err := m.markerKeeper.GetMarkerByDenomWithPerms(ctx, pool)
```
SpicyLemon, line 279: Please clean up the stuff in here.

1. We probably should create a `MarkerExists(ctx, denom)` keeper method (in the marker module). It can use the index to check if a marker exists without having to read the stuff into state.
2. Switch to using the more direct keeper methods for manipulating permissions.
3. Nothing is changing in the marker except the permissions, so there's no need to set it again.

SpicyLemon:
There's an existing bug that I'd like to see fixed in this PR.

In `x/marker/types/accessgrant.go`, in `ValidateGrants`, there's a `registered` map that's not being populated. Basically, turn it into this:

```
// ValidateGrants checks a collection of grants and returns any errors encountered or nil
func ValidateGrants(grants ...AccessGrant) error {
	registered := make(map[string]bool)
	for _, grant := range grants {
		if _, exists := registered[grant.Address]; exists {
			return ErrDuplicateAccessEntry
		}
		registered[grant.Address] = true
		if err := grant.Validate(); err != nil {
			return err
		}
	}
	return nil
}
```

Note that the duplicate check should be first, then populate `registered` then validate it.

PR #2810, opened 2026-08-12
Key name records on a hash of the full name so different names can't share a record

### LINKED ISSUE #2683: Use hashed name for key instead of just the segments.
## Summary

In the `x/name` module, for the name records, use a hashed version of the full name as the key instead of a hashed version of the segments.

## Problem Definition

In the `x/name` module, when storing a name record, we use a key that's the result of extracting all the name segments and hashing them. This has two problems: 1) It's a lot of extra work for nothing. 2) It can cause collisions in the keys for different names.

## Proposal

Since this is changing the keys used to read/write some records, it'll require a migration.

Update `getNamePrefixByType` (in `x/name/types/keys.go`) to just get a hash of the full name (rather than splitting it and using the segments. You'll need a copy of the original function for the migration, though.

Then, create a migration that looks up all the old keys, deletes the entry, and re-writes it using the new key. This migration should be done as a module migration (i.e. bump the module's consensus version etc.).

____

#### For Admin Use

- [ ] Not duplicate issue

eastagiletracker:
This PR proposes keying `x/name` records on a hash of the full name, so that two names made up of the same letters split into different segments can no longer land on one record — Fixes #2683. We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/238. You can sign in with your GitHub ID to claim ownership of the project.

## The defect

`GetNameKeyPrefix` hashed the name's segments concatenated back to front with nothing between them, so `leaf.mid.pb` hashes `"pb" + "mid" + "leaf"` and `midleaf.pb` hashes `"pb" + "midleaf"` — the same bytes, and therefore the same store key. Three such pairs, run against `main` at 8d92467:

```
$ go test ./x/name/types/ -run 'TestNameKeySuite/TestNameKeyPrefixSegmentCollisions' -count=1 -v
--- FAIL: TestNameKeySuite/TestNameKeyPrefixSegmentCollisions/extra_segment_in_the_middle
    Error:    Should not be: "03def7306f705fc6861d1f0979f06ec9b62d2ad26be5a1461109218fad7d57014e"
    Messages: "leaf.mid.pb" and "midleaf.pb" are different names and must not share a store key
--- FAIL: TestNameKeySuite/TestNameKeyPrefixSegmentCollisions/extra_segment_at_the_front
    Messages: "kyc.identity.pb" and "identitykyc.pb" are different names and must not share a store key
--- FAIL: TestNameKeySuite/TestNameKeyPrefixSegmentCollisions/two_segments_vs_one
    Messages: "aa.bb" and "bbaa" are different names and must not share a store key
```

One key means one record, and the first name bound owns it. With `midleaf.name` bound to one account, the keeper answers for `leaf.mid.name` — a name nobody bound — with that account's record and no error:

```
GetRecordByName("leaf.mid.name") -> record=midleaf.name: cosmos1dmz82cgdj5ck2aslm4xcgw2u85q843zr2yfmnx, err=<nil>
NameExists("leaf.mid.name")      -> true
ResolvesTo("leaf.mid.name", owner-of-midleaf.name) -> true
```

`ResolvesTo` is the permission check behind every attribute write and delete (`x/attribute/keeper/keeper.go` lines 168, 259, 327, 394 and 457) and behind `DeleteName` (`x/name/keeper/msg_server.go:124`), so whoever binds one name of a colliding pair holds the other name's permissions. The other half is a denial: binding the second name returns `name is already bound to an address` even though it was never bound. Both names in a pair need only a shared unrestricted parent, so the pairs are reachable by ordinary `MsgBindNameRequest` calls rather than only by governance.

## The change

* `GetNameKeyPrefix` hashes the full normalized name, dots included, which is what `x/attribute` already does for its own name keys (`GetNameKeyBytes`). Segment trimming and both `ErrNameInvalid` cases are unchanged, so `" name . domain "` still keys the same as `name.domain`.
* `GetLegacyNameKeyPrefix` keeps the old derivation, only so the migration can find the existing entries.
* `Migrate2to3` rewrites the name records and the address index entries under their new keys, and the module's consensus version goes 2 → 3. Single-segment names hash to the same key as before, so only multi-segment records move. All deletes happen before any writes, and re-running the migration is a no-op.
* `x/name/spec/02_state.md` described a dot-joined list of per-label hashes, which is not what the code stored either before or after; it now matches the code.

## Verification

Baseline first on a clean `main`, then again with the change, same command each time:

```
$ go test ./x/name/... ./x/attribute/... ./internal/... -count=1
ok  github.com/provenance-io/provenance/x/name/client/cli     14.389s
ok  github.com/provenance-io/provenance/x/name/keeper         10.627s
ok  github.com/provenance-io/provenance/x/name/simulation     13.754s
ok  github.com/provenance-io/provenance/x/name/types           4.458s
ok  github.com/provenance-io/provenance/x/attribute/keeper    13.994s
... (attribute cli/rest/simulation/types and internal/... all ok)
```

Both runs are green — nothing that passed before fails now. `go build ./...` and `golangci-lint run ./x/name/...` (v2.4.0, your `.golangci.yml`) are also clean. New coverage: the collision cases above, a keeper test that binds `midleaf.name` and `leaf.mid.name` to different accounts and checks each resolves only to its own owner, a round-trip migration test that writes records under legacy keys and reads them back through the keeper afterwards, and a repeat-run test. Each of them fails against the old derivation — the keeper one with `SetNameRecord(leaf.mid.name): name is already bound to an address`.

Three existing expectations moved: `TestSetup`'s exported genesis, `TestIterateRecord`, and the reverse-lookup CLI output all pin the order records come back in, which is store-key order and necessarily changes with the key. Only the order in those expectations changed.

Worth flagging: #2411 moves this module to collections and carries the segment hash over unchanged in `HashedStringKeyCodec.ComputeHash`, so whichever lands first, the other needs the same one-line hashing change. I kept to the approach described in #2683 (hash the full name, re-key in a module migration) rather than the un-hashed key alternative, since it keeps the fixed-length keys.

## How this was managed

This work was tracked on a live board built from this repository's own issues, pull requests and milestones — 2313 stories and 46 labels: the story is [Use hashed name for key instead of just the segments](https://eastagiletracker.com/projects/238/stories/116895) and the board it lives on is at https://eastagiletracker.com/projects/238.

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented collisions between names with identical labels arranged in different segments.
  * Improved name lookups so each distinct name resolves only to its correct owner.
  * Preserved independent records when one of multiple similarly structured names is deleted.
  * Normalized name handling by ignoring surrounding whitespace.

* **Migrations**
  * Added automatic migration support for existing name records and address indexes, preserving lookup behavior during upgrades.

### DIFF
--- a/.changelog/unreleased/bug-fixes/2683-name-record-key-collisions.md
+++ b/.changelog/unreleased/bug-fixes/2683-name-record-key-collisions.md
@@ -1,0 +1,1 @@
+* Key name records on a hash of the full name so that names made up of the same letters split into different segments, e.g. `midleaf.pb` and `leaf.mid.pb`, no longer share a record [#2683](https://github.com/provenance-io/provenance/issues/2683).
--- a/x/name/client/cli/cli_test.go
+++ b/x/name/client/cli/cli_test.go
@@ -270,12 +270,12 @@ func (s *IntegrationTestSuite) TestReverseLookupCommand() {
 		{
 			"query name, json output",
 			[]string{s.accountAddr.String(), fmt.Sprintf("--%s=json", cmtcli.OutputFlag)},
-			"{\"name\":[\"example.attribute\",\"attribute\"],\"pagination\":{\"next_key\":null,\"total\":\"0\"}}",
+			"{\"name\":[\"attribute\",\"example.attribute\"],\"pagination\":{\"next_key\":null,\"total\":\"0\"}}",
 		},
 		{
 			"query name, text output",
 			[]string{s.accountAddr.String(), fmt.Sprintf("--%s=text", cmtcli.OutputFlag)},
-			"name:\n- example.attribute\n- attribute\npagination:\n  next_key: null\n  total: \"0\"",
+			"name:\n- attribute\n- example.attribute\npagination:\n  next_key: null\n  total: \"0\"",
 		},
 		{
 			"query name that does not exist, text output",
--- a/x/name/keeper/keeper_test.go
+++ b/x/name/keeper/keeper_test.go
@@ -96,10 +96,10 @@ func (s *KeeperTestSuite) TestSetup() {
   name: test.root
   restricted: false
 - address: %[1]s
-  name: name
+  name: example.name
   restricted: false
 - address: %[1]s
-  name: example.name
+  name: name
   restricted: false
 - address: %[3]s
   name: %[2]s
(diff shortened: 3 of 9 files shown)


SpicyLemon:
While we appreciate the PR and attention, this change was handled in PR #2411.

PR #2811, opened 2026-08-12
Vault v1.2.4

nullpointer0x00:
## Description

Bumps `github.com/provlabs/vault` from `v1.2.3` to `v1.2.4` and syncs everything downstream of it:

* `go.mod` / `go.sum` bump.
* Regenerated `third_party/proto/provlabs/vault/v1/*.proto` and `client/docs/swagger-ui/swagger.yaml`.
* Added `/provlabs.vault.v1.Query/Params` to the CosmWasm stargate query whitelist (`internal/provwasm/stargate_whitelist.go`). It was the only Provenance module with a `Params` rpc that wasn't whitelisted, and v1.2.4 adds a param (`gov_only_vault_creation`) a contract may want to read before attempting a `CreateVault`.

No `app.go` wiring was needed. The three new vault invariants register automatically through the existing `registerInvariants()` loop (`app/app.go:1051`), which picks up any module implementing `module.HasInvariants`.

---

## Upgrade notes

**Vault consensus version 2 → 3.** v1.2.4 ships a v2→v3 migration that is required, not optional. It:

* Seeds the new materialized total-value entry for every existing vault. The `total-value` invariant reads this state without repairing it, so an unseeded vault reports as broken on an invariant-enabled chain.
* Enables `gov_only_vault_creation` on `pio-mainnet-1` only. The param defaults to **disabled** so dev/docker/testnet chains can create vaults without a proposal — meaning any chain other than mainnet leaves `CreateVault` open to any signer until an `UpdateParams` proposal says otherwise.

The `forsythia` handler (v1.30.0) calls `runModuleMigrations`, so mainnet picks this up. ⚠️ **`forsythia-rc2` is currently `{}` with no handler** (`app/upgrades.go:61`), which resolves to a true no-op that returns the version map unchanged without running migrations. If this bump ships in v1.30.0-rc2, rc2 chains get the v1.2.4 binary without the v2→v3 migration. Worth confirming that's intended before tagging rc2.

**Flat fee for the new msg.** `MsgRepriceVaultRequest` has no `x/flatfees` entry, so it bills at `Params.DefaultCost`. That's low risk — the msg is NAV-authority-only — but if a non-default cost is wanted at activation rather than later, it has to go in the upgrade handler; otherwise it's a `tx flatfees update` gov proposal any time after.

---

### Features

* **`RepriceVault` transaction** — the batched form of `UpdateVaultNAV` with identical per-entry rules, plus an optional `resume` that unpauses the vault in the same state transition. Lets a pricing oracle run a repricing cadence without a second signature from the admin or asset manager. Batch is capped at `MaxRepriceBatchSize` (1000); duplicate denoms rejected. Leaving `resume` unset applies the batch and leaves the pause untouched, covering both a restatement too large for one transaction (so the vault reopens exactly once, on the last message) and a NAV authority writing a held asset down during someone else's pause without being able to lift it. Setting `resume` requires a strict pause the current NAV authority took itself, checked before any price is written — operator, forced, and automatic pauses, and pauses predating a NAV authority rotation, still need a management unpause.
* **`PauseVault` now accepts the NAV authority as a signer**, since a pricing oracle observes a depeg-type event before anyone else. `UnpauseVault` is unchanged — admin or asset manager only.
* **`VaultAccount` gained `paused_by` and `paused_forced`**, recording who took the current pause and whether it waived the strict reconcile and valuation gate. Both clear on unpause. A vault already paused when this ships carries neither, so its pause needs a management unpause.
* **Vault invariants registered with the crisis module** — `total-value`, `share-supply`, and `escrowed-shares`. A lookup entry with no vault account behind it is logged and counted rather than halting an invariant-enabled chain.

### API Breaking

* `MsgAcceptAssetRequest` now carries the full `payment` being approved instead of the `source` and `external_id` fields that referenced it. Fields 3 and 4 are reserved. Settlement fails unless the pending payment matches the carried terms exactly, so the asset manager's signature commits to the economic terms of the deal.

### State Machine Breaking

* Added the `gov_only_vault_creation` module param (see upgrade notes above).
* `UpdateVaultNAV` now requires the vault to be paused to reprice an asset the vault **holds** — that price step moves the share price, and on a live vault a user could swap in ahead of it and out after it, taking the difference from existing shareholders. Pricing a denom the vault does not hold stays available while live, as does restating a held asset at the unit price it already carries.
* Materialized each vault's total value into module state (prefix 12), so reading total vault value costs one store read instead of one per priced denom.
* Reconcile the vault before pricing a swap-out so the limits gate the current value.
* Range-reduce the `e^(rt)` exponent in `ExpDec` to prevent sign inversion.
* `BridgeMintShares` and `BridgeBurnShares` rejected on a paused vault, bringing the bridge inside the pause circuit breaker.

### Bug Fixes

* Reject deposits from an address on the underlying marker's deny list, which the marker bypass used by swap-in and interest/principal deposits previously skipped.
* Validate every genesis pending swap-out entry and skip malformed requests in the `EndBlocker`, so an imported entry can no longer panic and halt the chain when it matures.
* Reject nil/zero/negative `assets` on `EstimateSwapIn` with `InvalidArgument` instead of panicking on a nil `big.Int`.
* Use `SafeAdd` for virtual asset/share offsets in `CalculateSharesProRata` / `CalculateRedeemProRata` — the redeem path has no recover.
* Republish the marker mirrored share NAV on any reconcile that moves net total vault value, so external consumers no longer read a stale price-per-share.
* Treat drained-denom NAV cleanup at the end of an outbound `AcceptAsset` as best-effort rather than reverting a settlement whose funds already moved.
* Enforce the two-year `MaxWithdrawalDelay` cap in `VaultAccount.Validate()` and compute swap-out payout time with a checked conversion, so an out-of-range delay can't wrap into the past and pay out in the requesting block.
* Compute bridge mint capacity with `SafeSub` in `BridgeMintShares`.
* Bound and validate the `shares` string on `EstimateSwapOut` before the vault lookup, without echoing it back in the error.
* Remove a live vault from the `PayoutVerificationSet` only in the same atomic write that transitions it, so a failed transition retries instead of stalling accrual.
* Stop the v1→v2 migration from halting the chain on a legacy vault that fails the current `VaultAccount.Validate()` — it's now paused and persisted with the tolerated error on `EventVaultPaused.forced_error`.
* Reject deposits priced against zero total assets with shares outstanding, so `SwapIn` fails closed and `EstimateSwapIn` agrees.
* Error on a missing or unreadable key in `PendingSwapOutQueue.Dequeue` instead of reporting success.
* Cap `OutstandingAumFee` at the vault's gross total value so an uncollectable fee can't accumulate into a claim larger than the vault holds.
* Reschedule payout and fee timeouts atomically.
* Export the payout verification set in genesis and re-derive it on import so vaults keep automatic interest scheduling.
* Default a nil paused balance to zero when auto-pausing a vault.

### Improvements

* Documented the NAV freshness model and the NAV authority's repricing responsibility.
* Jitter swap-out retry times deterministically so identical failures stop re-clustering.
* Documented the drain-before-rotate procedure for bridge address rotation — rotation does not move or burn share balance held by the outgoing bridge, so mint capacity stays reduced until it's transferred and burned.
* Cap each vault's NAV table at 2000 priced denoms.
* Unpause reads materialized total vault value instead of walking the NAV table.
* Block `ibc/` voucher denoms from vault underlying assets and NAV entries.

Full commit history: https://github.com/provlabs/vault/compare/v1.2.3...v1.2.4

---

- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work. — upstream [provlabs/vault#270](https://github.com/provlabs/vault/pull/270)
- [x] Wrote unit and integration [tests](https://github.com/provenance-io/provenance/blob/main/CONTRIBUTING.md#testing) — n/a, dependency bump; covered upstream. The whitelist addition is covered by `Test_StargateWhitelistedURLs`.
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`). — n/a, vault specs live upstream; swagger regenerated.
- [x] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code). — n/a, no new exported symbols.

## Summary by CodeRabbit

* **New Features**
  * Added support for repricing vaults with batched NAV updates and optional resumption.
  * Added configurable governance-only vault creation.
  * Added vault payment details for asset settlement workflows.
  * Added payout verification state to genesis export and import.
  * Added vault parameter query support.

* **Enhancements**
  * Improved pause and resume controls with initiator and forced-pause tracking.
  * Clarified validation requirements for swap-out shares.
  * Expanded transaction documentation and authorization rules.


SpicyLemon (dismissed):

SpicyLemon (approved):

SpicyLemon:
This PR was approved despite the failing protobuf check. The protobuf check fails because a message was moved from one file to another (in the same package). That can be breaking for code generation in some languages. From a protobuf wire-compatibility standpoint, it's not breaking, though.

